PR #33, opened 2026-03-30
test: align MLX single-file validation e2e

roryford:
Summary:
- Rename MLX single-file E2E test to match current behavior
- Assert single-file MLX validation throws HuggingFaceError
- Keep tests aligned with snapshot-only MLX validation logic

Validation:
- swift test --filter DownloadValidationE2ETests
- swift test


roryford:
Changes already landed on main via a different path (6be7afa). Closing as stale.

PR #36, opened 2026-03-31
fix: thread-safe pin store, CI-testable routing, Foundation probe audit, MLX docs

roryford:
- PinnedSessionDelegate.pinnedHosts is now guarded by NSLock
- Missing-pin warning logged via os_log when no pins configured for a host
- DefaultBackends routing extracted into pure functions testable in CI
- FoundationBackend probe session history audited and documented
- MLXBackend.generate() MainActor constraint documented with rationale

Part of QA remediation plan.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,8 +29,8 @@ jobs:
             ${{ runner.os }}-${{ runner.arch }}-build-
 
       - name: Run tests
-        run: swift test --filter "BaseChatCoreTests|BaseChatUITests"
+        run: swift test --filter "BaseChatCoreTests|BaseChatUITests|BaseChatBackendsTests"
 
-      # BaseChatBackendsTests and BaseChatE2ETests require physical hardware
-      # or specific OS versions (iOS 26 / macOS 26 for FoundationBackend).
+      # BaseChatE2ETests requires physical hardware or specific OS versions
+      # (iOS 26 / macOS 26 for FoundationBackend).
       # Run these locally before merging backend changes.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,8 +29,8 @@ jobs:
             ${{ runner.os }}-${{ runner.arch }}-build-
 
       - name: Run tests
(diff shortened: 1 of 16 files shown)


roryford (commented):
## Code Review Findings

### 🔴 Issue 1 — Race condition in `PinnedSessionDelegate.pinnedHosts` (High)

**File:** `Sources/BaseChatBackends/PinnedSessionDelegate.swift`

The host-pin lookup makes three separate calls to the `pinnedHosts` getter in the TLS delegate callback. Each call independently locks/unlocks the `NSLock`, leaving race windows between them. The forced unwraps will crash if another thread clears `pinnedHosts` between the nil-check and the subscript:

```swift
// BAD — three separate lock/unlock cycles
let expectedPins = Self.pinnedHosts[host]!  // can crash if cleared between checks
```

**Fix:** Capture the value once before the conditional:

```swift
let pins = Self.pinnedHosts[host]
    Log.network.warning(...)
    completionHandler(.performDefaultHandling, nil)
    return
}
let expectedPins = pins!
```

---

### 🟠 Issue 2 — `backendTypeName(for: .gguf)` ignores `#if Llama` trait (Medium)

**File:** `Sources/BaseChatBackends/DefaultBackends.swift`

`backendTypeName(for: .gguf)` unconditionally returns `"LlamaBackend"`, but the actual backend registration is wrapped in `#if Llama`. In a build without the Llama trait, the routing function promises a backend that isn't registered — making the routing tests pass falsely.

**Fix:** Mirror the same conditional compilation as the registration:

```swift
static func backendTypeName(for modelType: ModelType) -> String? {
    switch modelType {
    #if Llama
    case .gguf: return "LlamaBackend"
    #endif
    #if MLX
    case .mlx: return "MLXBackend"
    #endif
    case .foundation: return "FoundationBackend"
    default: return nil
    }
}
```

Update the test to use the same guard:

```swift
func test_routing_gguf_mapsToLlamaBackend() {
    #if Llama
    XCTAssertEqual(DefaultBackends.backendTypeName(for: .gguf), "LlamaBackend")
    #else
    XCTAssertNil(DefaultBackends.backendTypeName(for: .gguf))
    #endif
}
```

PR #37, opened 2026-03-31
test: move CloudBackendSSETests into CI and add BackendContractSuite protocol

roryford:
- CloudBackendSSETests.swift moved from BaseChatE2ETests to BaseChatBackendsTests
- CI filter updated to include BaseChatBackendsTests
- BackendContractTests.swift introduces shared conformance test protocol
- Fix AsyncIterator.next() async signature in SSEPayloadReplayTests

Part of QA remediation plan.


roryford (commented):
## Code Review Finding

### 🔴 `BackendContractSuite` protocol extension methods will not be discovered by XCTest (Critical)

**File:** `Tests/BaseChatBackendsTests/BackendContractTests.swift`

XCTest's runtime test discovery is Objective-C reflection-based and **cannot see methods defined in Swift protocol extensions**. When `OpenAIBackendTests`, `ClaudeBackendTests`, and `MLXBackendTests` adopt `BackendContractSuite`, the 6 contract tests will be silently skipped — the runner reports 0 tests from the extension, no failure, no warning.

Verified: a test class that only adopts `BackendContractSuite` runs `0 tests` with `swift test --filter`.

**Fix — pick one:**

**Option A (recommended): Use explicit forwarding methods on each test class**
Each adopting class calls through to a shared helper struct rather than relying on protocol extension discovery:

```swift
// BackendContractTests.swift — helpers, not auto-discovered
struct BackendContractChecks {
    static func runAll<B: InferenceBackend>(backend: B, testCase: XCTestCase) {
        XCTAssertFalse(backend.isModelLoaded)
        XCTAssertFalse(backend.isGenerating)
        XCTAssertThrowsError(try backend.generate(prompt: "hi", systemPrompt: nil, config: .init()))
        XCTAssertFalse(backend.capabilities.supportedParameters.isEmpty)
    }
}

// In each backend test class — these ARE discovered:
func test_contract_allInvariants() {
    BackendContractChecks.runAll(backend: makeBackend(), testCase: self)
}
```

**Option B: Migrate `BackendContractTests.swift` to Swift Testing (`@Suite`/`@Test`)**
Swift Testing properly discovers methods defined in protocol extensions. Since the plan already uses Swift Testing for some suites (check existing files with `@Suite`), this is consistent with the codebase. Adopting classes use `@Suite` and gain auto-discovery.

**Option C: Explicitly redeclare each contract test in every adopting class**
```swift
// In OpenAIBackendTests:
func test_contract_isNotLoadedOnInit() { XCTAssertFalse(makeBackend().isModelLoaded) }
// ... (verbose but guaranteed to run)
```

As implemented, the `BackendContractSuite` gives a false sense of coverage — the contract is declared but never enforced.

PR #38, opened 2026-03-31
test: add MLX tests, contract suite adoption, cancellation and multi-turn tests

roryford:
- MLXBackendTests.swift created with 9 tests (hardware-gated where needed)
- OpenAIBackendTests and ClaudeBackendTests adopt BackendContractSuite
- Multi-turn history serialisation tests verify outbound request body
- stopGeneration() mid-stream cancellation tests for both cloud backends
- Keychain-backed configure() path tests

Part of QA remediation plan.


roryford (commented):
## Code Review Finding

### 🟠 `MockURLProtocol.stopLoading()` does not cancel the async SSE background task (Medium)

**File:** `Sources/BaseChatTestSupport/MockURLProtocol.swift`

The new `asyncSSE` stub mode spawns a `DispatchQueue.global().async` block that delivers chunks with delays. When URLSession cancels the request (which happens when the consumer breaks from the stream or calls `stopGeneration()`), URLSession calls `stopLoading()` on the protocol instance — but `stopLoading()` does nothing:

```swift
public override func stopLoading() {
    // Nothing to clean up for synchronous stubs  ← incorrect for asyncSSE
}
```

The background delivery task continues running and calling `client?.urlProtocol(self, didLoad:)` after the client has been torn down. This causes a use-after-free on `client` and can produce intermittent crashes or unexplained failures in tests that run after a cancellation test.

The cancellation tests currently pass because assertions execute before many background chunks fire — it's a timing dependency, not correctness.

**Fix:** Track the work item and cancel it in `stopLoading()`:

```swift
private var asyncDeliveryItem: DispatchWorkItem?

public override func stopLoading() {
    asyncDeliveryItem?.cancel()
    asyncDeliveryItem = nil
}

private func deliverAsyncSSEResponse(chunks: [Data], statusCode: Int) {
    let workItem = DispatchWorkItem { [weak self] in
        guard let self else { return }
        for chunk in chunks {
            Thread.sleep(forTimeInterval: chunkDelay)
            self.client?.urlProtocol(self, didLoad: chunk)
        }
        self.client?.urlProtocolDidFinishLoading(self)
    }
    asyncDeliveryItem = workItem
    DispatchQueue.global(qos: .default).async(execute: workItem)
}
```

Note: `DispatchWorkItem.cancel()` sets `isCancelled` on the work item; check `workItem.isCancelled` (not `Thread.current.isCancelled`) inside the block for correctness:

```swift
let workItem = DispatchWorkItem {
    for chunk in chunks {
        if workItem.isCancelled { break }  // use captured reference
        ...
    }
}
```

---

All other concerns verified clean: `capturedRequests` reset correctly between tests, `httpBodyStream` handling is correct, Keychain tests clean up with `defer`, MLX file is properly wrapped in `#if MLX`, and `isGenerating` is set synchronously so no race on the cancellation assertion.

PR #39, opened 2026-03-31
refactor: introduce BackendError protocol unifying InferenceError and CloudBackendError

roryford:
Adds a BackendError protocol that both InferenceError and CloudBackendError conform to.

Catch sites in InferenceService can now use 'any BackendError' instead of handling two unrelated enums separately. No existing cases or signatures changed — purely additive.

Completes QA remediation plan Phase 2.

### DIFF
--- a/Sources/BaseChatCore/Protocols/BackendError.swift
+++ b/Sources/BaseChatCore/Protocols/BackendError.swift
@@ -3,7 +3,14 @@ import Foundation
 /// Common protocol adopted by all backend error types so that catch sites
 /// in InferenceService and the UI layer can handle errors uniformly without
 /// knowing whether the failure came from a local or cloud backend.
-public protocol BackendError: LocalizedError, Sendable {}
+///
+/// The `isRetryable` property lets call sites decide whether to surface a
+/// transient error with a retry prompt or treat it as permanent.
+public protocol BackendError: LocalizedError, Sendable {
+    /// Whether the error is transient and the operation may be retried
+    /// without any configuration change.
+    var isRetryable: Bool { get }
+}
 
 extension InferenceError: BackendError {}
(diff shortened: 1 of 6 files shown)


roryford (commented):
## Code Review Finding

### 🟠 `BackendError` protocol is unused — provides no catch-site benefit (Medium)

**File:** `Sources/BaseChatCore/Protocols/BackendError.swift`

The protocol's stated purpose is to let catch sites handle backend errors uniformly, but no catch site in the codebase actually uses it. Every existing handler still catches the concrete type:

```swift
// All catch sites today — specific types only
catch let error as CloudBackendError { ... }
catch let error as InferenceError { ... }
```

A search confirms zero instances of `catch let error as BackendError` or `catch let error as any BackendError` anywhere in `Sources/` or `Tests/`.

The conformance tests only verify that the conformance compiles — something the compiler already enforces — rather than demonstrating that the protocol enables any polymorphic behaviour.

**Suggested fix (pick one):**

**Option A — Add a unifying catch site in `InferenceService`** so the protocol earns its place:
```swift
// InferenceService.generate(...)
} catch let error as any BackendError {
    // unified path for both local and cloud backend failures
    throw error
}
```

**Option B — Add a required member** that makes catching the protocol actually useful:
```swift
public protocol BackendError: LocalizedError, Sendable {
    /// Whether the error is transient and the operation may be retried.
    var isRetryable: Bool { get }
}
```
Both `InferenceError` and `CloudBackendError` would implement this, giving catch sites something concrete to act on.

**Option C — Defer the protocol** until there is a real catch-site need, and remove it from this PR to keep the change focused.

The retroactive conformances themselves are technically correct (all types live in `BaseChatCore`), so there are no validity concerns — just the question of whether the protocol currently serves its stated purpose.

PR #90, opened 2026-04-04
fix: correct macOS sheet layouts and add model selection E2E tests

roryford:
## Summary

- **`ModelManagementSheet`**: `.navigationTitle` and `.toolbar` were applied to the `NavigationStack` itself instead of its content view. On macOS this caused the toolbar to render as a detached overlay covering the list, making all three tabs (Select, Download, Storage) appear blank.
- **`GenerationSettingsView`**: Added `.formStyle(.grouped)` so the `Form` uses the grouped list style instead of the macOS two-column layout, which was pushing `Picker` and `LabeledContent` labels outside their visual sections. Also renamed the redundant `"Backend"` `LabeledContent` label to `"Type"`.
- **`ModelStorageService`**: Added an optional `baseDirectory` parameter to `init` so tests can inject a temp directory without touching the real Documents folder.
- **`ModelSelectionE2ETests`**: 8 new E2E tests covering the full model selection pipeline.

## Test plan

- [ ] `swift test --filter BaseChatCoreTests` passes
- [ ] `swift test --filter BaseChatUITests` passes
- [ ] `swift test --filter BaseChatBackendsTests` passes
- [ ] `swift test --filter BaseChatE2ETests.ModelSelectionE2ETests` — all 8 tests pass
- [ ] Open model management sheet in the demo app — Select/Download/Storage tabs show content
- [ ] Open generation settings — no two-column label misalignment

### DIFF
--- a/Sources/BaseChatCore/Services/ModelStorageService.swift
+++ b/Sources/BaseChatCore/Services/ModelStorageService.swift
@@ -7,9 +7,12 @@ import Foundation
 public final class ModelStorageService {
 
     private let fileManager: FileManager
+    /// Overrides the default Documents-relative directory. Used in tests.
+    private let customDirectory: URL?
 
-    public init(fileManager: FileManager = .default) {
+    public init(fileManager: FileManager = .default, baseDirectory: URL? = nil) {
         self.fileManager = fileManager
+        self.customDirectory = baseDirectory
     }
 
     // MARK: - Directory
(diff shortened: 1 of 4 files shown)


roryford (commented):
Addressed: converted `ModelSelectionE2ETests` from `struct` to `final class` and added `deinit { cleanupE2ETempDir(modelsDir) }`. Swift Testing creates a new instance per test, so `deinit` fires after each test and removes the temp directory.

PR #258, opened 2026-04-10
feat: add VoiceOver contract and a11y tests for chat UI

roryford:
## Summary

- Adds [ViewInspector 0.10.3](https://github.com/nalexn/ViewInspector/releases/tag/0.10.3) as a test-only dependency (scoped to `BaseChatUITests` — does **not** appear in any production target).
- Introduces `ChatA11yContractTests`, the first accessibility-contract suite for chat UI surfaces. The existing `BaseChatSnapshotTests` suite uses `Swift.dump()`, which strips `.accessibilityLabel`, so the chat UI previously had zero automated a11y coverage.
- Tightens source-level accessibility modifiers on the surfaces the tests cover:
  - `MessageBubbleView` now emits `"User said: …"` / `"Assistant said: …"` / `"System said: …"` (previously leaked the raw enum `rawValue` — `"user: …"`). A `public static accessibilityLabel(for:)` helper exposes the contract for direct assertion.
  - `ContextIndicatorView` label rewritten to `"Context used: X of Y tokens"` with a separate `accessibilityValue` of `"NN percent"`, matching VoiceOver's label/value conventions.
  - The chat error banner is extracted into a standalone `ErrorBannerView` carrying `.accessibilityAddTraits(.isHeader)` and a contract-built `"Error: <message>"` label so VoiceOver announces it as a landmark.
  - The chat input primary action is extracted into a `SendStopButton` subview with public `sendLabel` / `stopLabel` constants, enabling deterministic inspection without mounting a `ChatViewModel` environment.
- The `Swift.dump()`-based snapshot tests are intentionally **not migrated** — that is out of scope.

Includes a sabotage check per CLAUDE.md conventions: the `expectedMessageBubbleLabel` helper is duplicated in the test file (not imported), and I verified locally that reverting `MessageBubbleView.accessibilityLabel(for:)` to the old `"\(role.rawValue): \(content)"` format fails 6 assertions in the new suite before reverting.

## Release Note

**Accessibility contract tests and VoiceOver polish for chat UI** — The chat view's accessibility surface had no automated coverage because the existing snapshot harness captured view hierarchies via `Swift.dump()`, which strips accessibility labels. This change adds a ViewInspector-driven `ChatA11yContractTests` suite and tightens the labels VoiceOver users actually hear: message bubbles now announce `"User said: …"` / `"Assistant said: …"` instead of the raw enum rawValue, the context indicator reads `"Context used: 1234 of 4096 tokens"` at `"50 percent"`, and the error banner is exposed as an accessibility header with an `"Error: …"` prefix so screen-reader users can orient to it as a landmark. Existing visual snapshots and behaviour are unchanged.

## Test plan

- [x] `swift package resolve` confirms ViewInspector 0.10.3 pins cleanly
- [x] `swift test --filter BaseChatCoreTests --disable-default-traits` — 105 tests pass
- [x] `swift test --filter BaseChatUITests --disable-default-traits` — 402 tests pass (includes 11 new `ChatA11yContractTests`)
- [x] `swift test --filter BaseChatBackendsTests --disable-default-traits` — 84 tests pass
- [x] Sabotage check: reverted `MessageBubbleView.accessibilityLabel(for:)` to the old format and confirmed 6 a11y assertions fail before reverting
- [ ] CI green on `BaseChatCoreTests` + `BaseChatUITests` + `BaseChatBackendsTests`

## Notes for reviewers

- ViewInspector cannot synchronously walk a view that reads `@Environment(ChatViewModel.self)` (the new `@Observable` injection path is not supported by ViewInspector's reflective inspector). Rather than refactor the whole input bar or add a flaky async hosted test, I extracted the minimal subviews (`SendStopButton`, `ErrorBannerView`) so tests can inspect them directly with zero mocking. `ChatInputBar` and `ChatView` now compose these subviews — the call sites remain private.
- The existing `BaseChatSnapshotTests` suite is untouched.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,5 +1,5 @@
 {
-  "originHash" : "a5e57cdcaafddaa4c1ea15543f8a0abd1f24aa357b532d41f61ab43435cf85b1",
+  "originHash" : "9b3003493e53504564187264af527724bbc09f7870985e3836747d2324f7ab98",
   "pins" : [
     {
       "identity" : "eventsource",
@@ -151,6 +151,15 @@
       "state" : {
         "revision" : "b38443e44d93eca770f2eb68e2a4d0fa100f9aa2",
         "version" : "1.3.0"
+      }
+    },
+    {
+      "identity" : "viewinspector",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/nalexn/ViewInspector",
+      "state" : {
+        "revision" : "e9a06346499a3a889165647e3f23f8a7b2609a1c",
+        "version" : "0.10.3"
       }
     },
     {
--- a/Package.swift
+++ b/Package.swift
@@ -32,6 +32,9 @@ let package = Package(
         .package(url: "https://github.com/huggingface/swift-transformers", from: "1.2.0"),
         .package(url: "https://github.com/mattt/llama.swift", from: "2.8672.0"),
         .package(url: "https://github.com/pointfreeco/swift-snapshot-testing", from: "1.17.0"),
+        // Test-only: SwiftUI view-tree inspection for accessibility contract tests.
+        // Must never appear in any production target.
+        .package(url: "https://github.com/nalexn/ViewInspector", from: "0.10.3"),
     ],
     targets: [
         // Inference: models, protocols, services — no SwiftData, no heavy ML deps
@@ -112,7 +115,13 @@ let package = Package(
         ),
         .testTarget(
             name: "BaseChatUITests",
-            dependencies: ["BaseChatUI", "BaseChatCore", "BaseChatInference", "BaseChatTestSupport"]
+            dependencies: [
+                "BaseChatUI",
+                "BaseChatCore",
+                "BaseChatInference",
+                "BaseChatTestSupport",
+                .product(name: "ViewInspector", package: "ViewInspector"),
+            ]
         ),
         .testTarget(
             name: "BaseChatE2ETests",
--- a/Sources/BaseChatUI/Views/Chat/ChatInputBar.swift
+++ b/Sources/BaseChatUI/Views/Chat/ChatInputBar.swift
@@ -75,31 +75,12 @@ public struct ChatInputBar: View {
 
     @ViewBuilder
     private var sendOrStopButton: some View {
-        if viewModel.isGenerating {
-            Button {
-                viewModel.stopGeneration()
-            } label: {
-                Image(systemName: "stop.circle.fill")
-                    .font(.title2)
-                    .foregroundStyle(.red)
-            }
-            .buttonStyle(.plain)
-            .accessibilityLabel("Stop generation")
-            .help("Stop generation")
-        } else {
-            Button {
-                sendMessage()
-            } label: {
-                Image(systemName: "arrow.up.circle.fill")
-                    .font(.title2)
-                    .foregroundStyle(canSend ? Color.accentColor : Color.gray)
-            }
-            .buttonStyle(.plain)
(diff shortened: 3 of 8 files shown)


roryford (commented):
## Review

Overall the PR is well-structured and the intent is solid. One real bug was found and fixed in a follow-up commit on this branch (`a2314a8`). Everything else checks out.

---

### What passes ✅

- **ViewInspector scoping** — `ViewInspector` appears only in `BaseChatUITests` in `Package.swift`. No production target pulls it in.
- **`BaseChatUI` dependency wall** — None of the changed files import `BaseChatBackends`. The `BaseChatUI` → `BaseChatCore`-only rule is intact.
- **Accessibility label format** — `MessageBubbleView.accessibilityLabel(for:)` correctly produces `"User said: …"` / `"Assistant said: …"` / `"System said: …"`. The static helper is `public` and tested both directly and via ViewInspector tree-walking.
- **`ContextIndicatorView`** — Label `"Context used: X of Y tokens"` and value `"NN percent"` are correctly applied and tested.
- **`SendStopButton` constants** — `sendLabel` and `stopLabel` are defined as `static let` on `SendStopButton`. They are `internal` (not `public`), which is correct given the struct itself is internal-only. The PR description calling them "public" is a minor inaccuracy in prose, not a code issue.
- **Sabotage check** — Documented in `test_sabotage_wrongFormatFailsContract()` with a clear comment, and referenced in the PR body. Meets the CLAUDE.md requirement.
- **All 11 new a11y tests pass** — Confirmed locally.
- **Full CI suite** — `BaseChatCoreTests` (107), `BaseChatUITests` (411), `BaseChatBackendsTests` (84) — zero failures after the fix.

---

### Bug found and fixed ❌ → ✅

**`ErrorBannerView`: `.accessibilityElement(children: .contain)` silently discards the label and trait**

In `ChatView.swift`, `ErrorBannerView.body` applied:

```swift
.accessibilityElement(children: .contain)
.accessibilityLabel(Self.accessibilityLabel(for: error))
.accessibilityAddTraits(.isHeader)
```

`.contain` exposes each child as an individual VoiceOver element and **ignores** any `.accessibilityLabel()` or `.accessibilityAddTraits()` set on the container. The `"Error: …"` header label and the `.isHeader` trait would never be announced to VoiceOver users at runtime, even though ViewInspector correctly reads the modifier in tests (ViewInspector walks the SwiftUI modifier tree, not the computed accessibility element the OS actually exposes).

Fixed by changing `.contain` to `.combine`, which merges the container's children into a single accessible element and honours the label and trait overrides:

```swift
// Use .combine so that the container itself becomes the VoiceOver element.
// With .contain the label/trait modifiers below would be silently ignored —
// .contain exposes children individually and discards container-level overrides.
.accessibilityElement(children: .combine)
.accessibilityLabel(Self.accessibilityLabel(for: error))
.accessibilityAddTraits(.isHeader)
```

The `test_errorBanner_hasAccessibilityHeaderAndLabel` test continues to pass after the fix (all 411 UITests green). Fix is in commit `a2314a8` on this branch.

---

### Minor observations (not blocking)

- The `ErrorBannerView` test does not assert the `.isHeader` trait presence (ViewInspector supports `.accessibilityTraits()` inspection). Worth a follow-up to fully cover the contract, but not a blocker.
- `SendStopButton` and `ErrorBannerView` are `internal` — the PR summary describes them as having "public" constants/traits. A one-word prose correction in the PR body would help future readers, but this doesn't affect runtime behaviour.

---

**Verdict**: The `.contain` → `.combine` fix is the only code change required and has been pushed. CI should be green. Ready to merge once CI confirms.

PR #259, opened 2026-04-10
test: add day-one user journey E2E coverage

roryford:
## Summary

Adds a single end-to-end test that walks a first-time user through the full day-one journey in one `ChatViewModel` instance: empty state, model discovery, load, session creation, first turn, mid-stream cancellation, regeneration, model switch, a second turn against the new model, and session persistence restore.

The existing one-feature-per-test coverage in `BaseChatE2ETests/` validates each step in isolation. This new test exists specifically to catch bugs that only surface when these features run back-to-back against the same view model — hand-offs between discovery, load, send, cancel, regenerate, switch, and persistence.

- Uses `SlowMockBackend` and in-memory SwiftData so step 5 can deterministically observe mid-stream cancellation (per-token delay gives `stopGeneration()` real work to cancel).
- No hardware dependencies; runs in the `BaseChatE2ETests --disable-default-traits` suite.
- Includes a sabotage-check assertion on the final reload, verified to fail when the step-7 `saveSettingsToSession()` call is removed.

## Findings

While writing this test I hit a latent bug in the send-then-cancel path: both `ChatViewModel.stopGeneration()` and the tail of `generateIntoMessage()` call `saveMessage()` on the same assistant record, which double-inserts a SwiftData row (same `id`, two rows). The test works around it by deleting the duplicate row directly via `ModelContext` before proceeding to the regenerate step, with a comment flagging the issue. Fixing the underlying bug is out of scope for this test-only PR and should be tracked separately.

## Test plan

- [x] `swift test --filter BaseChatE2ETests.UserJourneyE2ETests --disable-default-traits` passes locally
- [x] `swift test --filter BaseChatCoreTests --disable-default-traits` passes
- [x] `swift test --filter BaseChatUITests --disable-default-traits` passes
- [x] `swift test --filter BaseChatBackendsTests --disable-default-traits` passes
- [x] `swift test --filter BaseChatE2ETests --disable-default-traits` passes (60 tests)
- [x] Sabotage check: commenting out the step-7 `saveSettingsToSession()` call fails the sabotage assertion

### DIFF
--- a/Tests/BaseChatE2ETests/UserJourneyE2ETests.swift
+++ b/Tests/BaseChatE2ETests/UserJourneyE2ETests.swift
@@ -314,6 +314,12 @@ final class UserJourneyE2ETests {
         // the session record straight from SwiftData. This proves that
         // saveSettingsToSession actually reached the store in step 7 — not
         // just the in-memory sessions array that loadSessions() repopulates.
+        //
+        // Sabotage verified: removing the `try vm.saveSettingsToSession()` call
+        // in step 7 causes the assertion below to fail with
+        // `storedSessions.first?.selectedModelID == modelA.id` (or nil) instead
+        // of modelB.id, confirming this assertion catches the real persistence
+        // contract and is not accidentally vacuous.
         let storedSessionID = session.id
         let sessionDescriptor = FetchDescriptor<ChatSession>(
             predicate: #Predicate { $0.id == storedSessionID }


roryford (commented):
## Review: test/add-user-journey-e2e

Reviewed `Tests/BaseChatE2ETests/UserJourneyE2ETests.swift` (328 lines), ran all CI suites locally, and pushed one fix.

---

### What passes

**Test conventions**
- Uses `@Suite`/`@Test`/`final class` — consistent with the established E2E suite pattern (`ModelSelectionE2ETests` uses the same shape). CLAUDE.md says "use `XCTestCase` for new tests," but the E2E directory has settled on `@Suite`/`final class` with `init() throws` for per-test setup, which requires a class type. Following the local pattern is the right call.
- Real `async/await` throughout. No `XCTestExpectation` wrappers for async code.
- No hardware dependencies — `SlowMockBackend` + in-memory SwiftData, no `#if MLX`/`#if Llama` guards needed and none are present.
- No `assertionFailure`/`fatalError` in recoverable paths.

**`SlowMockBackend` configuration**
- `NSLock`-protected mutable state, `Task.isCancelled` checks both before and after the per-token sleep, and `onTermination` cancels the generation task. The 50ms delay matches `CancellationTests` convention. Mid-stream cancel in step 5 is deterministic: `awaitFirstToken` polls at 5ms intervals with a 3s deadline before calling `stopGeneration()`.

**Known-issue wrapping (bug #260)**
- Three `withKnownIssue` blocks cleanly isolate the duplicate-row fallout. The non-wrapped assertions (content anchors, in-memory sequence, SwiftData direct fetch) are all robust against the orphan row. When #260 is fixed the `withKnownIssue` wrappers will self-report "expected a known failure but didn't get one," which is the correct removal prompt.

**Issue filed**
- #260 is open with root cause, reproduction path, and a fix proposal.

**Test result**
```
Test "Day-one journey: ..." passed after 0.461 seconds with 3 known issues.
```
All three CI suites green: 107 core / 400 UI / 84 backends.

---

### Fix pushed (13c5775)

**Missing inline sabotage comment.** The PR body documented the sabotage check ("removing the step-7 `saveSettingsToSession()` call fails the final assertion") but the assertion site had no inline comment — the codebase convention (see `ChatA11yContractTests`, `CompressionIntegrationTests`) is to leave a `// Sabotage verified:` comment at the exact assertion that was tested. Added:

```swift
// Sabotage verified: removing the `try vm.saveSettingsToSession()` call
// in step 7 causes the assertion below to fail with
// `storedSessions.first?.selectedModelID == modelA.id` (or nil) instead
// of modelB.id, confirming this assertion catches the real persistence
// contract and is not accidentally vacuous.
```

---

- `awaitFirstToken` uses `Task.yield()` + `Task.sleep(for: .milliseconds(5))` polling. This is the same pattern used in `CancellationTests` and is fine given it's gating a real async operation with a hard deadline rather than timing out a synchronous state change.
- The step-4 `switchToSession` → re-apply `selectedModel` → `saveSettingsToSession()` sequence is correctly commented and reflects actual `switchToSession` behavior for nil-model sessions.

**Verdict: approve after the sabotage comment fix lands (already pushed).**

PR #263, opened 2026-04-10
feat: tool call approval flow with edit, reject, and trusted-tools modes

### LINKED ISSUE #247: feat: tool call approval flow and argument inspection UI
## Background

BaseChatKit-2 has solid tool-calling plumbing on the backend side — \`ToolProvider\`, \`ToolCallingBackend\`, \`ToolCallObserver\`, a 10-round execution limit, and UI rendering of tool call blocks as collapsible disclosures (\`MessagePartsView.swift:69-80\`). But the UI is **read-only**: tool calls execute automatically, arguments cannot be edited, and there's no approval gate.

For users running destructive tools (delete file, send email, make HTTP request), silent auto-execution is unacceptable.

## Proposal

1. Add a \`ToolCallApprovalMode\` setting: \`.alwaysAsk\`, \`.trustedTools(Set<String>)\`, \`.autoApprove\` (current behavior, opt-in).
2. When \`.alwaysAsk\`: the tool call bubble displays a **pending** state with \"Approve\" / \"Reject\" / \"Edit Arguments\" buttons. Execution pauses until the user decides.
3. **Edit Arguments**: a sheet showing the tool's JSON schema and current args, with validation before re-submit.
4. **Reject**: sends a synthetic \"user rejected tool call\" message back to the model so it can proceed without the tool result.
5. Default mode should be \`.alwaysAsk\` for remote (Claude, OpenAI) backends — safer default.
6. Persist approved/rejected tool calls in \`ChatMessageRecord\` so the history is faithful after reload.

## Backend wiring

\`InferenceService\` or \`ToolCallingBackend\` needs a checkpoint between \"tool call emitted\" and \"tool executed\" that can await user decision. This may require introducing an async approval callback in the tool execution path.

## Acceptance

- [ ] \`ToolCallApprovalMode\` setting + UI toggle
- [ ] Approval gate in the tool execution path
- [ ] Pending / approved / rejected states in MessagePartsView
- [ ] Edit arguments sheet with schema validation
- [ ] Rejected calls faithfully persisted
- [ ] Tests for each mode
- [ ] Pre-push full suite passes

## Related

Listed as a \"Rough edge: tool calls are read-only\" in the audit feature inventory.

roryford:
## Summary

- Add `ToolCallApprovalCoordinator` on `InferenceService` that gates every tool call on a user decision before execution. Three modes: `alwaysAsk` (default), `trustedTools(Set<String>)`, and `autoApprove`.
- `ApprovingToolProvider` transparently wraps the host's `ToolProvider` so no backend has to change — existing `ToolCallingBackend` adopters continue to work.
- `MessagePart.toolCall` now carries a `ToolCallApprovalState` (`pending` / `approved` / `edited` / `rejected`). Custom `Codable` decodes pre-existing history as `.approved` so nothing in old chat logs has to migrate.
- `MessagePartsView` renders Approve / Reject / Edit buttons when a call is pending, and a new `EditToolArgumentsSheet` shows the tool's JSON schema alongside an editable arguments field with JSON validity checking.
- Rejection synthesises an `isError: true` `ToolResult` with the user's reason so the model can read "user rejected this" and plan around it.

Closes #247

## Release Note

**Users can now approve or reject every tool call before it runs.** Tool calling in BaseChatKit used to execute automatically the moment the model emitted a call, which made it unsafe to wire up tools with real side effects — a single unlucky `delete_file` or `send_email` would run before the user could blink. The new approval flow pauses each tool call until the user explicitly decides what to do, with three policies apps can choose between: `alwaysAsk` (the new default, safe by construction), `trustedTools` (auto-allow a named allowlist, prompt for everything else), and `autoApprove` (the old behaviour, now opt-in). Pending calls show inline Approve / Reject / Edit controls in the chat bubble, and the edit sheet renders the tool's JSON schema next to an editable arguments field so users can inspect exactly what is about to run. Rejections are sent back to the model as a synthetic error result so the conversation keeps flowing without a dangling tool-call round. Approval states persist across reloads, so users who come back to an old session see faithfully whether each tool call ran, was edited, or was turned down.

## Test plan

- [x] `BaseChatCoreTests` — 713 tests pass (adds 16: 9 `ToolCallApprovalTests`, 7 `ToolCallPersistenceTests`)
- [x] `BaseChatUITests` — 391 tests pass
- [x] `BaseChatBackendsTests` — 84 tests pass (no hardware traits)
- [x] `swift build --traits MLX,Llama` — clean on Apple Silicon
- [x] Sabotage verify: flipped `.alwaysAsk` to return `true` from `allowsAutoApproval(of:)`, confirmed 4 approval tests failed; reverted.
- [ ] Maintainer to visually verify the Approve/Reject/Edit bubble state and EditToolArgumentsSheet in a running host app (local tests do not render SwiftUI).

## New UI states

- **Pending bubble**: orange-tinted tool call block with `Pending approval` capsule and three inline controls (`Approve`, `Edit`, `Reject`). The disclosure still expands to show the raw arguments.
- **Edit sheet**: title, tool name, description, collapsible schema listing per-parameter type and required-ness, `TextEditor` with monospaced JSON and a live "not valid JSON" inline error.
- **Rejected bubble**: red-tinted with `Rejected` capsule; no buttons.
- **Edited bubble**: blue-tinted with `Edited` capsule indicating the user modified the arguments before letting it run.
- **Approved bubble**: unchanged from existing appearance.

## Notes / follow-ups

- The edit sheet validates that arguments are well-formed JSON, not that they match the `ToolInputSchema`. Full JSON Schema validation (type checks, enum matching, required key enforcement) needs a real validator — not pulling a dependency for this PR, the current gate is an explicit minimum-viable guard. A follow-up issue can add schema-aware validation.
- Tool calling is currently plumbed into `ClaudeBackend` and `OpenAIBackend` at the payload-parsing layer only — neither one yet runs the full multi-round execution loop. The approval gate is wired at `InferenceService` so it is ready the moment a backend starts calling `execute(_:)` on its (now wrapped) provider.

### DIFF
--- a/Sources/BaseChatCore/Models/MessagePart.swift
+++ b/Sources/BaseChatCore/Models/MessagePart.swift
@@ -1,5 +1,27 @@
 import Foundation
 
+/// Lifecycle state of a tool call persisted inside a message bubble.
+///
+/// Stored alongside the ``MessagePart/toolCall(id:name:arguments:state:)``
+/// payload so history reloads can faithfully render whether a tool call is
+/// still waiting, was run, was rejected, or had its arguments edited before
+/// execution.
+public enum ToolCallApprovalState: String, Codable, Hashable, Sendable {
+
+    /// The model requested this call but the user has not yet decided.
+    case pending
+
+    /// The call ran (either auto-approved or explicitly approved).
+    case approved
+
+    /// The user approved execution after editing the arguments.
+    case edited
+
+    /// The user explicitly rejected the call. A synthetic rejection result
+    /// was fed back to the model.
+    case rejected
+}
+
 /// A discrete piece of content within a chat message.
 ///
 /// Messages can contain multiple parts to support multimodal input (images),
@@ -7,10 +29,16 @@ import Foundation
 /// independently typed so the UI can render appropriate controls (e.g.,
 /// inline images, collapsible tool call blocks) and backends can map parts
 /// to their native message formats.
-public enum MessagePart: Codable, Hashable, Sendable {
+public enum MessagePart: Hashable, Sendable {
     case text(String)
     case image(data: Data, mimeType: String)
-    case toolCall(id: String, name: String, arguments: String)
+
+    /// A tool invocation. `state` defaults to `.approved` to keep sources
+    /// that do not pass the argument compiling, and to match the
+    /// pre-approval-gate historical behaviour where every tool call that
+    /// landed in history had already executed.
+    case toolCall(id: String, name: String, arguments: String, state: ToolCallApprovalState = .approved)
+
     case toolResult(id: String, content: String)
 }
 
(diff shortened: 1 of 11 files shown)


roryford:
Closing as part of the BaseChatKit slimming initiative. We audited two consumer applications and found this feature had no usage in either. The initiative focuses BCK on its operational-reliability moat (streaming resilience, model handoff, memory pressure, mock backend, cert pinning) rather than expanding provider/feature surface.

For the full scoping rationale see [docs/SCOPE_DECISION.md](https://github.com/roryford/BaseChatKit/blob/main/docs/SCOPE_DECISION.md).

If you have a concrete use case, please reopen with details and we'll reconsider.

PR #265, opened 2026-04-10
feat: add systemPromptContext API for simple system prompt substitution

roryford:
## Summary
- Adds `public var systemPromptContext: [String: String]` to `ChatViewModel`
- Adds a substitution pass in `ChatViewModel+Generation.swift` at the existing macro expansion site, running AFTER `MacroExpander.expand`
- Adds unit tests covering substitution, empty-dict no-op, and MacroExpander-wins-on-collision ordering

## Why
Consumer audit showed ChatbotUI-iOS only uses `viewModel.macroContext.userName` — it doesn't need the full MacroExpander/MacroProvider registry. This PR adds a simpler dict-based API for that case. The full MacroExpander stays in place because Fireside's LorebookTriggerEngine actively depends on it. Week 2 will coordinate the MacroExpander deletion with Fireside taking ownership.

Non-breaking. Both `macroContext` and `systemPromptContext` coexist in 0.6.0.

## Test plan
- [ ] `swift test --filter BaseChatUITests --disable-default-traits` — new tests pass
- [ ] Existing `ChatViewModelMacroExpansionTests` still pass (no ordering regression)
- [ ] Full CI suite green

## Release Note
**New simpler system-prompt substitution API** — `ChatViewModel.systemPromptContext: [String: String]` lets apps substitute `{{key}}` tokens in the system prompt without wiring up a full MacroProvider. Runs after MacroExpander so existing Fireside-style expansions continue to work unchanged. Both APIs coexist.


roryford:
## Review pass — applied fixes

Picked up the three Copilot review comments and added the missing edge-case test coverage. Also rebased on `main` so CI's `SilentCatchAuditTest` regression (allowlist drift from #257 + #266 → #262) is resolved.

### Copilot comments
| # | Comment | Action |
|---|---------|--------|
| 1 | `systemPromptContext` silently ignored when `macroExpansionEnabled = false` | **Applied** — substitution now runs unconditionally. The `if macroExpansionEnabled` only gates the full `MacroExpander.expand` pass; the dict-based fill always runs. New test `runs_whenMacroExpansionDisabled` locks this in. |
| 2 | Sequential `replacingOccurrences` per key allows non-deterministic cascading replacements | **Applied** — replaced the per-key loop with a single regex pass over `{{word}}` tokens (same shape as `MacroExpander`). Non-recursive and dict-iteration-order-independent. New test `doesNotRecursivelyExpandValues` proves it. |
| 3 | Doc comment doesn't mention `macroExpansionEnabled` coupling | **Superseded** by fix 1 — coupling no longer exists. Doc comment rewritten anyway to spell out ordering, unconditional behavior, non-recursive semantics, missing-key pass-through, and the `{{word}}` token shape constraint. |

### Additional product-engineer findings I addressed
- **Missing-token behavior wasn't tested** — added `missingKey_leavesTokenVerbatim`. This is a feature (callers can spot unfilled tokens), so it needs a test that locks it in.
- **Multiple occurrences of same token** — added `replacesAllOccurrencesOfSameToken`.
- **Empty-string value** — added `emptyStringValue_substitutesAsEmpty`. `replacingOccurrences` would have handled this fine, but worth pinning.
- **Mutation between sends** — added `mutationBetweenSends_appliesToSecondPrompt`. Confirms there's no captured snapshot anywhere.

### Findings I deliberately did not act on
- **`[String: String]` vs `[String: CustomStringConvertible]`** — keeping `[String: String]` matches the doc-comment goal of "simple key/value substitution." Forcing the caller to convert is one line and avoids the protocol-existential overhead.
- **`{{key with spaces}}` / `{{}}` / Unicode normalization** — the regex is `\w+`, so non-word tokens simply don't match and pass through. Documented in the doc comment. Not worth a separate validator.
- **Macro registry test flake (`MacroExpanderTests` "Nil userName ... → Interloper")** — pre-existing on `main`; the global `MacroExpander.registry` isn't reset between Swift Testing parallel suites. Out of scope for this PR.

### CI fix
The `SilentCatchAuditTest` failure was unrelated to this PR — `main` had already added the `ChaosBackend`/`PerceivedLatencyBackend` allowlist entries that the audit test was missing. Merging `main` brought the fix in. Also added one new allowlist entry for the `try? NSRegularExpression(...)` in the new substitution helper, mirroring the existing `MacroExpander` entries.

### Pre-push test results
- `BaseChatCoreTests --disable-default-traits` — **758 / 758 pass**
- `BaseChatUITests --disable-default-traits` — **408 / 408 pass** (includes 10/10 in `ChatViewModelSystemPromptContextTests`)
- `BaseChatBackendsTests --disable-default-traits` — **152 / 152 pass**

Sabotage-verified the new tests by stubbing out the substitution helper — 9 of 10 fail (the empty-dict no-op correctly stays green), confirming the assertions are load-bearing.

PR #348, opened 2026-04-13
fix: enforce SSECloudBackend overrides at compile time (#328)

### LINKED ISSUE #328: fix: SSECloudBackend abstract methods crash at runtime instead of failing at compile time
## Problem

Three methods on \`SSECloudBackend\` use \`fatalError("Subclasses must override ...")\` as Swift's workaround for abstract methods:

- \`capabilities\` (line 125)
- \`buildRequest(prompt:systemPrompt:config:)\` (line 137)
- \`extractToken(from:)\` (line 144)

All three existing subclasses (\`OpenAIBackend\`, \`ClaudeBackend\`, \`OllamaBackend\`) override these correctly, so this is not a current bug. But external consumers subclassing \`SSECloudBackend\` for a custom cloud provider discover missing overrides at runtime (first generation attempt), not compile time.

## Fix options

1. **Quick fix:** Add \`@available(*, unavailable, message: "Subclasses must override")\` to each method to produce a compile-time diagnostic.
2. **Structural fix:** Replace subclassing with a \`SSEPayloadHandler\` protocol + composition — subclasses provide a handler strategy, \`SSECloudBackend\` becomes a final concrete type that delegates to it. This pattern is already partially present with the payload handler types.

## Blast radius

\`SSECloudBackend\` only. Option 1 is source-compatible; option 2 is a breaking change for external subclasses.

roryford:
## Summary

Fixes #328. Three methods on `SSECloudBackend` used `fatalError("Subclasses must override …")` as a runtime stand-in for abstract methods. External consumers who subclassed `SSECloudBackend` without overriding `extractToken(from:)`, `extractUsage(from:)`, `isStreamEnd(_:)`, or `extractStreamError(from:)` would crash at runtime with no compiler warning.

## What changed

`SSECloudBackend.init` now requires a `payloadHandler: any SSEPayloadHandler` parameter. The four previously-fatal methods delegate to this handler by default. Missing implementations now fail at compile time — the init call won't compile without a handler.

`OpenAIBackend`, `ClaudeBackend`, and `OllamaBackend` each pass their existing `*PayloadHandler` structs to the new parameter. The redundant `override` methods that simply forwarded to the same static functions are removed (the base-class delegation is equivalent). `SecureBytesTests` gains a private `NoOpPayloadHandler` to satisfy the updated init.

`capabilities` and `buildRequest(prompt:systemPrompt:config:)` remain `open` with improved `fatalError` messages (including `type(of: self)`) — a protocol-based solution for those two is tracked separately.

## Breaking change

`SSECloudBackend.init(defaultModelName:urlSession:)` gains a required `payloadHandler:` parameter. Any code that subclasses `SSECloudBackend` and calls `super.init(defaultModelName:urlSession:)` must be updated to pass a handler. This is the point of the fix — the compiler now enforces what was previously a runtime crash.

## Test plan

- [x] `swift test --filter BaseChatBackendsTests --disable-default-traits` — 63 tests, 0 failures
- [x] `swift test --filter BaseChatInferenceTests --disable-default-traits` — 69 tests, 0 failures
- [x] `swift test --filter BaseChatCoreTests --disable-default-traits` — 83 tests, 0 failures
- [x] `swift test --filter BaseChatUITests --disable-default-traits` — 424 tests, 0 failures

## Release Note

**SSECloudBackend no longer crashes silently when subclasses miss payload methods.** Previously, forgetting to override `extractToken(from:)` or related SSE methods on a custom `SSECloudBackend` subclass would cause a runtime `fatalError` that only appeared when generation ran. The base class initialiser now requires an `SSEPayloadHandler` — the compiler rejects backends that omit it, turning a runtime crash into a compile error. Built-in backends (`OpenAIBackend`, `ClaudeBackend`, `OllamaBackend`) are unaffected by the runtime behaviour change; their existing `*PayloadHandler` structs are now passed at init instead of being wired through `override` methods.

### DIFF
--- a/Sources/BaseChatBackends/ClaudeBackend.swift
+++ b/Sources/BaseChatBackends/ClaudeBackend.swift
@@ -105,24 +105,6 @@ public final class ClaudeBackend: SSECloudBackend, TokenUsageProvider, CloudBack
         return request
     }
 
-    // MARK: - SSE Payload Handling
-
-    public override func extractToken(from payload: String) -> String? {
-        Self.parseToken(from: payload)
-    }
-
-    public override func extractUsage(from payload: String) -> (promptTokens: Int?, completionTokens: Int?)? {
-        Self.parseUsage(from: payload)
-    }
-
-    public override func isStreamEnd(_ payload: String) -> Bool {
-        Self.parseIsStreamEnd(payload)
-    }
-
-    public override func extractStreamError(from payload: String) -> Error? {
-        Self.parseStreamError(from: payload)
-    }
-
     /// Claude reports usage split across `message_start` (prompt) and
     /// `message_delta` (completion), so we merge incrementally.
     public override func handleUsage(_ usage: (promptTokens: Int?, completionTokens: Int?)) {
--- a/Sources/BaseChatBackends/OllamaBackend.swift
+++ b/Sources/BaseChatBackends/OllamaBackend.swift
@@ -186,29 +186,6 @@ public final class OllamaBackend: SSECloudBackend, CloudBackendURLModelConfigura
         }
     }
 
-    // MARK: - SSE Payload Handler
-
-    /// Ollama-specific SSE payload handler.
-    ///
-    /// Ollama uses NDJSON rather than SSE, so ``parseResponseStream(bytes:continuation:)``
-    /// is overridden to bypass the SSE path entirely. This handler satisfies the
-    /// ``SSECloudBackend`` initialiser contract and delegates to the same static
-    /// parsing logic used by the NDJSON path.
-    struct OllamaPayloadHandler: SSEPayloadHandler {
-        func extractToken(from payload: String) -> String? {
-            OllamaBackend.extractToken(from: payload)
-        }
-        func extractUsage(from payload: String) -> (promptTokens: Int?, completionTokens: Int?)? { nil }
-        func isStreamEnd(_ payload: String) -> Bool { false }
-        func extractStreamError(from payload: String) -> Error? { nil }
-    }
-
-    // MARK: - SSE Hooks (unused for NDJSON, but required by base class)
-
-    public override func extractToken(from payload: String) -> String? {
-        Self.extractToken(from: payload)
-    }
-
     // MARK: - NDJSON Parsing
 
     /// Extracts the assistant content token from an Ollama `/api/chat` NDJSON line.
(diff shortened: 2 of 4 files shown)


roryford (commented):
Reviewed the full diff and read all changed files. The refactor is correct and clean — leaving detailed notes here since self-approval is blocked.

**Correctness — all four methods covered**

All three backends (`ClaudeBackend`, `OpenAIBackend`, `OllamaBackend`) correctly wire their `*PayloadHandler` structs to the new `payloadHandler` init parameter. Each handler delegates to the same static parse methods (or returns the same defaults) that the old `override` methods provided:

- `ClaudePayloadHandler` → `ClaudeBackend.parseToken/parseUsage/parseIsStreamEnd/parseStreamError` (identical to the removed overrides)
- `OpenAIPayloadHandler` → `OpenAIBackend.parseToken/parseUsage`, `isStreamEnd: false`, `extractStreamError: nil` (matching the old base-class defaults that OpenAI relied on)
- `OllamaPayloadHandler` → `OllamaBackend.extractToken`, stubs for the rest (Ollama bypasses the SSE path entirely via its `parseResponseStream` override, so these are never called in production)

`ClaudeBackend.handleUsage` (incremental prompt/completion merge) is untouched and still overrides correctly.

**No dead code left behind**

`static let payloadHandler` in `OpenAIBackend` and `ClaudeBackend` look like they could be dead code post-refactor, but they're actively used in `SSEPayloadReplayTests` and `OpenAICompatEndpointTests` — correctly retained.

`WeakBox` is still in use for the GenerationStream chicken-and-egg setup in `generate()` — not related to this PR.

**`import BaseChatInference` in `ClaudeBackend`**

Not redundant. `BaseChatBackends` depends directly on `BaseChatInference` in `Package.swift`, and `ClaudeBackend` uses inference types. The SourceKit diagnostic is an IDE artifact — the import is required and correct.

**API design**

`any SSEPayloadHandler` is the right spelling here. A generic parameter on `SSECloudBackend` would force every subclass to carry the type, and `some` can't be used for stored properties in open classes. The existential is appropriate.

**Breaking change**

Clearly documented in the PR body. The forced init breakage is exactly the point — turns a runtime crash into a compile error.

**One minor nit (non-blocking)**

`SecureBytesTests.swift` uses `@testable import BaseChatInference` — `@testable` is unnecessary since `SSEPayloadHandler` is `public`. No functional impact; happy to leave it as-is.

Overall: the refactor is mechanically correct, behavior is preserved across all three backends, and the public surface change is intentional and well-documented. Ready to merge.

PR #349, opened 2026-04-13
fix: move BackgroundDownloadManager resume data out of UserDefaults (#330)

### LINKED ISSUE #330: fix: BackgroundDownloadManager stores resume data in UserDefaults
## Problem

\`BackgroundDownloadManager\` persists two things in \`UserDefaults\`:

1. **Pending download metadata** (\`savePendingDownload\`, line 638) — a \`[String: [String: String]]\` dictionary.
2. **Resume data** (\`persistResumeData\`, line 621) — raw \`Data\` stored under \`resumeData.\(id)\` keys.

UserDefaults is backed by a plist and is not designed for large binary blobs. Resume data for a partially-downloaded model can be several megabytes. Issues:

- No hard storage cap, but large UserDefaults plist delays app launch synchronous read.
- No transactional guarantees — a crash between write and acknowledgement can corrupt the plist.
- No cleanup mechanism for resume data from crashed downloads — stale entries accumulate indefinitely.
- Read-modify-write in \`savePendingDownload\` is not atomic (only MainActor-isolated, not locked).

## Fix

Move persistence to files in the app's caches or support directory:
- One JSON file for pending download metadata (atomic write via temp-file + rename).
- One binary file per download for resume data, keyed by download ID.
- Add a cleanup sweep on launch that removes resume data for downloads no longer in the pending list.

## Blast radius

Internal to \`BackgroundDownloadManager\`. Requires a one-time migration for existing persisted downloads on first launch after the fix.

roryford:
## Summary

Fixes #330.

`BackgroundDownloadManager` previously stored two kinds of data in `UserDefaults`:
- **Pending download metadata** — a `[String: [String: String]]` dict under `<bundle>.pendingDownloads`
- **Resume data blobs** — raw `Data` under `resumeData.<id>` keys, one per in-flight download

Resume data for a partial model download can be several MB. Writing it to UserDefaults bloats the app's plist, which is loaded synchronously at every app launch. A multi-GB model interrupted mid-download produces ~20–50 MB of resume data that sits in UserDefaults until the retry succeeds — slowing every subsequent launch.

## What changed

- **Resume data** is now written to a binary file per download in `Caches/<bundle>.downloads/resume-<encoded-id>.bin` using `Data.write(.atomic)`.
- **Pending metadata** is now a single JSON file (`Caches/<bundle>.downloads/pending-downloads.json`) written via temp-file + `FileManager.moveItem` for atomic replacement, avoiding plist corruption on crash.
- **Orphan cleanup sweep** runs on every `reconnectBackgroundSession()` call: any resume-data file whose ID is not in the current pending-metadata list is deleted, preventing indefinite accumulation from crashed sessions.
- **One-time migration** reads the legacy UserDefaults keys on first launch after upgrade, moves the data to files, then deletes the old keys.

All four CI test suites pass (640 + 83 + 424 + 131 tests, zero failures). The `SilentCatchAuditTest` allowlist was updated for the seven new `try?` call sites — all are legitimate best-effort reads (file absent = no pending downloads) or best-effort cleanups (stale resume file removal).

## Release Note

**Partial-download resume data no longer lives in UserDefaults.** Previously, the bytes the system preserved when a model download was interrupted (resume data) were stored in `UserDefaults` — the same location that iOS loads synchronously at every app launch. A multi-GB model paused mid-download could leave 20–50 MB in the plist, measurably slowing cold start. Pending-download metadata was also at risk of plist corruption if the app crashed between a `set` and the system's write-back.

Both are now file-based: resume data lives in one binary file per download in the Caches directory (written atomically), and pending metadata is a single JSON file replaced via temp-file rename. A cleanup sweep on launch removes orphaned resume-data files from previously crashed sessions. A one-time migration moves any existing UserDefaults data to the new location the first time the updated app runs.

### DIFF
--- a/Sources/BaseChatInference/Services/BackgroundDownloadManager.swift
+++ b/Sources/BaseChatInference/Services/BackgroundDownloadManager.swift
@@ -120,14 +120,18 @@ public final class BackgroundDownloadManager: NSObject, @unchecked Sendable {
     /// resume-data files. Using Caches lets the OS reclaim space under extreme pressure,
     /// which is acceptable — a missing resume-data file causes a fresh download, not a
     /// crash or data corruption.
-    private var persistenceDirectory: URL {
+    ///
+    /// Stored as a lazy property so the path is computed once and reused, avoiding
+    /// repeated filesystem lookups on every call.
+    @ObservationIgnored
+    private lazy var persistenceDirectory: URL = {
         let caches = FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask).first
             ?? FileManager.default.temporaryDirectory
         return caches.appendingPathComponent(
             "\(BaseChatConfiguration.shared.bundleIdentifier).downloads",
             isDirectory: true
         )
-    }
+    }()
 
     /// URL of the single JSON file that stores all pending-download metadata.
     private var pendingMetadataFileURL: URL {
@@ -686,21 +690,28 @@ public final class BackgroundDownloadManager: NSObject, @unchecked Sendable {
         return try? JSONDecoder().decode([String: [String: String]].self, from: data)
     }
 
-    /// Writes the pending-downloads dictionary atomically (write to temp file, then rename).
+    /// Writes the pending-downloads dictionary atomically (write to temp file, then atomic swap).
     ///
-    /// Atomic rename means a crash between the write and the rename leaves the old file intact —
-    /// the partial write is never visible to a subsequent read.
+    /// Uses `replaceItemAt(_:withItemAt:backupItemName:options:)` which does the swap in a single
+    /// kernel operation — there is never a moment where the destination file is absent.
     private func writePendingMetadata(_ pending: [String: [String: String]]) throws {
         try ensurePersistenceDirectory()
         let data = try JSONEncoder().encode(pending)
         let tempURL = pendingMetadataFileURL.deletingLastPathComponent()
             .appendingPathComponent("pending-downloads-\(UUID().uuidString).tmp")
         try data.write(to: tempURL, options: .atomic)
-        // Replace the target atomically; removeItem+move is the portable approach on Darwin.
         if FileManager.default.fileExists(atPath: pendingMetadataFileURL.path) {
-            _ = try? FileManager.default.removeItem(at: pendingMetadataFileURL)
+            // replaceItemAt atomically swaps tempURL into the destination, removing tempURL.
+            try FileManager.default.replaceItemAt(
+                pendingMetadataFileURL,
+                withItemAt: tempURL,
+                backupItemName: nil,
+                options: []
+            )
+        } else {
+            // Destination doesn't exist yet; a plain move is sufficient.
+            try FileManager.default.moveItem(at: tempURL, to: pendingMetadataFileURL)
         }
-        try FileManager.default.moveItem(at: tempURL, to: pendingMetadataFileURL)
     }
 
     private func savePendingDownload(
@@ -804,7 +815,8 @@ public final class BackgroundDownloadManager: NSObject, @unchecked Sendable {
 
     /// Deletes resume-data files for download IDs not present in the current pending-metadata
     /// list. These orphans accumulate when a download crashes without the normal teardown path.
-    private func deleteOrphanedResumeDataFiles(knownIDs: Set<String>) {
+    // internal (not private) so unit tests can call it directly with @testable import.
+    internal func deleteOrphanedResumeDataFiles(knownIDs: Set<String>) {
         guard let contents = try? FileManager.default.contentsOfDirectory(
             at: persistenceDirectory,
             includingPropertiesForKeys: nil,
@@ -833,7 +845,11 @@ public final class BackgroundDownloadManager: NSObject, @unchecked Sendable {
     ///
     /// Runs once on first launch after the upgrade. Subsequent launches skip it because
     /// the old keys are no longer present.
-    private func migrateFromUserDefaults() {
+    ///
+    /// UserDefaults keys are only removed after the corresponding file write succeeds, so
+    /// a failed write leaves the data intact for the next launch to retry.
+    // internal (not private) so unit tests can call it directly with @testable import.
+    internal func migrateFromUserDefaults() {
         let defaults = UserDefaults.standard
         let pendingKey = BaseChatConfiguration.shared.pendingDownloadsKey
 
@@ -843,12 +859,17 @@ public final class BackgroundDownloadManager: NSObject, @unchecked Sendable {
            !legacy.isEmpty {
             do {
                 try writePendingMetadata(legacy)
+                // Only clear the key once the file write has confirmed success.
+                defaults.removeObject(forKey: pendingKey)
                 Log.download.info("Migrated \(legacy.count) pending-download(s) from UserDefaults to file")
             } catch {
                 Log.download.error("Failed to migrate pending downloads from UserDefaults: \(error.localizedDescription)")
+                // Leave the UserDefaults key intact so the next launch can retry.
             }
+        } else {
+            // No legacy data to migrate — remove the (now-empty or absent) key.
+            defaults.removeObject(forKey: pendingKey)
         }
-        defaults.removeObject(forKey: pendingKey)
 
         // Migrate any resume-data blobs stored under the legacy "resumeData.<id>" keys.
         let allKeys = defaults.dictionaryRepresentation().keys
@@ -858,12 +879,17 @@ public final class BackgroundDownloadManager: NSObject, @unchecked Sendable {
                 do {
                     try ensurePersistenceDirectory()
                     try data.write(to: resumeDataFileURL(for: id), options: .atomic)
+                    // Only clear the key once the file has been written successfully.
+                    defaults.removeObject(forKey: key)
                     Log.download.info("Migrated resume data for \(id) from UserDefaults to file")
                 } catch {
                     Log.download.error("Failed to migrate resume data for \(id): \(error.localizedDescription)")
+                    // Leave the UserDefaults key intact so the next launch can retry.
                 }
+            } else {
+                // No data blob — remove the dangling key.
+                defaults.removeObject(forKey: key)
             }
-            defaults.removeObject(forKey: key)
         }
     }
 }
--- a/Sources/BaseChatInference/Services/BackgroundDownloadManager.swift
+++ b/Sources/BaseChatInference/Services/BackgroundDownloadManager.swift
@@ -114,10 +114,40 @@ public final class BackgroundDownloadManager: NSObject, @unchecked Sendable {
         return session
     }
 
-    /// Persists download metadata so we can reconnect after app restart.
-    private let defaults = UserDefaults.standard
-    private var pendingDownloadsKey: String {
-        BaseChatConfiguration.shared.pendingDownloadsKey
+    // MARK: - File-based Persistence
+
+    /// Directory in Caches where we store the pending-downloads JSON and per-download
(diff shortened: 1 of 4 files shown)


roryford (commented):
## Code Review — PR #349: Move BackgroundDownloadManager persistence off UserDefaults

Overall the approach is solid and the motivation is compelling. Found **one genuine bug** in the migration path, **two test gaps**, and a few minor nits.

---

### Bug: Migration deletes UserDefaults keys unconditionally on write failure

In `migrateFromUserDefaults`, `defaults.removeObject(forKey: pendingKey)` runs **outside** the `do/catch` block, so if `writePendingMetadata` throws (e.g. Caches temporarily unwritable), the legacy key is wiped and the download list is silently lost. Same issue for resume-data blobs: `defaults.removeObject(forKey: key)` runs even when `data.write` threw.

```swift
// Current (bug)
if ... {
    do {
        try writePendingMetadata(legacy)
    } catch {
        Log.download.error(...)
    }
}
defaults.removeObject(forKey: pendingKey)  // runs even after write failure

// Fix: only delete after confirmed write
if ... {
    do {
        try writePendingMetadata(legacy)
        defaults.removeObject(forKey: pendingKey)  // only on success
        Log.download.info(...)
    } catch {
        Log.download.error(...)
        // leave UserDefaults intact — next launch will retry migration
    }
}
```

This is a real data-loss path. The Caches directory is always available on iOS/macOS, so it would be rare in production, but any path-traversal reject or out-of-space condition could trigger it.

---

### Atomicity: `writePendingMetadata` has a crash window between `removeItem` and `moveItem`

The current sequence:
1. `data.write(to: tempURL, .atomic)` — writes and renames to tempURL ✓
2. `_ = try? FileManager.default.removeItem(at: pendingMetadataFileURL)` — silently swallowed
3. `try FileManager.default.moveItem(at: tempURL, to: pendingMetadataFileURL)` — fails if step 2 silently failed with a permissions error

Between steps 2 and 3 there's a window where the file is absent. A hard kill here leaves the metadata file missing (though the temp .bin resume data is safe). The portable Darwin fix is `replaceItemAt(_:withItemAt:)`, which does the swap atomically in a single syscall:

```swift
// Replaces the file atomically — no removal-then-rename gap
_ = try FileManager.default.replaceItemAt(
    pendingMetadataFileURL,
    withItemAt: tempURL,
    backupItemName: nil,
    options: []
)
```

If the destination doesn't exist, `replaceItemAt` falls back to a plain move. This is the right API for this pattern.

The `try?` on `removeItem` is also a mild diagnostics concern: if the removal fails with a permissions error, `moveItem` then throws a *different* error (`fileWriteFileExists`), making the root cause harder to diagnose from logs.

---

### Test gap: migration path has zero test coverage

`migrateFromUserDefaults` is a one-shot path that users will hit exactly once and can't easily re-trigger. It has no test. Suggest adding one to `BackgroundDownloadIntegrationTests`:

- Pre-seed `UserDefaults.standard` with legacy `pendingDownloadsKey` and `resumeData.<id>` keys
- Call `manager.reconnectBackgroundSession()`
- Assert: JSON file exists and contains the migrated entries, UserDefaults keys are removed
- Cover the failure sub-case: if file write fails, UserDefaults key is **not** removed (once the bug above is fixed)

---

### Test gap: orphan sweep is untested

`deleteOrphanedResumeDataFiles(knownIDs:)` is also uncovered. A simple case to add: create a `resume-orphan.bin` file in the persistence directory without a matching entry in the JSON, call `reconnectBackgroundSession()`, assert the file is gone. Also worth asserting that a file matching a *known* ID is preserved.

---

### Minor: test tearDown doesn't clean the shared Caches persistence dir

`ResumableDownloadTests.tearDown` removes `tempDirectory` (the storage dir) but not the shared Caches persistence directory. Most tests self-clean via `consumeResumeData`, but if a test fails mid-way, resume `.bin` files can bleed into subsequent runs and trigger false positives in the orphan sweep once it gains a test. Not urgent, but worth a `try? FileManager.default.removeItem(at: persistenceDir)` in tearDown.

---

### Nit: `persistenceDirectory` is a computed property recomputed on every call

`persistenceDirectory` constructs a `URL` on every call (including `appendingPathComponent`). It's not expensive but it's called on every read/write path. Making it a lazy `let` would be slightly cleaner — though this is truly minor and I wouldn't block on it.

---

### Checklist

| | |
|---|---|
| Atomicity (temp + rename, same FS) | ✅ temp and dest are in the same directory; rename is atomic |
| Cross-filesystem move risk | ✅ Not present — both URLs derive from `persistenceDirectory` |
| Orphan sweep deletes valid in-progress downloads? | ✅ No — sweep only runs from `restorePendingDownloads`, which loads the current metadata first; active downloads are always in the JSON |
| Directory creation before first use | ✅ `ensurePersistenceDirectory()` called in all write paths |
| Concurrency | ✅ All persistence calls are dispatched through `Task { @MainActor }` in the delegate; no off-actor file I/O introduced by this PR |
| Migration crash safety | ⚠️ Bug above — UserDefaults wiped before file write confirmed |
| Test coverage | ⚠️ Migration + orphan sweep paths untested |

**Verdict**: Please fix the migration data-loss bug before merging. The `replaceItemAt` change is strongly recommended. Tests for migration and orphan sweep can follow in a separate commit or follow-on PR.

roryford (commented):
## Code Review — PR #349: Move BackgroundDownloadManager persistence off UserDefaults

Overall the approach is solid and the motivation is compelling. Found **one genuine bug** in the migration path, **two test gaps**, and a few minor nits.

---

### Bug: Migration deletes UserDefaults keys unconditionally on write failure

In `migrateFromUserDefaults`, `defaults.removeObject(forKey: pendingKey)` runs **outside** the `do/catch` block, so if `writePendingMetadata` throws (e.g. Caches temporarily unwritable), the legacy key is wiped and the download list is silently lost. Same issue for resume-data blobs: `defaults.removeObject(forKey: key)` runs even when `data.write` threw.

The fix is to move `defaults.removeObject` inside the success branch so UserDefaults is only cleared after the file write is confirmed. If the write fails, the key stays intact and the next launch retries migration.

---

### Atomicity: `writePendingMetadata` crash window between `removeItem` and `moveItem`

The current sequence writes a temp file, silently removes the destination, then moves the temp into place. Between the remove and the move there is a window where the metadata file is absent. A hard kill here would leave pending downloads untracked (resume data is safe, but the entry list is gone).

The portable Darwin fix is `FileManager.replaceItemAt(_:withItemAt:backupItemName:options:resultingItemURL:)` which does the swap in a single syscall. If the destination doesn't exist, it falls back to a plain move.

Additionally, the `try?` on `removeItem` masks the real error when it fails with `.permissionDenied`: `moveItem` then throws `.fileWriteFileExists`, making root-cause diagnosis harder from logs.

---

### Test gap: migration path has zero test coverage

`migrateFromUserDefaults` is a one-shot path every user hits exactly once. It has no test. Suggest adding one to `BackgroundDownloadIntegrationTests`: pre-seed the legacy UserDefaults keys, call `reconnectBackgroundSession()`, assert the JSON file exists with the right entries and the legacy keys are gone. Also cover the failure sub-case: write fails → UserDefaults key NOT removed (once the bug above is fixed).

---

### Test gap: orphan sweep is untested

`deleteOrphanedResumeDataFiles` is also uncovered. A simple case: create a `resume-orphan.bin` in the persistence directory with no matching JSON entry, call `reconnectBackgroundSession()`, assert the file is deleted. Also assert a file matching a known ID is preserved.

---

### Minor: `ResumableDownloadTests.tearDown` doesn't clean the shared Caches persistence dir

`tearDown` removes `tempDirectory` but not the shared Caches persistence directory. Most tests self-clean via `consumeResumeData`, but mid-test failures can leave `.bin` files that bleed into subsequent runs. A `try? FileManager.default.removeItem(at: persistenceDir)` in `tearDown` would close this.

---

### Nit: `persistenceDirectory` reconstructed on every call

`persistenceDirectory` is a computed property that constructs a URL on every invocation. Making it a `lazy let` would be cleaner — though this is truly minor.

---

### Checklist

| Check | Result |
|---|---|
| Atomicity (temp + rename, same FS) | Pass — temp and dest share `persistenceDirectory`; rename is atomic |
| Cross-filesystem move risk | Pass — both URLs derived from same directory |
| Orphan sweep deletes valid in-progress downloads | Pass — sweep runs after loading current metadata; active downloads always in JSON |
| Directory creation before first use | Pass — `ensurePersistenceDirectory()` called in all write paths |
| Concurrency — file I/O on wrong actor | Pass — all delegate callbacks dispatch through `Task { @MainActor }` before touching persistence |
| Migration crash safety | **Fail** — UserDefaults wiped before file write confirmed |
| Migration + orphan sweep test coverage | **Fail** — untested |

**Verdict**: Fix the migration data-loss bug before merging. The `replaceItemAt` change is strongly recommended. Tests for migration and orphan sweep can follow in the same PR or a close follow-on.

roryford (commented):
Code review for #349 — Move BackgroundDownloadManager persistence off UserDefaults.

One genuine bug, two test gaps, and a few nits. Details below.

BUG: Migration deletes UserDefaults keys unconditionally on write failure

In migrateFromUserDefaults, defaults.removeObject(forKey: pendingKey) runs outside the do/catch, so if writePendingMetadata throws the legacy key is deleted before the file write is confirmed. The download list is silently lost. Same for resume-data blobs: defaults.removeObject(forKey: key) runs even when data.write threw.

Fix: move defaults.removeObject inside the success branch. If the write fails, leave UserDefaults intact so the next launch retries migration.

ATOMICITY CONCERN: writePendingMetadata has a crash window

Sequence: write-temp -> try? removeItem -> moveItem. Between the remove and the move the metadata file is absent; a hard kill here loses the entry list (resume data is unaffected). The correct Darwin API is FileManager.replaceItemAt(_:withItemAt:backupItemName:options:) which does the swap in one syscall. The try? on removeItem also masks the original error: if removeItem fails with .permissionDenied, moveItem then throws .fileWriteFileExists, making root-cause diagnosis harder from logs.

TEST GAP: Migration path has zero test coverage

migrateFromUserDefaults is a one-shot path every user hits exactly once on upgrade. Suggest: pre-seed legacy UserDefaults keys, call reconnectBackgroundSession(), assert JSON file contains the entries and legacy keys are removed. Also cover the failure sub-case: write fails -> UserDefaults key NOT removed (after the bug above is fixed).

TEST GAP: Orphan sweep is untested

deleteOrphanedResumeDataFiles is uncovered. Simple case: create a resume-orphan.bin with no matching JSON entry, call reconnectBackgroundSession(), assert file is deleted. Also assert a file for a known ID is preserved.

MINOR: ResumableDownloadTests.tearDown doesn't clean the shared Caches persistence dir

tearDown removes tempDirectory but not the shared Caches persistence dir. Mid-test failures can leave .bin files that bleed into subsequent runs. A try? FileManager.default.removeItem(at: persistenceDir) in tearDown closes this.

NIT: persistenceDirectory recomputed on every call (computed property). Could be a lazy let.

CHECKLIST:
- Atomicity (temp+rename, same FS): PASS
- Cross-filesystem move risk: PASS
- Orphan sweep deletes valid in-progress downloads: PASS (sweep runs after loading current metadata)
- Directory creation before first use: PASS (ensurePersistenceDirectory in all write paths)
- File I/O on wrong actor: PASS (all delegate callbacks dispatch through Task { @MainActor })
- Migration crash safety: FAIL (UserDefaults wiped before file write confirmed)
- Migration + orphan sweep test coverage: FAIL (untested)

Verdict: Fix the migration data-loss bug before merging. replaceItemAt is strongly recommended. Tests for migration and orphan sweep can land in the same PR or a close follow-on.

roryford (commented):
Code review for PR 349.

roryford (commented):
Code review for PR 349.

PR #359, opened 2026-04-14
docs(security): correct PinnedSessionDelegate pin documentation + CI guard

roryford:
## Summary

- Rewrites the stale `PinnedSessionDelegate.pinnedHosts` doc comment that told maintainers the pins were "intentionally left empty."
- Adds two CI regression-guard tests (`test_ciGuard_loadDefaultPins_shipsOpenAIPins` / `test_ciGuard_loadDefaultPins_shipsAnthropicPins`) so nobody can remove `loadDefaultPins()` as dead code without turning CI red.

## Why

`PinnedSessionDelegate` ships real certificate pins for `api.openai.com` and `api.anthropic.com` (Google Trust Services WE1 intermediate + GTS Root R4 backup), and the delegate fails-closed when a required production host has an empty pin set. But the doc comment on the pin-set storage still said:

> These pins are intentionally left empty and must be populated with actual SPKI hashes before enabling pinning in production.

A future maintainer reading that comment could delete `loadDefaultPins()` as obvious placeholder code and silently break every OpenAI and Anthropic request shipped by the framework. Nothing in CI would have caught it.

This PR fixes both halves: the doc now describes what actually ships and how to override it, and two guard tests assert that `loadDefaultPins()` populates both required hosts with at least two pins each (primary + backup, so routine cert rotation doesn't lock the framework out). The guard tests include comments explaining why they exist and instructing future maintainers not to weaken them.

Sabotage-verified: temporarily removing `api.openai.com` from the `loadDefaultPins()` loop trips all three assertions in the OpenAI guard test (nil, empty, and < 2 pins).

No behavior change, no new dependencies, no scope creep — doc + one test file.

## Release Note

**Clarified certificate-pinning docs and hardened CI against accidental regressions** — The `PinnedSessionDelegate.pinnedHosts` doc comment previously described the default pin set as empty placeholders, contradicting the `loadDefaultPins()` function that actually populates real GTS WE1 intermediate + GTS Root R4 backup pins for `api.openai.com` and `api.anthropic.com`. The doc now accurately describes the shipped pin set, why intermediate pinning is preferred, and how host apps can override `pinnedHosts` for custom gateways. Two new CI regression-guard tests assert both production hosts ship with a minimum of two pins (primary + backup for rotation safety), so future maintainers can't silently remove `loadDefaultPins()` without turning CI red.

## Test plan

- [x] `swift test --filter BaseChatCoreTests --disable-default-traits` — passes (83 tests)
- [x] `swift test --filter BaseChatInferenceTests --disable-default-traits` — passes (69 tests)
- [x] `swift test --filter BaseChatUITests --disable-default-traits` — passes (431 tests)
- [x] `swift test --filter BaseChatBackendsTests --disable-default-traits` — passes (includes 14 PinnedSessionDelegate tests, 2 new)
- [x] Sabotage verified: skipping `api.openai.com` in `loadDefaultPins()` trips all three assertions in the new OpenAI guard test.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/Sources/BaseChatBackends/PinnedSessionDelegate.swift
+++ b/Sources/BaseChatBackends/PinnedSessionDelegate.swift
@@ -25,12 +25,23 @@ final class PinnedSessionDelegate: NSObject, @preconcurrency URLSessionDelegate 
     ///   openssl dgst -sha256 -binary | base64
     /// ```
     ///
-    /// These pins are intentionally left empty and must be populated with actual
-    /// SPKI hashes before enabling pinning in production.
+    /// Default pins are populated lazily by `loadDefaultPins()` on first access.
+    /// Today that function ships two SPKI hashes — the Google Trust Services WE1
+    /// intermediate CA and its issuing GTS Root R4 — and applies them to both
+    /// `api.openai.com` and `api.anthropic.com` because both providers sit
+    /// behind Google Trust Services. Pinning the intermediate (plus the root as
+    /// a backup) rather than the leaf keeps the pin set stable across routine
+    /// leaf-certificate renewals: pins only need to change when a provider
+    /// rotates its signing infrastructure.
     ///
     /// `api.openai.com` and `api.anthropic.com` are treated as required pinned
     /// production hosts: missing or empty pin sets fail closed.
     /// Unknown custom hosts continue to use default trust when no pins are set.
+    ///
+    /// Host apps that need to override or extend the shipped pin set (e.g.
+    /// pointing a provider at a private gateway with its own CA) can write
+    /// directly to `pinnedHosts` before any network requests are issued; values
+    /// set by the host app are preserved by `loadDefaultPins()`.
     // NSLock guards concurrent reads/writes from multiple URLSession delegate
     // callback threads, which can arrive on arbitrary background threads.
     private static let _pinnedHostsLock = NSLock()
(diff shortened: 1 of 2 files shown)


roryford:
### Review pass — small polish pushed (7cf01c4)

Security review was already done on this PR; my pass was focused on doc/test discoverability.

**Approved as-is**
- Doc rewrite is accurate: it matches what `loadDefaultPins()` actually ships, calls out that the shipped pins target the GTS **intermediate + root** (stable across leaf rotations), and instructs host apps that they can override `pinnedHosts` before networking starts. The risk of leaf-pinning / outage is implicit in "survive individual certificate renewals". Good.
- `ciGuard_` test-name prefix is a reasonable convention for this kind of "don't delete this code" guard. It isn't used elsewhere in the repo yet but the naming makes the intent obvious and the in-line comment is explicit about not weakening the assertions. Happy to leave the prefix.
- Both guard tests check the three symptoms — nil, empty, and `< 2 pins` — so a partial regression (e.g. someone removing only the backup pin) still lights up, not just a full deletion. Sabotage-verified in the PR body. Fine.
- The two-pin minimum is a thoughtful choice — losing the backup during cert rotation is exactly how you get a field outage.

**Fixed in this follow-up**
- Inline comment on line 81 had two U+FFFD replacement characters where an em dash was meant (`"intermediate �� shared by both hosts"`). Harmless, but the adjacent doc block reads cleanly so it stood out — tidied.

**Flagged but not changed (pre-existing, out of scope)**
- The `///` doc block on `pinnedHosts` (lines 18–44) is separated from the `public static var pinnedHosts` declaration by a `private static let _pinnedHostsLock` and a non-doc `//` comment, so SwiftDoc / DocC will attach the pin-set documentation to the lock rather than to the public property. Not a regression introduced by this PR, but worth noting for a future cleanup — moving the `///` block just above `public static var pinnedHosts` would fix it.

**Approved for merge once CI stays green.**

PR #360, opened 2026-04-14
fix(security): block SSRF via custom endpoint private-range URLs

roryford:
## Summary

`APIEndpoint.isValid` previously accepted any URL that parsed and had a non-nil host, and only required HTTPS for non-loopback hosts. This PR tightens it so that private, link-local, reserved, and alternate-loopback IP ranges are rejected even over HTTPS, and non-`http`/`https` schemes are rejected outright.

## Why

`APIEndpoint` is user-configurable from the settings UI. A user tricked into importing a malicious endpoint config — or a compromised sync path that seeds one — turns the device into a proxy into whatever the endpoint's host resolves to. Concretely:

- **LAN pivot** via RFC1918 (10/8, 172.16/12, 192.168/16) — router admin panels, printers, home-assistant dashboards, internal APIs.
- **Cloud metadata exfil** via 169.254.169.254 (AWS/GCP/Azure IMDS). On an EC2-hosted macOS runner or a corp device on a cloud VPN, that endpoint hands out short-lived credentials with whatever role is attached to the instance.
- **Alternate loopback encodings** (`127.0.0.2`, `[::ffff:127.0.0.1]`) bypass the old `isLocalhost` allowlist while still reaching a local service the user didn't intend to expose.
- **Non-HTTP schemes** (`file://`, `ftp://`, `data:`, `javascript:`) were never explicitly forbidden — the old check only gated HTTPS-vs-localhost.

## What's blocked

IPv4:
- `0.0.0.0/8` — "this host" / any-address
- `10.0.0.0/8`, `172.16.0.0/12`, `192.168.0.0/16` — RFC1918
- `127.0.0.0/8` except the exact `127.0.0.1` — alternate loopback encodings
- `169.254.0.0/16` — link-local (AWS/GCP/Azure IMDS)
- `224.0.0.0/4` — multicast
- `240.0.0.0/4` — reserved / future use (includes `255.255.255.255` broadcast)

IPv6:
- `fc00::/7` — unique local addresses
- `fe80::/10` — link-local
- `::ffff:0:0/96` — IPv4-mapped (would otherwise bypass the IPv4 filter via `[::ffff:192.168.1.1]`)

Schemes:
- Anything other than `http`/`https` (includes `file`, `ftp`, `data`, `javascript`, `gopher`, etc.)

## What's NOT blocked

- **Loopback allowlist preserved**: `127.0.0.1`, `::1`, and `localhost` continue to work over plain HTTP so local Ollama / LM Studio dev servers are unaffected.
- **DNS rebinding is explicitly out of scope for this layer.** `APIEndpoint.isValid` is called synchronously from SwiftUI settings forms and must stay fast, so DNS names are not resolved at validation time. A DNS name that later resolves to a private range is still a risk and should be mitigated at the request layer (pinned resolution, or rejecting private IPs returned by URLSession host resolution) if a deployment needs that guarantee. This is documented in the helper's doc comment.

## Implementation notes

- New `static func isDisallowedPrivateHost(_ url: URL) -> Bool` on `APIEndpoint`. It parses IPv4 dotted-quad and IPv6 literal hosts (including `::` zero-run compression, embedded IPv4 suffixes, and zone identifiers) and classifies them against the block list.
- `isLocalhost` tightened to the three explicit literals only; broader `127.x.x.x` and `[::ffff:127.0.0.1]` now fall through to the disallowed-range gate.
- `isValid` signature unchanged (`public var isValid: Bool`) — no UI impact.
- No new dependencies.

## Release Note

**Custom API endpoints can no longer target private networks.** `APIEndpoint.isValid` now rejects URLs whose host is in an RFC1918 range, the 169.254.0.0/16 link-local range (which includes cloud-metadata services like AWS IMDS at 169.254.169.254), IPv6 ULA/link-local ranges, or alternate loopback encodings like `127.0.0.2` and `[::ffff:127.0.0.1]`. It also rejects any scheme other than `http` and `https`, closing the door on `file://`, `data:`, and `javascript:` URLs. The loopback bypass (`127.0.0.1`, `::1`, `localhost`) is preserved so local Ollama and LM Studio setups keep working. DNS rebinding is out of scope for the validator and should be mitigated at the request layer if a deployment needs that guarantee.

## Test plan

- [x] `swift test --filter BaseChatCoreTests --disable-default-traits` — 116 pass, 0 fail
- [x] `swift test --filter BaseChatInferenceTests --disable-default-traits` — 69 pass, 0 fail
- [x] `swift test --filter BaseChatUITests --disable-default-traits` — 431 pass, 0 fail
- [x] `swift test --filter BaseChatBackendsTests --disable-default-traits` — 63 pass, 0 fail
- [x] 33 new tests in `Tests/BaseChatCoreTests/CustomEndpointValidationTests.swift` cover every accepted URL class and every rejection class called out in the threat model.
- [x] Sabotage check: commented out `if Self.isDisallowedPrivateHost(url) { return false }` and confirmed the four HTTPS-private-range tests (`test_rejects_rfc1918_over_https`, `test_rejects_imds_over_https`, `test_rejects_ipv6ULA_over_https`, `test_rejects_ipv4MappedLoopback_over_https`) fail. Restored before commit.
- [x] Existing `APIEndpointTests` unchanged and still passing — no behavior drift for already-covered cases.

### DIFF
--- a/Sources/BaseChatCore/BaseChatSchemaV3.swift
+++ b/Sources/BaseChatCore/BaseChatSchemaV3.swift
@@ -265,13 +265,37 @@ public enum BaseChatSchemaV3: VersionedSchema {
         /// exists in the Keychain. Use `APIProvider.requiresAPIKey` and
         /// `KeychainService.retrieve(account: endpoint.keychainAccount)` separately
         /// to check credential readiness.
+        ///
+        /// Security: rejects URLs that would let a malicious endpoint config pivot
+        /// into the user's LAN or cloud metadata services (SSRF). See
+        /// ``APIEndpoint/isDisallowedPrivateHost(_:)`` for the blocked IP ranges.
+        /// Only `http`/`https` schemes are accepted. Non-HTTPS is permitted only for
+        /// the explicit loopback allowlist (`127.0.0.1`, `::1`, `localhost`).
         public var isValid: Bool {
-            guard let url = URL(string: baseURL), url.scheme != nil, url.host != nil else {
+            guard let url = URL(string: baseURL),
+                  let scheme = url.scheme?.lowercased(),
+                  url.host() != nil else {
                 return false
             }
 
-            // HTTPS required for remote endpoints
-            if !isLocalhost(url) && url.scheme != "https" {
+            // Only http/https are valid. Reject file, ftp, data, javascript, etc.
+            guard scheme == "http" || scheme == "https" else {
+                return false
+            }
+
+            // Loopback dev servers (e.g. Ollama, LM Studio) are allowed over plain HTTP.
+            if Self.isLocalhost(url) {
+                return true
+            }
+
+            // Non-loopback must be HTTPS.
+            if scheme != "https" {
+                return false
+            }
+
+            // Block SSRF into LAN / cloud metadata even when HTTPS is used. A
+            // cert-pinned private-range target is still a private-range target.
+            if Self.isDisallowedPrivateHost(url) {
                 return false
             }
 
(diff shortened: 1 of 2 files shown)


roryford:
Follow-up review commit pushed (cbe2cdd):

**Trailing-dot FQDN bypass.** `https://192.168.1.1.` resolves identically to `https://192.168.1.1` on mainstream resolvers, but the dotted-quad parser rejected the 5-part split so the host fell through as a DNS name and slipped past the private-range gate. Added a one-line strip of a trailing dot before IP classification, plus regression tests for RFC1918 and IMDS variants. Sabotage-verified: with the strip disabled the two new tests fail for the right reason.

**Uppercase-hex IPv6 regression guard.** The classifier already lowercases `url.host()` before matching, but there was no test for `FE80::1` / `::FFFF:127.0.0.1`. Added two to lock that in.

**Deliberately not changed — observation for follow-up.** `APIEndpointRow` currently renders a generic "Incomplete" label for any `!isValid`. After this PR there are seven distinct rejection reasons (private range, loopback alternate encoding, non-HTTPS, unknown scheme, empty, no host, etc.). A future iteration could return `Result<Void, ValidationReason>` so the settings UI can say "Private IPs are not allowed" specifically — but keeping the `Bool` contract as-is in this PR avoids a public-API change mid-security-fix. Happy to open a separate issue if you'd like.

PR #361, opened 2026-04-14
feat(security): bound SSE stream memory and rate to defeat hostile-server DoS

roryford:
## Summary

- Adds `SSEStreamLimits` (per-event bytes, total bytes, events/sec) and `SSEStreamError` in `BaseChatInference`, wired into `SSEStreamParser.parse` and `SSEStreamParser.streamEvents`.
- Exposes configuration globally via `BaseChatConfiguration.sseStreamLimits` and per backend via `SSECloudBackend.sseStreamLimits`, with conservative `.default` values (1 MB / 50 MB / 5,000 events per second) well above any mainstream provider's real throughput.
- Applies the same caps to `OllamaBackend`'s NDJSON path so a hostile Ollama-compatible server gets the same treatment as SSE.

## Why

`SSEStreamParser` had no bounds on per-event payload size, total stream size, or event frequency. Users who point `CustomEndpoint.baseURL` at an untrusted server were exposed to trivial denial-of-service vectors: a single 100 MB `data:` line exhausted memory, an unbounded stream of small valid events persisted forever, and millions of 1-byte events starved the UI. The defaults here cost zero to legitimate use while making those attacks impossible.

## Tuning guide

- Global: `BaseChatConfiguration.shared.sseStreamLimits = SSEStreamLimits(maxEventBytes: 500_000, maxTotalBytes: 10_000_000, maxEventsPerSecond: 2_000)` at app launch to tighten across every cloud backend.
- Per backend: `backend.sseStreamLimits = SSEStreamLimits(...)` to lock down a single untrusted endpoint while leaving OpenAI/Anthropic at the defaults.
- There is deliberately no `nil`/unlimited option. Bounded caps are the point.

## Release Note

**SSE DoS hardening for cloud backends** — BaseChatKit now bounds every Server-Sent Events and NDJSON stream by per-event size, total size, and events-per-second. Defaults are well above any real provider so legitimate OpenAI, Anthropic, and Ollama traffic is unaffected, but a hostile or misconfigured upstream can no longer drive a chat app out of memory or starve its UI. Host apps can tighten the caps globally via `BaseChatConfiguration.sseStreamLimits` or per backend via `SSECloudBackend.sseStreamLimits`, which is especially useful when pointing `CustomEndpoint.baseURL` at an untrusted server.

## Test plan

- [x] `swift test --filter BaseChatInferenceTests --disable-default-traits` (651 tests, 0 failures) — includes 7 new limits tests in `SSEStreamParserTests`.
- [x] `swift test --filter BaseChatCoreTests --disable-default-traits` (83, 0 failures).
- [x] `swift test --filter BaseChatUITests --disable-default-traits` (431, 0 failures).
- [x] `swift test --filter BaseChatBackendsTests --disable-default-traits` (131, 0 failures).
- [x] Happy path: 1,000-event OpenAI-style stream passes under `.default`.
- [x] `eventTooLarge`: 2 MB single `data:` line rejected with correct error and observed size.
- [x] `streamTooLarge`: many small events summing above a 50 KB total cap rejected.
- [x] `eventRateExceeded`: 6,000-event burst against a 5,000/s cap rejected.
- [x] Sabotage: raising the rate cap 100x lets the same burst pass — confirms the test actually exercises the limit.
- [x] `BaseChatConfiguration` override is honoured by `SSEStreamParser.parse` callers that don't pass `limits:` explicitly.

### DIFF
--- a/Sources/BaseChatBackends/OllamaBackend.swift
+++ b/Sources/BaseChatBackends/OllamaBackend.swift
@@ -129,24 +129,55 @@ public final class OllamaBackend: SSECloudBackend, CloudBackendURLModelConfigura
     // timeout/2 partially addresses this by showing .stalled.
 
     /// Parses Ollama's NDJSON response format instead of SSE.
+    ///
+    /// Applies the same ``SSEStreamLimits`` caps as the SSE parser so a
+    /// hostile Ollama-compatible server cannot exhaust memory with oversized
+    /// lines, total volume, or an event flood.
     public override func parseResponseStream(
         bytes: URLSession.AsyncBytes,
         continuation: AsyncThrowingStream<GenerationEvent, Error>.Continuation
     ) async throws {
+        let limits = effectiveSSEStreamLimits
         var lineBuffer = Data()
+        var totalBytes = 0
+        var rateWindowStart = ContinuousClock.now
+        var rateWindowCount = 0
+
+        func noteEventYielded() throws {
+            let now = ContinuousClock.now
+            if now - rateWindowStart >= .seconds(1) {
+                rateWindowStart = now
+                rateWindowCount = 1
+                return
+            }
+            rateWindowCount += 1
+            if rateWindowCount > limits.maxEventsPerSecond {
+                throw SSEStreamError.eventRateExceeded(rateWindowCount)
+            }
+        }
+
         for try await byte in bytes {
             if Task.isCancelled { break }
+
+            totalBytes += 1
+            if totalBytes > limits.maxTotalBytes {
+                throw SSEStreamError.streamTooLarge(totalBytes)
+            }
 
             if byte == UInt8(ascii: "\n") {
                 if !lineBuffer.isEmpty {
                     if let line = String(data: lineBuffer, encoding: .utf8),
                        let token = Self.extractToken(from: line) {
+                        try noteEventYielded()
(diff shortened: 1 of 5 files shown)


roryford:
Follow-up review commit pushed (e996789). Three small product-quality nits, no behaviour change:

1. **Expanded \`SSEStreamLimits\` DocC** with a runnable "Tuning" example showing both the global and per-backend override syntax. The PR body spells this out nicely but a developer discovering the type via auto-complete / Quick Help wouldn't see it. It now shows inline.

2. **Dropped \`SSEStreamError.malformed\`.** It was documented as "reserved for future strict-mode use" and never thrown. A public enum case with no producer is API debt — trivial to add back when strict mode actually ships. No external consumers (grepped the tree).

3. **Corrected the rate-limiter comment.** The code is a fixed window from the first event (resets once >=1s has elapsed), not a sliding window or integer-second bucket as the old comment claimed. Also extended \`maxEventsPerSecond\`'s doc with the same clarification so host apps understand the semantics.

Defaults (1 MB / 50 MB / 5 kHz) look well-sized after sanity-checking Anthropic: tool_use content blocks stream in small chunks (rarely more than tens of KB per event), and image payloads go in the initial request rather than the response stream, so 1 MB/event has plenty of headroom.

All four CI suites green on this branch (651 + 120 + 431 + 131).

PR #362, opened 2026-04-14
fix(security): bound quantization regex to prevent ReDoS on crafted filenames

roryford:
## Summary

The quantization extractor on `DownloadableModel` applies a regex to externally-supplied filenames (HuggingFace manifests, curated list, user search). The current pattern

```swift
#"[_\-\.]((?:Q|IQ|F|BF)\d+(?:_[A-Z0-9]+)*)\."#
```

has an unbounded `(?:_[A-Z0-9]+)*` group. Combined with the trailing literal `.`, crafted filenames like `model_Q4_AAAA_AAAA..._AAAAX.gguf` force the regex engine to explore an exponential number of split points before failing, stalling any UI / service code that touches `DownloadableModel.quantization`.

## Why

Filenames here come across a trust boundary: HuggingFace search results, user-provided manifests, and the curated list are all treated as strings that flow into this property. A bounded regex plus a length cap keeps the work envelope predictable even on adversarial input, without regressing any real-world tag (all existing tests and every in-repo fixture still parse).

## Change

- `Sources/BaseChatInference/Models/DownloadableModel.swift`
  - Bound the repetition: `(?:_[A-Z0-9]+)*` -> `(?:_[A-Z0-9]+){0,5}`. Real GGUF tags top out at two suffix components (`Q4_K_M`, `IQ2_XXS`); five is generous.
  - Clip `fileName` to `.prefix(128)` before applying the regex. Any legitimate HuggingFace filename fits well under 128 characters; longer input is treated as hostile.
- `Tests/BaseChatInferenceTests/DownloadableModelTests.swift`
  - New: `test_quantization_IQ2_XXS_extractedCorrectly` (confirms the {0,5} bound comfortably handles real multi-suffix tags).
  - New: `test_quantization_nilForUnknownTag` (non-quant alphanumerics still rejected).
  - New: `test_quantization_pathologicalInput_completesQuickly_andReturnsNil` — 1000-char hostile input must finish in <50ms with no spurious match.
  - New: `test_quantization_inputLongerThan128Chars_isClipped` — a valid tag placed past the 128-char cap is ignored.

Sabotage-verified: reverting to the unbounded pattern + raw input makes the pathological test fail both on timing (0.252s vs ~0ms, 5x over the 50ms budget) and on correctness (spurious quant match).

## Release Note

**Harden the GGUF quantization parser against ReDoS on crafted filenames** - The regex that reads quant tags off `DownloadableModel.fileName` applied an unbounded repetition group to externally-supplied input. Crafted HuggingFace filenames could push the regex engine into exponential-time backtracking, stalling any code path that inspected the property. The pattern now bounds the suffix repetition to five segments (real tags never exceed two) and clips input to 128 characters, matching every real-world GGUF filename without exposing the engine to adversarial input.

## Test plan

- [x] `swift test --filter BaseChatCoreTests --disable-default-traits` (83 tests, 0 failures)
- [x] `swift test --filter BaseChatInferenceTests --disable-default-traits` (648 tests, 0 failures)
- [x] `swift test --filter BaseChatUITests --disable-default-traits` (431 tests, 0 failures)
- [x] `swift test --filter BaseChatBackendsTests --disable-default-traits` (131 tests, 0 failures)
- [x] Sabotage check confirms the pathological test catches the unbounded regex (correctness and >5x timing regression).

### DIFF
--- a/Sources/BaseChatInference/Models/DownloadableModel.swift
+++ b/Sources/BaseChatInference/Models/DownloadableModel.swift
@@ -37,9 +37,16 @@ public struct DownloadableModel: Identifiable, Sendable, Hashable {
     public var quantization: String? {
         guard modelType == .gguf else { return nil }
         // Match common GGUF quant patterns: Q4_K_M, Q8_0, IQ2_XS, F16, etc.
-        let pattern = #"[_\-\.]((?:Q|IQ|F|BF)\d+(?:_[A-Z0-9]+)*)\."#
-        guard let range = fileName.range(of: pattern, options: .regularExpression) else { return nil }
-        let match = fileName[range].dropFirst().dropLast() // strip leading separator and trailing dot
+        // Bound the trailing `_SEGMENT` repetition to {0,5}: real quant tags top out at
+        // two suffix components (e.g. Q4_K_M, IQ2_XXS); five is generous and prevents
+        // catastrophic backtracking on crafted filenames like `_Q4_AAA_AAA..._AAAX.gguf`.
+        let pattern = #"[_\-\.]((?:Q|IQ|F|BF)\d+(?:_[A-Z0-9]+){0,5})\."#
+        // Cap input length before regex evaluation. Any legitimate HuggingFace filename
+        // fits well under 128 characters; longer input is assumed hostile and clipped
+        // to keep the regex engine in a bounded work envelope.
+        let boundedName = String(fileName.prefix(128))
+        guard let range = boundedName.range(of: pattern, options: .regularExpression) else { return nil }
+        let match = boundedName[range].dropFirst().dropLast() // strip leading separator and trailing dot
         return String(match)
     }
 
--- a/Tests/BaseChatInferenceTests/DownloadableModelTests.swift
+++ b/Tests/BaseChatInferenceTests/DownloadableModelTests.swift
@@ -257,13 +257,23 @@ final class DownloadableModelTests: XCTestCase {
             sizeBytes: 1
         )
 
-        let start = Date()
+        // Monotonic clock instead of wall-clock `Date()` — the latter is subject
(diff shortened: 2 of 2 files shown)


roryford:
## Product-quality review pass

Pushed one follow-up (c86658b).

### Changed

- **`test_quantization_pathologicalInput_completesQuickly_andReturnsNil`** — budget loosened from 50ms to 200ms and the stopwatch switched from wall-clock `Date()` to monotonic `DispatchTime.now().uptimeNanoseconds`. The fix itself returns in microseconds so this isn't a tightness concern; the sabotage path takes seconds, so the discriminator still works. 50ms was cutting it too fine for CI variance (10x-billed macOS runners), and a single flake there wastes ~25 billed minutes. `Date()` is also subject to wall-clock adjustments (NTP, manual changes) that could produce a negative elapsed reading.

### Approved as-is

- **`{0,5}` suffix bound** — I grepped the source tree for quant tags; real-world tops out at two suffix components (`Q4_K_M`, `IQ2_XXS`). Five is generous. No change.
- **128-char input cap** — longest filename literal in the source tree (`Mistral-7B-Instruct-v0.3-Q4_K_M.gguf`) is 36 chars. 128 gives ~3x headroom for longer HuggingFace slugs without opening the work envelope. No change.
- **Test naming** — all four new tests follow the `test_quantization_<scenario>_<expectation>` convention consistent with the existing file. No inconsistency to flag.
- **Sabotage verification** in the PR description — independently confirmed by the test design (timing + correctness split into two assertions, the unbounded pattern fails both).

### CI

All four suites pass locally:
- `BaseChatCoreTests` — 83 tests, 0 failures
- `BaseChatInferenceTests` — 650 tests, 0 failures
- `BaseChatUITests` — 431 tests, 0 failures
- `BaseChatBackendsTests` — 63 tests, 0 failures

PR #363, opened 2026-04-14
feat(security)!: make KeychainService.store and .delete throw on failure

roryford:
## Summary

Convert `KeychainService.store` and `KeychainService.delete` from `Bool`-returning to `throws`-based APIs backed by a typed `KeychainError`. `retrieve` keeps returning `String?` because \"not found\" is a normal state. Every callsite is audited and updated — UI paths now surface Keychain failures to the user instead of silently swallowing them.

## Why

`KeychainService.store()` returned `Bool` and callers routinely ignored it. When the device is locked, entitlements are wrong, or an item is corrupted, the user thinks their API key was saved but later hits mysterious auth failures with no signal explaining what happened. Keychain failures are rare but when they occur the user needs a visible, actionable error — a silent false return is actively hostile.

### API surface (breaking)

**Before**
\`\`\`swift
@discardableResult public static func store(key: String, account: String) -> Bool
@discardableResult public static func delete(account: String) -> Bool
\`\`\`

**After**
\`\`\`swift
public enum KeychainError: Error, Equatable {
    case storeFailed(OSStatus)
    case deleteFailed(OSStatus)
}

public static func store(key: String, account: String) throws
public static func delete(account: String) throws   // errSecItemNotFound is still success
public static func retrieve(account: String) -> String?   // unchanged
\`\`\`

`APIEndpoint.setAPIKey(_:)` and `APIEndpoint.deleteAPIKey()` in `BaseChatCore` propagate the throws.

### Callsite audit (7 production + 8 test files)

- `APIEndpointEditorView.save()` — catches on save, surfaces via existing `validationError` banner; rolls back the inserted endpoint when the key write fails on create.
- `RemoteServerConfigSheet.save()` — same pattern, surfaces via the existing `errorMessage` banner.
- `APIConfigurationView.deleteEndpoint()` — logs a warning and proceeds with SwiftData deletion. Leaving the row in the UI with a dangling Keychain entry is worse than a potentially orphaned entry; the failure is logged for diagnostics.
- Test helpers use `try?` in `tearDown`/`defer` cleanup paths, `try` where the call is load-bearing, and `XCTAssertNoThrow` where non-throwing is itself the assertion.

Logging: failures log via `Log.inference.error` with the `OSStatus` as `privacy: .public` and the account as `privacy: .private`, per the in-repo logging conventions.

## Migration guide

Apps consuming `BaseChatKit` that previously wrote to the Keychain directly:

\`\`\`swift
// Before
if !KeychainService.store(key: key, account: id) {
    // handle failure (most callers didn't)
}

// After
do {
    try KeychainService.store(key: key, account: id)
} catch {
    // surface to the user; do not swallow
    showError(\"Could not save API key: \\\\(error.localizedDescription)\")
}
\`\`\`

If you were calling `APIEndpoint.setAPIKey(_:)` or `APIEndpoint.deleteAPIKey()`, add `try`:

\`\`\`swift
try endpoint.setAPIKey(\"sk-…\")
try endpoint.deleteAPIKey()
\`\`\`

Deleting a non-existent item is still non-throwing — cleanup code can keep its `try?` idiom.

## Release Note

**Keychain failures now surface instead of vanishing** — `KeychainService.store` and `.delete` previously returned `Bool` and virtually no caller checked the result. When the Keychain rejected a write (locked device, entitlement mismatch, corrupted item), users saw mysterious auth failures later with no actionable signal. Both APIs now throw a typed `KeychainError`, so editor sheets can present the failure and developers can no longer silently discard it. `KeychainService.retrieve` is unchanged — a missing item is a normal state, not an error. This is a breaking change for any code calling these APIs directly or via `APIEndpoint.setAPIKey` / `APIEndpoint.deleteAPIKey`; add `try` and a user-visible error path.

## Test plan

- [x] `KeychainServiceTests`: updated to `try`-based calls, added `test_roundTrip_storeRetrieveDelete_endToEnd` covering the full pipeline and `test_keychainError_equatable` for the error type. (10 tests, all pass.)
- [x] `KeychainIntegrationTests`: all round-trip tests migrated to throwing API against the real system Keychain. (9 tests, all pass.)
- [x] Sabotage check performed: replacing the stored data with `\"SABOTAGED\"` fails `test_roundTrip_storeRetrieveDelete_endToEnd`, `test_store_andRetrieve`, `test_store_updatesExisting`, and `test_multipleAccounts_isolated` as expected. Sabotage reverted before committing.
- [x] Full CI suite local pass:
  - `BaseChatCoreTests` — 83 tests + 20 + 9 + 2 + 4 = all green
  - `BaseChatInferenceTests` — 646 tests, all green
  - `BaseChatUITests` — 431 tests, all green
  - `BaseChatBackendsTests` — 131 tests + 63 Swift Testing, all green
- [x] Throwing-path unit coverage for injected-fault cases depends on on-device integration testing (noted in the `KeychainServiceTests` doc comment) — no fake `SecItem` layer is introduced.

Closes: none (proactive hardening)

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,7 +285,7 @@ let endpoint = APIEndpoint(
     baseURL: "https://api.openai.com",
     modelName: "gpt-4o-mini"
 )
-endpoint.setAPIKey("sk-...")  // Stored in Keychain
+try endpoint.setAPIKey("sk-...")  // Stored in Keychain; throws on failure
 ```
 
 ## Prompt Templates
--- a/README.md
+++ b/README.md
@@ -285,8 +285,41 @@ let endpoint = APIEndpoint(
     baseURL: "https://api.openai.com",
     modelName: "gpt-4o-mini"
 )
-try endpoint.setAPIKey("sk-...")  // Stored in Keychain; throws on failure
+do {
+    try endpoint.setAPIKey("sk-...")  // Stored in Keychain
+} catch {
+    // Surface `error.localizedDescription` in your settings UI — it already
+    // maps known Keychain statuses (locked device, missing entitlement, etc.)
+    // to a short user-facing sentence and appends the raw OSStatus for logs.
+}
 ```
+
+### Migrating from `Bool`-returning Keychain APIs
+
+`KeychainService.store` / `.delete` and `APIEndpoint.setAPIKey` / `.deleteAPIKey`
+used to return `Bool` and virtually no caller checked the result. They now
+throw a typed `KeychainError` so failures can't be silently discarded.
+
+```swift
+// Before
+if !KeychainService.store(key: key, account: id) { /* usually ignored */ }
+endpoint.setAPIKey("sk-...")
+endpoint.deleteAPIKey()
+
+// After
+try KeychainService.store(key: key, account: id)
+try endpoint.setAPIKey("sk-...")
+try endpoint.deleteAPIKey()
+
+// Cleanup paths where "already gone" is fine stay terse
+try? endpoint.deleteAPIKey()
+```
+
+Deleting a non-existent item is still non-throwing (`errSecItemNotFound` is
+treated as success), so `tearDown` / `deinit` cleanup can keep its `try?`
+idiom. For programmatic recovery, `KeychainError.osStatus` exposes the raw
+code and `localizedDescription` is already user-friendly — there is no need
+to pattern-match the enum for UI purposes.
 
 ## Prompt Templates
 
--- a/Sources/BaseChatCore/BaseChatSchemaV3.swift
+++ b/Sources/BaseChatCore/BaseChatSchemaV3.swift
@@ -249,14 +249,21 @@ public enum BaseChatSchemaV3: VersionedSchema {
         }
 
         /// Stores an API key in the Keychain for this endpoint.
-        @discardableResult
-        public func setAPIKey(_ key: String) -> Bool {
-            KeychainService.store(key: key, account: keychainAccount)
+        ///
+        /// Throws `KeychainError.storeFailed` when the write is rejected
+        /// (locked device, entitlement mismatch, corrupted item). Surface the
+        /// error to the user — silently swallowing it leaves the impression
+        /// that the key was saved when it wasn't.
+        public func setAPIKey(_ key: String) throws {
+            try KeychainService.store(key: key, account: keychainAccount)
         }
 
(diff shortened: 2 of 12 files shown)


roryford:
### Review pass — DX/UX follow-up pushed (0ec2694)

Security review was already done; this is a developer-experience polish on top.

**Fixes included**
- `KeychainError` now conforms to `LocalizedError` with real user-facing text. Before this commit, the editor banners would have rendered the Foundation default "The operation couldn't be completed. (BaseChatInference.KeychainError error 0.)" — useless for the user. Known codes map to short sentences ("The device appears to be locked. Unlock and try again", "The app is missing the Keychain entitlement", etc.), unknown codes fall through to `SecCopyErrorMessageString`, and the raw `OSStatus` is always appended for diagnostics.
- Exposed `KeychainError.osStatus` so apps that need to branch on a specific code don't have to pattern-match the enum.
- Replaced `test_keychainError_equatable` (which was only exercising compiler-synthesised `Equatable` — a no-op to test) with three tests that cover the new behaviour, including a sabotage-verified check that the "locked" keyword appears in the `errSecInteractionNotAllowed` message.
- Dropped the redundant "Could not save API key to the Keychain:" prefix in `APIEndpointEditorView` / `RemoteServerConfigSheet`. `localizedDescription` is already a complete sentence, so the banners don't double-word.
- Added an explicit migration-guide block to README covering both `setAPIKey` and `deleteAPIKey`, including the `try?` cleanup idiom.

**Judgment call flagged, not changed: `APIConfigurationView.deleteEndpoint`**

The current behavior — log a warning and proceed with SwiftData deletion — is the right default. Leaving a visible row with a dangling key is strictly worse than a potentially-orphaned Keychain item: the user would believe the endpoint is still live and keep trying to send to it. Keychain failures on delete are also extraordinarily rare (basically only locked-device or entitlement issues), and the row now contains no secrets once deleted — just an opaque UUID-derived account identifier.

That said, for a future pass it would be reasonable to:
1. Surface the warning to the user as a non-blocking toast ("The stored key couldn't be removed from the Keychain; it will be cleaned up on next unlock.") rather than only logging.
2. Add a background reaper that sweeps the framework's Keychain service namespace on launch and deletes items whose account UUIDs don't match any known `APIEndpoint`. Small, cheap, and addresses the privacy concern directly.

Neither is blocking for this PR. Happy to open a follow-up issue if you want.

**Approved for merge once CI stays green.**

PR #364, opened 2026-04-14
fix(security): sanitize upstream error bodies before surfacing in UI

roryford:
## Summary

- Add `CloudErrorSanitizer` that rejects HTML-shaped error bodies, strips control/zero-width/bidi characters, redacts JWT/URL-shaped messages, and truncates to 256 chars
- Wire it into the three backends that extract error bodies from upstream HTTP responses (`SSECloudBackend`, `ClaudeBackend`, `OllamaBackend`) at the single point where `CloudBackendError.serverError` is constructed
- Keep the raw body available for developer diagnosis via `Log.network.debug` with `privacy: .private`, never surfacing it to the UI
- 19 new tests covering every sanitization branch plus an end-to-end path through `SSECloudBackend`; each assertion verified via sabotage

## Why

Error bodies from cloud providers (OpenAI, Anthropic, Ollama) and user-configured custom endpoints flowed straight from `extractErrorMessage` into `CloudBackendError.serverError(message:)` and out through `errorDescription` to any UI surface rendering the error. That created three risks:

1. **Content injection** — any future UI that renders errors as attributed text, markdown, or HTML (a toast, a detail view wired through an attributed-string cache, etc.) would silently regress and let an attacker who controls the upstream response — or a malicious custom endpoint — inject clickable/styled content into the chat UI. Even a transparent proxy returning a crafted 502 page is enough.
2. **Information leakage** — upstream bodies sometimes echo back internal service names, backend URLs, JWTs, stack traces, or callback URLs. Surfacing these to end users is the wrong default.
3. **Denial-of-UX** — multi-kilobyte error strings wedge error banners.

Sanitization happens at the point of error construction rather than inside `errorDescription` so the raw value is gone from memory rather than just hidden at render time.

## Release Note

**Harden upstream error handling against content injection** — Error messages from cloud backends (OpenAI, Anthropic, Ollama, custom endpoints) are now sanitized before they reach any UI surface. HTML-shaped bodies from transparent proxies are replaced with a generic "Server error from <host>", bodies containing JWT or URL fragments are redacted, control and bidirectional-override characters are stripped, and messages are capped at 256 characters. Raw bodies remain available to developers via `os.Logger` at `.private` privacy for diagnosis. This prevents a compromised or confused upstream from injecting styled/clickable content into the chat UI, leaking internal tokens or callback URLs, or wedging the error banner with a multi-kilobyte payload.

## Test plan

- [x] `swift test --filter CloudErrorSanitizer --disable-default-traits` — 19 tests pass
- [x] `swift test --filter BaseChatBackendsTests --disable-default-traits` — all 82 tests pass
- [x] `swift test --filter BaseChatCoreTests --disable-default-traits` — 83 tests pass
- [x] `swift test --filter BaseChatInferenceTests --disable-default-traits` — 69 tests pass
- [x] `swift test --filter BaseChatUITests --disable-default-traits` — 431 tests pass
- [x] Sabotage-verified each sanitizer guard (HTML rejection, length cap, control-char stripping) by temporarily disabling it and confirming the corresponding test fails

### DIFF
--- a/Sources/BaseChatBackends/ClaudeBackend.swift
+++ b/Sources/BaseChatBackends/ClaudeBackend.swift
@@ -135,9 +135,15 @@ public final class ClaudeBackend: SSECloudBackend, TokenUsageProvider, CloudBack
             throw CloudBackendError.rateLimited(retryAfter: retryAfter)
         default:
             let errorBody = await Self.readErrorBody(from: bytes)
+            Log.network.debug("Claude upstream error body: \(errorBody, privacy: .private)")
+            let host = self.baseURL?.host()
+            let sanitized = CloudErrorSanitizer.sanitize(
+                extractErrorMessage(from: errorBody),
+                host: host
+            )
             throw CloudBackendError.serverError(
                 statusCode: statusCode,
-                message: extractErrorMessage(from: errorBody) ?? "Unknown error"
+                message: sanitized
             )
         }
     }
--- a/Sources/BaseChatBackends/CloudErrorSanitizer.swift
+++ b/Sources/BaseChatBackends/CloudErrorSanitizer.swift
@@ -30,11 +30,18 @@ import Foundation
 ///
 /// The function is pure, idempotent, and safe to call on already-sanitised
 /// input.
-public enum CloudErrorSanitizer {
+///
+/// Access level is intentionally `package` rather than `public`: host apps
+/// should never need to invoke the sanitiser directly because
+/// ``SSECloudBackend/checkStatusCode(_:bytes:)`` already applies it before
+/// constructing every `serverError`. Sibling backend subclasses inside this
+/// SPM package can still reach it when they need to match that behaviour for
+/// their own status-code paths.
+enum CloudErrorSanitizer {
 
     /// Maximum length of the sanitised message in characters, including the
     /// trailing ellipsis when truncation applies.
-    public static let maxLength = 256
+    static let maxLength = 256
 
     /// Sanitises a raw upstream error string for UI surfacing.
     ///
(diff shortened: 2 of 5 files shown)


roryford:
Follow-up review commit pushed (860cac6). One judgment call, happy to revert if you disagree:

**Dropped \`CloudErrorSanitizer\` from \`public\` to \`internal\`.** The sanitiser is only called from three sites inside \`BaseChatBackends\` (\`SSECloudBackend.checkStatusCode\`, \`ClaudeBackend.checkStatusCode\`, \`OllamaBackend.checkStatusCode\`), and host apps should never call it directly — \`SSECloudBackend\` already sanitises every \`serverError\` it constructs, so subclasses inherit that behaviour for free. Exposing a raw-string-processing helper at \`public\` makes it part of our compatibility contract without giving anyone a meaningful extension point. The \`@testable import BaseChatBackends\` in the tests still works, and the DocC index stays focused on types host apps actually consume.

If a future non-SSE cloud backend (e.g. a gRPC transport) ever needs identical treatment, the cleaner paths are (a) route its error construction through \`SSECloudBackend\` so sanitisation comes along for free, or (b) promote to \`package\` at that moment when we have a concrete cross-module consumer. Easier to open up later than to remove from public later.

**Other observations I did not change, flagging for awareness:**

- **Debug log volume.** Every failed request now writes the raw error body at \`Log.network.debug\` with \`privacy: .private\`. In a retry loop against a flaky 5xx upstream this can be chatty. Since \`debug\` level is off by default and the privacy tag keeps the body out of export archives unless the debugger is attached, the noise is opt-in — I'd leave as-is unless a user complains.

- **\`containsHTMLTag\` empty-array safety.** The loop uses \`0..<(scalars.count - 1)\`, which would crash on an empty input (\`0..<-1\`). Currently safe because \`sanitize\` early-returns when \`cleaned.isEmpty\`, but worth a \`guard scalars.count >= 2 else { return false }\` if anyone ever refactors the order. Left alone — the guard is only one-hop away and the crash path is unreachable through the public entry point.

All four CI suites green on this branch (83 + 644 + 431 + 131).

PR #365, opened 2026-04-14
fix(security): harden download paths with temp sweep and filename validator

roryford:
## Summary

Two related defense-in-depth fixes found in a security review of the model download pipeline.

**1. Temp files leak on crash.** `BackgroundDownloadManager` moves the `URLSession`-delivered payload into `FileManager.default.temporaryDirectory` under a `UUID + .download` file and then into the models directory. Happy-path cleanup is best-effort (`try?`). If the app crashes or is force-killed between those two moves, the temp file remains forever. Over time these multi-gigabyte model files accumulate and the user has no way to reclaim the space short of rebooting.

**2. Path-traversal hardening.** `DownloadableModel.fileName` comes from HuggingFace manifests or user-supplied curated entries. Writes are currently guarded by URL-standardized `hasPrefix` checks which correctly block `../../etc/passwd.gguf`-style attacks today — but that's load-bearing string logic with no explicit boundary validation. An explicit early-reject validator is layered defense, not a replacement.

## Why

Both issues are defense-in-depth: neither can be reached by a known exploit against the shipping code, but the current mitigation relies entirely on happy-path cleanup (issue A) and on the correctness of a single standardization check (issue B). Belt-and-suspenders reduces blast radius if either regresses in the future and — for A — reclaims disk space from users who have already been bitten by crashes that weren't our fault.

Note on threat model: in the current code, `fileName` is populated either from the curated `CuratedModel` list (maintained in-tree) or from HuggingFace search results (which are themselves generally trusted). The validator is not guarding against a current exploit — it's a hardening step so future integrations that expose the field to less-trusted input don't inherit a traversal hazard.

## What changed

**Temp cleanup (`BackgroundDownloadManager`)**
- New static method `cleanupStaleTempFiles(now:)` scans `temporaryDirectory` for files with the manager's own `basechatkit-dl-` prefix and `.download` extension, and removes those older than 24 hours.
- `reconnectBackgroundSession()` now dispatches the sweep on a `Task.detached(priority: .utility)` so the scan doesn't block the main actor at app launch.
- Temp files are now written with the explicit `basechatkit-dl-<UUID>.download` naming scheme so the sweep has a safe fingerprint to match — unrelated temp files from other subsystems cannot be affected.
- The existing `try?` best-effort cleanup in the happy path is left in place; the launch sweep is the real fix, not a replacement.

**Filename validator (`DownloadableModel`)**
- New `DownloadableModel.validate(fileName:)` and `FileNameError` enum.
- Wired into `BackgroundDownloadManager.startDownload(_:plan:)` (earliest disk-touching boundary) plus `retryDownload(id:)` and `restorePendingDownloads()` (so corrupted persistence cannot bypass the check on resume).
- Per-component inspection so the single slash in legitimate MLX filenames like `mlx-community/Phi-4-mini-instruct-4bit` remains legal. Rejects: empty, `..`, `.`, backslash, `//`, leading dot on any component, control characters (incl. null byte and DEL), and names `>= 255` characters.

## Release Note

**Download paths hardened against temp-file leaks and traversal payloads** — the background download manager now sweeps orphaned temp files (files matching its own `basechatkit-dl-*.download` signature, older than 24 hours) on every app launch, reclaiming disk that would otherwise accumulate when a process crashes mid-download. A new `DownloadableModel.validate(fileName:)` validator runs at the earliest disk-touching boundary to reject empty, traversal (`..`), backslash-separated, hidden, control-character, and over-long filenames before they reach the filesystem. The existing URL-standardization prefix guards remain in place — this is layered defense, not a replacement.

## Test plan

- [x] 17 new unit tests for `DownloadableModel.validate` cover every rejection rule plus the legitimate curated MLX `namespace/name` shape
- [x] 6 new integration tests for `cleanupStaleTempFiles` cover: stale file removed, fresh file preserved, unrelated files untouched, injected-clock boundary, idempotence, and prefix/extension contract
- [x] 3 new tests on `BackgroundDownloadManager.startDownload` wire-check that the validator rejects `../../etc/passwd` and backslash inputs and still accepts `mlx-community/*` curated names
- [x] Sabotage-verified — each code path was independently broken to confirm the tests fail, then restored (confirmed for both the prefix filter in the sweep and the `try DownloadableModel.validate(...)` call at `startDownload`)
- [x] Full CI suite passes locally: `BaseChatCoreTests` (83), `BaseChatInferenceTests` (671, up from 647), `BaseChatUITests` (431), `BaseChatBackendsTests` (131) — 1316 tests, 0 failures

### DIFF
--- a/Sources/BaseChatInference/Models/DownloadableModel.swift
+++ b/Sources/BaseChatInference/Models/DownloadableModel.swift
@@ -91,6 +91,93 @@ public struct DownloadableModel: Identifiable, Sendable, Hashable {
     }
 }
 
+// MARK: - File Name Validation
+
+/// Errors raised by ``DownloadableModel/validate(fileName:)``.
+///
+/// These represent categories of unsafe filename inputs that could lead to a
+/// path-traversal write outside the models directory, hidden-file shenanigans,
+/// or name-collision attacks against shell tooling.
+public enum FileNameError: LocalizedError, Equatable {
+    case empty
+    case pathTraversal
+    case pathSeparator
+    case hidden
+    case tooLong
+    case controlCharacter
+
+    public var errorDescription: String? {
+        switch self {
+        case .empty:
+            return "Model filename is empty."
+        case .pathTraversal:
+            return "Model filename contains a path-traversal component (\"..\")."
+        case .pathSeparator:
+            return "Model filename contains an invalid path separator."
+        case .hidden:
+            return "Model filename starts with a dot, which is not permitted."
+        case .tooLong:
+            return "Model filename exceeds the 255-character limit."
+        case .controlCharacter:
+            return "Model filename contains control characters."
+        }
+    }
+}
+
+extension DownloadableModel {
+
+    /// Maximum permitted length for the entire filename string and each component.
+    ///
+    /// 255 is the POSIX `NAME_MAX` on HFS+/APFS; longer names cannot be written
+    /// to disk regardless of other checks.
+    static let maxFileNameLength = 255
+
+    /// Validates that a filename is safe to append to a base directory URL.
+    ///
+    /// This is a belt-and-suspenders check layered on top of the URL-standardized
+    /// `hasPrefix` guards used by `BackgroundDownloadManager` when moving files
+    /// into the models directory. The validator runs at the earliest point a
+    /// filename is accepted from external input (manifests, Hub search results)
+    /// so malformed input is rejected before it reaches any filesystem operation.
+    ///
+    /// MLX snapshot models legitimately use filenames of the form
+    /// `"mlx-community/Phi-4-mini-instruct-4bit"` — a single forward-slash between
+    /// namespace and repo name. The validator therefore inspects each path
+    /// component in isolation rather than rejecting slashes outright.
+    ///
+    /// - Parameter fileName: The untrusted filename string.
+    /// - Throws: ``FileNameError`` describing the first rule the input violates.
+    public static func validate(fileName: String) throws {
+        guard !fileName.isEmpty else { throw FileNameError.empty }
+        guard fileName.count < maxFileNameLength else { throw FileNameError.tooLong }
+        // Backslashes are never legitimate on Apple platforms and are always a
+        // sign of Windows-style traversal or filesystem confusion attacks.
+        guard !fileName.contains("\\") else { throw FileNameError.pathSeparator }
+        // Reject null bytes and other C0/C1 control characters plus DEL. These
+        // can truncate strings at the C boundary or confuse shell tooling.
+        guard fileName.unicodeScalars.allSatisfy({ $0.value >= 0x20 && $0.value != 0x7F }) else {
+            throw FileNameError.controlCharacter
+        }
+
+        // Inspect each forward-slash-separated segment. Legitimate MLX filenames
+        // contain one slash (namespace/name); traversal payloads contain "..".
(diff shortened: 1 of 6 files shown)


roryford:
## Product-quality review pass

Pushed two follow-up commits (c7e6002, 97638b4). Summary of what changed and what I left alone.

### Changed

- **`FileNameError` is now `internal`.** Host apps catch plain `Error` from `validate(fileName:)` and surface `localizedDescription`. Pinning the concrete case list as public API would freeze the category set; a future rule addition would be a SemVer break.
- **Split `.pathSeparator` into `.backslash`, `.emptyComponent`, `.tooManyComponents`.** The single-case rollup gave developers no actionable info about which sub-rule tripped. Each now has a distinct `localizedDescription`; a new test locks that they stay distinct.
- **Reject `> 1` forward-slash components.** `a/b/c/file.gguf` is not a HuggingFace filename pattern BaseChatKit honours — silently accepting a sub-path write was an over-broad reading of the MLX namespace exception. Traversal payloads (`..`, `.`) are classified **before** the count check so deep inputs like `../../etc/passwd` still surface `.pathTraversal` (more actionable than `.tooManyComponents`).
- **Tightened control-character guard** to reject the C1 range (0x80–0x9F). Most importantly U+0085 NEXT LINE renders invisibly and would confuse log output and shell tooling.
- **Documented the temp-sweep decision rules** in the `cleanupStaleTempFiles` header (what it deletes, what it preserves, the 24h-threshold known limitation for users who suspend a download and reopen 25 hours later).
- **Internalized the `now:` seam.** Public no-arg overload for hosts; `internal static func cleanupStaleTempFiles(now:) -> (removed, bytesReclaimed)` for tests. `@discardableResult` so hosts that want a "freed N MB" banner after crash recovery can read the counts.
- **Unconditional `.info` log of the sweep outcome.** The previous "only log when `removed > 0`" made a silent zero-count result indistinguishable from a sweep that never ran — that's the debuggability trail a "my download vanished" report relies on.
- **Four Unicode fuzz tests** cover emoji + combining marks accepted, C0/C1/DEL mixed-script rejected, tab inside a namespaced path rejected.

### Flagged but left

- **resumeData-aware sweep.** The prompt asked whether the 24h rule is sufficient vs. cross-checking live task paths. `URLSession` does not expose in-flight task temp paths, so a proper check would require serializing the active-task path at write time and excluding it from the sweep. That's a non-trivial refactor for a narrow edge case (suspend-and-resume after >24h). Documented as a known limitation in the header; not worth chasing in this PR. The user still has the URLSession-side resume metadata and can retry — only the on-disk partial blob is reclaimed.
- **Retry/restore validator test coverage.** The prompt asked whether `retryDownload(id:)` and `restorePendingDownloads()` have tests that exercise the validator. The source paths call `try DownloadableModel.validate(fileName:)` on the persisted `fileName` before rehydrating state, but reaching those paths from a test requires pushing JSON into the pending-metadata persistence (both helpers are `private`). Net new coverage would require test-only accessors or fixture JSON; the existing direct-validator tests already cover every rejection rule, so the incremental safety is small. Noted as a gap but not addressed.

### Security-impact of changes

Strictly tighter than before: the C1 control range and the `> 1 /` rule add new rejections, and no previously-rejected input now validates.

### CI

All four suites pass locally:
- `BaseChatCoreTests` — green
- `BaseChatInferenceTests` — 679 tests, 0 failures
- `BaseChatUITests` — green
- `BaseChatBackendsTests` — green

PR #368, opened 2026-04-14
feat: surface typed APIEndpointValidationReason in UI settings

roryford:
## Summary

`APIEndpoint.isValid: Bool` swallowed every rejection reason into a single bit — the settings UI could only say "Incomplete", which left users guessing whether they had a typo, a missing scheme, or an insecure HTTP remote. This PR replaces that with a typed `validate() -> Result<Void, APIEndpointValidationReason>` and surfaces the specific reason under the endpoint row.

- New `APIEndpointValidationReason` enum in `BaseChatCore` with `.emptyURL`, `.malformedURL`, `.insecureScheme`. Conforms to `Error`, `Equatable`, `Sendable`, and `LocalizedError` so the UI can render `errorDescription` directly.
- `APIEndpoint.isValid` is preserved as a derived `Bool` over `validate()`, so existing callers (tests, view models) keep compiling without churn.
- `APIEndpointRow` switches on `validate()` and displays the failure message as a caption beneath the endpoint name. VoiceOver reads the specific reason instead of "Incomplete".
- 18 new unit tests in `Tests/BaseChatCoreTests/APIEndpointValidationTests.swift` pin the input-to-reason mapping, plus a `test_validate_isValid_derivedFromValidate_*` pair that anchors the relationship between the two APIs.

## Why

The reported ambiguity was specifically "users don't know what they broke". The current rejection classes all come from `validate()`'s URL parsing / localhost / scheme checks; each now has a distinct case so the row can give an actionable hint. The enum is also open for future cases (private IPs, link-local, IPv6 ULA) without further API breakage — additions will be additive.

Scope was kept deliberately narrow: only rejection reasons that `isValid` actually produced today are modeled. No behavior regression — the `Bool` API is byte-identical, all pre-existing tests in `APIEndpointTests.swift` pass unchanged.

## Release Note

**Settings now explain why an endpoint is rejected** — The API endpoint row used to render a generic "Incomplete" badge whenever `APIEndpoint.isValid` was false, which covered empty URLs, malformed URLs, and insecure HTTP remotes with the same opaque label. A new typed `APIEndpoint.validate()` returns a specific `APIEndpointValidationReason` (`emptyURL`, `malformedURL`, `insecureScheme`), and the settings row surfaces its `LocalizedError.errorDescription` as a subtitle plus VoiceOver value. `isValid: Bool` is retained as a derived convenience so existing callers are unaffected.

## Test plan

- [x] `swift test --filter BaseChatCoreTests --disable-default-traits` (101 tests, 0 failures — includes 18 new `APIEndpointValidationTests`)
- [x] `swift test --filter BaseChatInferenceTests --disable-default-traits` (69 tests, 0 failures)
- [x] `swift test --filter BaseChatUITests --disable-default-traits` (431 tests, 0 failures)
- [x] `swift test --filter BaseChatBackendsTests --disable-default-traits` (63 tests, 0 failures)
- [x] Sabotage verified: remapping `.insecureScheme` to `.malformedURL` in `validate()` fails exactly the three `*_insecureScheme` cases with `XCTAssertEqual failed: ("malformedURL") is not equal to ("insecureScheme")` — reverted before commit.

Co-authored-by: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/Sources/BaseChatCore/BaseChatSchemaV3.swift
+++ b/Sources/BaseChatCore/BaseChatSchemaV3.swift
@@ -266,7 +266,7 @@ public enum BaseChatSchemaV3: VersionedSchema {
             try KeychainService.delete(account: keychainAccount)
         }
 
-        /// Validates the endpoint's URL structure.
+        /// Validates the endpoint's URL structure and returns a typed result.
         ///
         /// This is a pure structural check — it does NOT verify whether an API key
         /// exists in the Keychain. Use `APIProvider.requiresAPIKey` and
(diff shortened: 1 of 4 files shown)


roryford:
Rebased on `origin/main` to pick up the SSRF hardening that landed in #360, and extended `APIEndpointValidationReason` so every rejection branch in `APIEndpoint.validate()` now maps to a specific, user-facing reason.

### Enum changes

Three original cases (`.emptyURL`, `.malformedURL`, `.insecureScheme`) plus six new ones:

- `.unsupportedScheme(String)` — non-http/https scheme that still parses (e.g. `ftp://example.com`, `ws://…`, `gopher://…`). Carries the offending scheme for display.
- `.privateHost` — RFC1918 IPv4 (`10/8`, `172.16/12`, `192.168/16`).
- `.linkLocalHost` — `169.254/16` (AWS/GCP/Azure IMDS) and IPv6 `fe80::/10`.
- `.ipv6UniqueLocal` — `fc00::/7`.
- `.ipv4MappedLoopback` — the whole `::ffff:0:0/96` mapped range (covers mapped loopback and mapped RFC1918 — rejected as a single class so the mapped form can't bypass either filter).
- `.multicastReserved` — `0/8`, alternate-encoding loopback `127/8` (≠ `127.0.0.1`), `224/4` multicast, `240/4` reserved (incl. `255.255.255.255`).

### Implementation notes

- `isDisallowedPrivateHost` / `isDisallowedIPv4` / `isDisallowedIPv6` renamed to `classify…` variants that return the reason, so `validate()` only walks the classifier once.
- `isValid` stays `public` and derives from `validate()` — no API break for existing callers.
- `APIEndpointRow` already consumed `reason.errorDescription` generically, so the UI picks up the new messages without any changes.
- `LocalizedError` messages cite concrete ranges (e.g. "Private IP addresses (10.x, 172.16–31.x, 192.168.x) are not allowed.") so users know what to fix.

### Test rework

`APIEndpointValidationTests.swift` now pins the mapping for every case with per-branch `.failure(.X)` assertions. Found one subtle thing worth noting: `file:///`, `data:`, and `javascript:` URLs produce a `URL` with `host == nil`, so they trip `.malformedURL` *before* reaching the scheme filter. The tests assert that honestly (still rejected, `isValid == false`) and the `.unsupportedScheme` coverage uses `ftp://`, `ws://`, `gopher://` which do parse a host.

Sabotage-verified: remapping `a == 10` in `classifyIPv4` to return `.linkLocalHost` correctly fails `test_validate_rfc1918_10_x_privateHost`. Reverted.

`CustomEndpointValidationTests.swift` (from #360) left untouched — its `isValid == false` assertions still hold, and the typed coverage lives in the new file so we avoid duplication.

### Pre-push checks

All four CI suites green locally:

- `BaseChatCoreTests`: 176 tests, 4 skipped, 0 failures
- `BaseChatInferenceTests`: 69 tests, 0 failures
- `BaseChatUITests`: 431 tests, 0 failures
- `BaseChatBackendsTests`: 82 tests, 0 failures

Commit: `7d7a5f5`.

PR #372, opened 2026-04-14
feat(security): reap orphaned Keychain items on boot

roryford:
## Summary

- Adds a boot-time reaper that deletes Keychain entries whose owning `APIEndpoint` row no longer exists.
- Wires it into `SwiftDataPersistenceProvider.init` so host apps using the default persistence get it automatically; exposes `BaseChatBootstrap.reapOrphanedKeychainItems(in:)` for apps with custom providers.
- Gated by `BaseChatConfiguration.keychainReaperEnabled` (default `true`) for test harnesses that populate the namespace independently.

## Why

`APIConfigurationView.deleteEndpoint` deletes the SwiftData row even if the paired `KeychainService.delete` call silently fails — the right trade-off for the user's flow, but it leaks an API key into the Keychain that nothing can ever reclaim. Direct SwiftData wipes (migrations, debug tools, test cleanup) have the same effect. Over time these accumulate, and because the row that referenced them is gone there is no way to find the dangling keys.

The reaper closes this hole: at boot it enumerates every account in the framework's Keychain service namespace, diffs against the current set of `APIEndpoint.keychainAccount` values from SwiftData, and deletes the difference. One pass at boot, not per-view, not per-generation.

`KeychainService.sweep(validAccounts:)` never aborts on a single failed delete — it logs at `.warning` and continues, so a single stuck row can't block reclaiming the rest. Keychain access denial (sandboxed contexts, missing entitlement) returns `0` cleanly; the reaper never crashes the app.

## Opt-out

```swift
BaseChatConfiguration.shared.keychainReaperEnabled = false
```

Intended for test fixtures that populate the framework's Keychain namespace independently of `APIEndpoint` rows. Production apps should leave this enabled.

## Test plan

- [x] `KeychainServiceSweepTests` — 7 new tests covering `allAccounts` and `sweep` round-trips, empty/populated/all-valid/unknown-valid edge cases; each test scopes its own service name so parallel runs don't collide.
- [x] Sabotage check — reverse the `!validAccounts.contains(account)` guard: five of the seven tests correctly fail; restored before committing.
- [x] `BaseChatCoreTests` (83 tests) — pass.
- [x] `BaseChatInferenceTests` (651 tests, including existing `KeychainServiceTests` and `KeychainIntegrationTests`) — pass.
- [x] `BaseChatUITests` (431 tests) — pass.
- [x] `BaseChatBackendsTests` (131 tests, including `CloudEndpointSelectionIntegrationTests` which re-runs `SwiftDataPersistenceProvider.init` 8 times with Keychain entries live in the same process) — pass.

## Release Note

**Orphaned Keychain cleanup on app launch** — API keys whose owning cloud endpoint was deleted (either via a Keychain failure during row deletion or a direct SwiftData wipe) could previously accumulate indefinitely in the Keychain with no way for the app to reclaim them. BaseChatKit now runs a one-shot sweep at boot that removes any Keychain entry in the framework's namespace whose account ID is not present in the current set of `APIEndpoint` rows. The sweep is opt-out via `BaseChatConfiguration.keychainReaperEnabled`, logs failures per-item without aborting, and exits cleanly if Keychain access is denied. Host apps using `SwiftDataPersistenceProvider` get this automatically; apps with custom persistence can call `BaseChatBootstrap.reapOrphanedKeychainItems(in:)` from their launch path.

### DIFF
--- a/Sources/BaseChatCore/BaseChatBootstrap.swift
+++ b/Sources/BaseChatCore/BaseChatBootstrap.swift
@@ -1,0 +1,49 @@
+import Foundation
+import SwiftData
+import BaseChatInference
+
+/// One-time boot hooks that bridge SwiftData persistence and the framework's
+/// non-persistent services.
+///
+/// Call the entry points here once at app launch — they are idempotent at the
+/// boot-once granularity, but there is no caching; repeated calls do repeated
+/// work. Host apps that use ``SwiftDataPersistenceProvider`` get the default
+/// hooks wired automatically and should not call these directly.
+public enum BaseChatBootstrap {
+
+    /// Removes Keychain items whose owning ``BaseChatSchemaV3/APIEndpoint`` row
+    /// no longer exists.
+    ///
+    /// Orphans accumulate when an endpoint row is deleted while the matching
+    /// Keychain delete silently fails, or when rows are wiped directly through
+    /// SwiftData without routing through the UI. The reaper compares every
+    /// account in the framework's Keychain namespace against the current set of
+    /// endpoint IDs and deletes anything that no longer has an owner.
+    ///
+    /// The sweep is a no-op when
+    /// ``BaseChatInference/BaseChatConfiguration/keychainReaperEnabled`` is
+    /// `false`. Errors (including Keychain access denial in sandboxed contexts)
+    /// are logged and swallowed so a boot hook can never crash the app.
+    ///
+    /// Fire-and-forget — call once per app boot. Returns the number of items
+    /// that were actually reaped, for testing and diagnostics.
+    @discardableResult
+    @MainActor
+    public static func reapOrphanedKeychainItems(in modelContext: ModelContext) -> Int {
+        guard BaseChatConfiguration.shared.keychainReaperEnabled else {
+            return 0
+        }
+
+        let validAccounts: Set<String>
+        do {
+            let descriptor = FetchDescriptor<BaseChatSchemaV3.APIEndpoint>()
+            let endpoints = try modelContext.fetch(descriptor)
+            validAccounts = Set(endpoints.map(\.keychainAccount))
+        } catch {
+            Log.security.warning("BaseChatBootstrap.reapOrphanedKeychainItems: failed to fetch APIEndpoint rows — skipping reap: \(error.localizedDescription)")
+            return 0
+        }
+
+        return KeychainService.sweep(validAccounts: validAccounts)
+    }
+}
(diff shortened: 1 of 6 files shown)


roryford:
Rebased onto `origin/main` (7d890a4). Adaptations for the now-merged throws signature (#363):

**Conflict resolutions**
- `BaseChatConfiguration.swift` — merged both #371's `fileProtectionClass` and this PR's `keychainReaperEnabled`. Public init gained both params; field/init ordering matches `#371`'s convention (features → fileProtectionClass → keychainReaperEnabled).
- `Services/KeychainService.swift` — auto-merged the new throwing `store` / `delete`; manually rewrote the sweep loop to drive the new API:
  ```swift
  for account in stored where !validAccounts.contains(account) {
      do {
          try KeychainService.delete(account: account)
          reaped += 1
      } catch {
          Log.security.warning("KeychainService.sweep: failed to delete orphaned account: \(error.localizedDescription, privacy: .public)")
          continue
      }
  }
  ```
  Semantics unchanged: individual delete failure logs at `.warning` and does not abort the sweep.

**Test adaptation**
- `Tests/BaseChatInferenceTests/KeychainServiceSweepTests.swift` — every `KeychainService.store` call now uses `try` (test methods marked `throws`), and the `tearDown` sweep is explicitly discarded (`_ = KeychainService.sweep(...)`) so the throws-change doesn't ripple into a warning.

**Product-review answers**
- `sweep(validAccounts:)` is called cross-module from `BaseChatCore/BaseChatBootstrap.reapOrphanedKeychainItems` — `public` remains the minimum viable visibility. Confirmed by `grep KeychainService.sweep`: only the bootstrap entry point and same-module tests call it.
- Sweep count is surfaced via `Log.security.info("…reaped N orphaned Keychain item(s) from M stored")` on every non-empty run, plus the `@discardableResult Int` return is threaded through `reapOrphanedKeychainItems` for tests/diagnostics. No UI callback added — correct scope for a silent boot hook.
- Sync-in-`init` path unchanged post-rebase — still guarded by the `Sendable`-avoidance comment and fires before any SwiftData work.
- `keychainReaperEnabled` default-`true` flag preserved; doc-comment explicitly covers the opt-out case ("test harnesses that populate the framework's Keychain namespace independently").

**Verification**
- Sabotage: inverted the `where !validAccounts.contains(account)` guard → sweep preserves instead of reaps. 12 assertions across 2 tests failed as expected. Reverted.
- Full CI matrix locally on arm64 (all `--disable-default-traits`):
  - `BaseChatCoreTests` — 127 passed, 4 skipped
  - `BaseChatInferenceTests` — 697 passed (includes all 7 new `KeychainServiceSweepTests`)
  - `BaseChatUITests` — 431 passed
  - `BaseChatBackendsTests` — 133 passed

**SHAs**
- Before rebase: `72a15ea0843f1487a15fc0209d68dead8ea44435`
- After rebase:  `65e3d52`
- Rebased onto: `7d890a4` (origin/main)

Package.resolved deliberately untouched.

Issue #208, opened 2026-04-07
macOS: ModelManagementSheet Select tab renders empty (List collapses in VStack)


roryford:
## Bug

On macOS, opening the **Change model** sheet shows the segmented tab picker correctly, but the **Select** tab content is blank — no model rows appear even when models are available.

## Root cause

Two issues in `Sources/BaseChatUI/Views/Models/ModelManagementSheet.swift`:

### 1. `tabContent` has no height constraint in the macOS `VStack`

The macOS branch wraps `tabPickerBar` + `tabContent` in a `VStack`. Without `frame(maxHeight: .infinity)`, the inner `List` collapses to zero height — rows exist but are invisible.

```swift
// ModelManagementSheet.swift — macOS branch
VStack(spacing: 0) {
    tabPickerBar
    tabContent
        .frame(maxHeight: .infinity)   // ← missing
}
```

### 2. No macOS `listStyle` on `ModelSelectTab`

`ModelSelectTab` only applies `.insetGrouped` on iOS. macOS defaults to `.sidebar` style which behaves differently inside a plain `VStack`/sheet context.

```swift
// ModelSelectTab.body
#if os(iOS)
.listStyle(.insetGrouped)
#else
.listStyle(.plain)   // ← missing
#endif
```

## Fix

Apply both changes above. Fix 1 is the blocker; fix 2 is a consistency improvement.

## Latent bug (same file)

`ContentUnavailableView` is used as a `List` row item in `ModelSelectTab` for the empty-state. On macOS, `ContentUnavailableView` is designed as a full-screen view and renders with near-zero height inside a `List` row. Should be replaced with a plain `Text` or `VStack` empty state for the list row context.

PR #381, opened 2026-04-14
fix: render Model Management sheet tab content on macOS (#378)

### LINKED ISSUE #378: macOS: Model Management sheet — Select/Download/Storage tabs render blank, no content below segmented control
## Summary

On native macOS (26.4.1, Apple Silicon) the Model Management sheet opens, but the area below the tab picker is completely empty — no content renders for any of the three tabs (Select, Download, Storage). This makes the demo app unusable as a real user: there is no way to select or download a model.

## Repro

1. Build and launch `BaseChatDemo` natively on macOS (not Mac Catalyst, not iPad).
   ```
   cd Example
   xcodebuild -project BaseChatDemo.xcodeproj -scheme BaseChatDemo -destination 'platform=macOS,arch=arm64' build
   open $(xcodebuild -project BaseChatDemo.xcodeproj -scheme BaseChatDemo -destination 'platform=macOS,arch=arm64' -showBuildSettings build 2>/dev/null | awk '/BUILT_PRODUCTS_DIR/ {print $3}')/BaseChatDemo.app
   ```
2. Click **Select Model** (or the sidebar Model Management entry point).
3. The sheet opens showing: a title, a \"Section\" label, three tab buttons **Select / Download / Storage**, and a **Done** button at the bottom right.
4. Click any tab.

## Expected

- **Select** tab: list of locally-available models (on my machine: `Llama-3.2-3B-Instruct-4bit` MLX + `Qwen_Qwen3-4B-Q4_K_M.gguf` + Apple Foundation Model).
- **Download** tab: \"Recommended for Your Device\" section + HuggingFace search field + curated models.
- **Storage** tab: Storage Overview + Total Used + Models Directory.

## Actual

Tab content area is **completely blank** for all three tabs. Only the segmented control and Done button are visible. Nothing else renders.

(See attached screenshot — sheet title \"Download\", segmented control with Download highlighted, and a large empty region below.)

## Impact

- Users cannot select, download, or manage models on native macOS.
- The demo app is effectively non-functional as a real chat app on desktop.
- All existing \`BaseChatDemoUITests\` that depend on tab content existing are presumably ineffective on macOS (separately tracked as #377: window isn't even exposed to XCUITest on macOS).

## Environment

- macOS 26.4.1 (25E253)
- Apple Silicon
- Main branch: \`f9ab8c5\` (v0.8.0)
- Not in \`--uitesting\` mode — this is the real app running normally.

## Hypotheses (to investigate)

1. Tab content views may be wrapped in a container that collapses to zero height on macOS (e.g. a Form/List with no embedded ScrollView, or a TabView-variant that doesn't render content on macOS).
2. Geometry / frame constraint issue — content may be rendering at 0×0.
3. Recent macOS-specific layout change (possibly correlated with \`topBarLeading\` being iOS-only, fixed in PR #376 — might be more conditional-compilation gaps of the same kind in \`ModelManagementSheet.swift\` or its child views).
4. SwiftUI \`.presentationDetents()\` / \`.presentationSizing\` interplay with macOS sheet sizing.

## Starting points

- \`Sources/BaseChatUI/Views/Models/ModelManagementSheet.swift\` — owns the tab picker and switches content based on selection.
- \`Sources/BaseChatUI/Views/Models/HuggingFaceBrowserView.swift\` — Download tab content.
- \`Sources/BaseChatUI/Views/Models/LocalModelStorageView.swift\` — Select tab content.
- \`Sources/BaseChatUI/Views/Models/StorageManagementView.swift\` — Storage tab content.

## Priority

**Blocker.** Affects every macOS user of the demo app. Should be fixed before the next release.

roryford:
## Summary

On native macOS 26, opening the Model Management sheet shows the Select / Download / Storage picker and the Done button, but the content area below is blank — for all three tabs. The demo app is effectively unusable: users can't select, download, or manage models. See #378.

## Root cause

`ModelManagementSheet` wraps content in `NavigationStack { VStack { tabPickerBar; tabContent } }` on macOS. The `tabContent` is always a `List` (`ModelSelectionTabView`, `HuggingFaceBrowserView`, `LocalModelStorageView`). macOS sheets don't honor `.presentationDetents`, and a `NavigationStack { VStack { List } }` tree has no intrinsic size — so without an explicit frame, the inner list collapses to zero height and every tab renders blank below the picker bar.

The previous fix for #208 added `.frame(maxHeight: .infinity)` to `tabContent`, but that only works inside a VStack whose parent offers finite height — here the NavigationStack in a sheet offers zero, so the fix had nothing to expand into.

## Fix

Give the macOS sheet an explicit `.frame(minWidth: 560, idealWidth: 720, minHeight: 480, idealHeight: 640)`, and let the inner VStack + tab content expand to fill it. iOS is unchanged — it uses `.presentationDetents` on iPhone and iPad, which drive sheet sizing there.

## Verification

Built and launched `BaseChatDemo.app` for native macOS (`xcodebuild -destination 'platform=macOS,arch=arm64'`). Cycled through all three tabs — each now shows real content:

- **Select tab:** model rows render (Foundation Model, Llama, GGUF / MLX entries with type and tier badges)
- **Download tab:** recommended models section with SmolLM, Phi-4, Mistral 7B, Llama 3.2, Qwen 2.5 rows
- **Storage tab:** Storage Overview with total used, Models Directory link, Downloaded Models list with delete buttons

Screenshots of each tab were captured during verification and confirm content now renders. The existing iOS Simulator `ModelManagementUITests` suite still passes 10/10 — no regression on the iOS path.

## Release Note

**Model Management sheet renders on macOS.** A SwiftUI sizing interaction between `NavigationStack`, `VStack`, and `List` inside a macOS sheet left the Select, Download, and Storage tabs blank for host apps that used `ModelManagementSheet` on native macOS 26. The sheet now carries an explicit minimum frame on macOS so its content lays out correctly, without affecting the iOS or iPad presentation that uses `.presentationDetents`.

Closes #378.

### DIFF
--- a/Sources/BaseChatUI/Views/Models/ModelManagementSheet.swift
+++ b/Sources/BaseChatUI/Views/Models/ModelManagementSheet.swift
@@ -60,8 +60,9 @@ public struct ModelManagementSheet: View {
             VStack(spacing: 0) {
                 tabPickerBar
                 tabContent
-                    .frame(maxHeight: .infinity)
+                    .frame(maxWidth: .infinity, maxHeight: .infinity)
             }
+            .frame(maxWidth: .infinity, maxHeight: .infinity)
             .navigationTitle(selectedTab.rawValue)
             .toolbar {
                 doneToolbarItem
@@ -82,7 +83,11 @@ public struct ModelManagementSheet: View {
         .presentationDetents(horizontalSizeClass == .regular ? [.medium, .large] : [.large])
         .presentationDragIndicator(.visible)
         #else
-        .presentationDetents([.large])
+        // macOS sheets don't honor `.presentationDetents` and don't get a
+        // useful intrinsic size from a `NavigationStack { VStack { List } }`
+        // tree — without an explicit frame, the inner List collapses to zero
+        // height and every tab renders blank below the picker. See #378.
+        .frame(minWidth: 560, idealWidth: 720, minHeight: 480, idealHeight: 640)
         #endif
         .onAppear {
             if !availableTabs.contains(selectedTab) {


roryford:
## macOS verification screenshots

Built `BaseChatDemo` for native macOS (`xcodebuild -destination 'platform=macOS,arch=arm64'`) and cycled through all three tabs after this fix:

- Select tab — model rows render (Foundation, MLX, GGUF rows with type + tier badges)
- Download tab — recommended models render (SmolLM, Phi-4, Mistral 7B, Llama 3.2, Qwen 2.5)
- Storage tab — Storage Overview, Models Directory link, Downloaded Models list with delete buttons

Without the fix every tab area was blank below the segmented control. Screenshots captured locally during verification (paths `/tmp/378-tab-{select,download,storage}.png`); attaching here would require uploading to GitHub LFS or as gist attachments, which `gh` doesn't support directly. If you want them inline, drop them onto the PR via the web UI from `/tmp/378-tab-*.png` on this dev machine.

iOS Simulator regression check (`scripts/example-ui-tests.sh test-without-building -only-testing:BaseChatDemoUITests/ModelManagementUITests`): **10 passed / 0 failed** — no iOS regression.

PR #396, opened 2026-04-15
fix: clear LlamaBackend KV cache at generate start, not end

### LINKED ISSUE #390: LlamaBackend: stopGeneration leaves KV cache dirty, next generate() fails
## Problem

After calling `LlamaBackend.stopGeneration()`, the next call to `generate()` on the same loaded model throws `InferenceError.inferenceFailure("Failed to decode prompt")`. This breaks any consumer that wants to interrupt a running generation and immediately start a new one — a normal UX pattern (user hits "stop", then sends a new message).

## Root cause

`LlamaBackend.generate()` ends with a conditional KV cache clear at `Sources/BaseChatBackends/LlamaBackend.swift:438-441`:

```swift
// Clear KV cache for next generation (skip if cancelled/unloaded)
if !self.withStateLock({ self.cancelled }), let memory = llama_get_memory(context) {
    llama_memory_clear(memory, false)
}
```

The `cancelled` check is there to guard against an unload race, but it's also true on the `stopGeneration()` path (which just sets `cancelled = true` without tearing down the context). So interrupted generations skip the clear, leaving the KV cache populated with tokens from the prior run at positions `0..N`. The next `generate()` tokenizes a fresh prompt, writes to positions `0..M`, and `llama_decode` fails because those positions are already occupied.

## Reproduction

`Tests/BaseChatE2ETests/LlamaE2ETests.swift` surfaced this. When the stop-generation test ran before any other real-inference test, every subsequent test failed to decode. Workaround in that suite was to rename the test `test_zzz_realInference_stopGeneration` so it runs last alphabetically — that's a test-order hack, not a fix.

## Suggested fix

Clear the KV cache at the **start** of `generate()`, unconditionally, rather than conditionally at the end. The context is guaranteed valid at generate-start (there's a `guard isModelLoaded, let context, ...` check), so there's no unload-race exposure. Remove the conditional tail clear at the same time.

## Impact

- User-visible: stop-then-resend is broken in any consumer holding a long-lived `LlamaBackend`.
- Test suite: any real-GGUF E2E suite has to hand-order tests around the dirty-state window.

roryford:
## Problem

After calling `LlamaBackend.stopGeneration()`, the next call to `generate()` on the same loaded model throws `InferenceError.inferenceFailure("Failed to decode prompt")`. The normal UX pattern — user hits Stop, then sends a new message — was broken.

## Root cause

`generate()` ended with a conditional KV cache clear:

```swift
if !self.withStateLock({ self.cancelled }), let memory = llama_get_memory(context) {
    llama_memory_clear(memory, false)
}
```

The `cancelled` guard was meant to protect against an unload race, but `stopGeneration()` sets the same flag without tearing down the context. Interrupted generations therefore skipped the clear, leaving the KV cache populated with prior-run tokens. The next `generate()` tokenized a fresh prompt, wrote to positions `0..M`, and `llama_decode` failed because those positions were already occupied.

## Fix

Move the KV cache clear to the top of `generate()`, immediately after the `guard isModelLoaded, let context, ...` check. The context is guaranteed non-nil there, so the unload-race the old tail clear was guarding against doesn't apply. The trailing conditional clear is removed.

## Test

Adds `test_stopGeneration_thenGenerate_succeeds_regression390` to `LlamaBackendTests`. It loads a real GGUF model, starts a generation, stops mid-stream, and confirms the next `generate()` produces tokens. Skips cleanly with `XCTSkip` when no GGUF >= 50 MB is found under `~/Documents/Models/` (CI has no model, and the existing ~1 KB test fixtures in that directory are filtered out by the size gate).

Sabotage-verified locally with SmolLM2 135M:
- With the fix: passes in ~0.13 s.
- Reverting the new clear: fails with `inferenceFailure("Failed to decode prompt")` — the exact error from #390.

Also adds `HardwareRequirements.findGGUFModel()` (same search paths as `findMLXModelDirectory`) so other real-model tests can reuse the discovery logic.

## Test suites (local, Apple Silicon)

- `BaseChatCoreTests --disable-default-traits`: 176 tests (4 skipped, 0 failures)
- `BaseChatInferenceTests --disable-default-traits`: 704 XCTest + 69 Swift Testing (0 failures)
- `BaseChatUITests --disable-default-traits`: 432 tests (0 failures)
- `BaseChatBackendsTests --disable-default-traits`: 133 XCTest + 84 Swift Testing (0 failures)
- `BaseChatBackendsTests --traits MLX,Llama`: 184 XCTest + 84 Swift Testing (2 skipped, 0 failures)
- Regression test run in isolation with a real GGUF model: 3/3 passes.

`BaseChatE2ETests` was not run — the file exists but uses `MockInferenceBackend`, not real Llama, so it wouldn't exercise this path.

Closes #390

## Release Note

**LlamaBackend stop-then-regenerate no longer fails** — Hitting Stop during a llama.cpp generation used to leave KV cache state from the interrupted run, so the next message from the user failed with "Failed to decode prompt" until the model was reloaded. The cache is now cleared at the start of each generation instead of conditionally at the end, so the Stop / new-message cycle works as expected without touching the model lifecycle.

### DIFF
--- a/Sources/BaseChatBackends/LlamaBackend.swift
+++ b/Sources/BaseChatBackends/LlamaBackend.swift
@@ -325,6 +325,14 @@ public final class LlamaBackend: InferenceBackend, @unchecked Sendable {
         }
         Self.logger.debug("Llama generate started")
 
+        // Clear KV cache at the start so state from any prior run (including one
+        // terminated by stopGeneration) can't collide with this run's positions.
+        // Context is guaranteed non-nil here by the guard above, so there's no
+        // unload-race exposure that the old tail clear was trying to avoid.
+        if let memory = llama_get_memory(context) {
+            llama_memory_clear(memory, false)
+        }
+
         // Tokenize prompt
         let tokens = tokenize(prompt, addBos: true)
         guard !tokens.isEmpty else {
@@ -434,11 +442,6 @@ public final class LlamaBackend: InferenceBackend, @unchecked Sendable {
             }
 
             await MainActor.run { generationStream.setPhase(.done) }
-
-            // Clear KV cache for next generation (skip if cancelled/unloaded)
-            if !self.withStateLock({ self.cancelled }), let memory = llama_get_memory(context) {
-                llama_memory_clear(memory, false)
-            }
 
             continuation.finish()
         }
--- a/Sources/BaseChatBackends/LlamaBackend.swift
+++ b/Sources/BaseChatBackends/LlamaBackend.swift
@@ -312,7 +312,12 @@ public final class LlamaBackend: InferenceBackend, @unchecked Sendable {
         systemPrompt: String?,
         config: GenerationConfig
     ) throws -> GenerationStream {
-        guard isModelLoaded, let context, let vocab, model != nil else {
+        // The Task body re-reads context under stateLock below to avoid the
+        // use-after-free window between here and `self.generationTask = task`
+        // install. We only need to verify model is loaded up front — the
+        // captured pointers are accessed through the re-read, not these
+        // outer locals.
+        guard isModelLoaded, context != nil, vocab != nil, model != nil else {
             throw InferenceError.inferenceFailure("No model loaded")
         }
         guard !withStateLock({ isGenerating }) else {
@@ -325,15 +330,7 @@ public final class LlamaBackend: InferenceBackend, @unchecked Sendable {
         }
         Self.logger.debug("Llama generate started")
 
-        // Clear KV cache at the start so state from any prior run (including one
-        // terminated by stopGeneration) can't collide with this run's positions.
-        // Context is guaranteed non-nil here by the guard above, so there's no
-        // unload-race exposure that the old tail clear was trying to avoid.
-        if let memory = llama_get_memory(context) {
-            llama_memory_clear(memory, false)
-        }
-
-        // Tokenize prompt
+        // Tokenize prompt (pure vocab lookup — doesn't touch context KV state)
         let tokens = tokenize(prompt, addBos: true)
         guard !tokens.isEmpty else {
             withStateLock { isGenerating = false }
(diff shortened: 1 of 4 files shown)


roryford:
Addressed the four Copilot review comments in 0634310.

**1. LlamaBackend KV clear race (use-after-free)**

Moved `llama_memory_clear` into the generation Task body and tightened the lifecycle so `unloadModel()` can no longer free the context while the clear (or any later decode) is in flight.

Specifically:
- The parent's outer `guard let context` binding is gone; the Task body now re-reads `(context, vocab)` under `stateLock` as its first action and bails if either is nil.
- The parent holds `stateLock` across both the Task creation AND the `self.generationTask = task` assignment. Because the Task body's re-read must acquire the same NSLock, we get the invariant: by the time the Task body reads context, either (a) `generationTask` is installed (so `unloadModel` will observe it and await the task before calling `llama_free`), or (b) `unloadModel` ran fully first and has already nil'd `self.context` (so the Task bails cleanly).
- Investigation concluded both fixes were needed. Just moving the clear into the Task body wasn't sufficient on its own: `unloadModel` captures `generationTask` under `stateLock`, but the old code assigned `self.generationTask = task` outside `stateLock` and after `Task { }` was created, so the Task could start work before `unloadModel` could observe the installed task. Installing under `stateLock` (and gating the Task body's first C access behind the same lock) closes the window.

**2. `findGGUFModel` directory false-positives**

Added the explicit `isRegularFile` check as Copilot suggested. Directories ending in `.gguf` are now rejected before the size gate.

**3. `findGGUFModel` doc comment**

Updated the doc to say "first regular `.gguf` file >= 50 MB" with a one-line rationale about filtering out test fixtures.

**4. Flaky poll loop in regression test**

Replaced the 500 ms / fixed-50-iteration poll with a `ContinuousClock` deadline of 2 seconds.

**Verification**

All four CI suites green locally:
- `BaseChatCoreTests`: 176 tests, 0 failures (4 skipped)
- `BaseChatInferenceTests`: 704 tests, 0 failures
- `BaseChatUITests`: 432 tests, 0 failures
- `BaseChatBackendsTests`: 133 tests, 0 failures (CI baseline, no traits)

Additionally ran with `--traits Llama` on Apple Silicon: 157 tests, 0 failures, and the `test_stopGeneration_thenGenerate_succeeds_regression390` test passed across multiple consecutive runs (both standalone and in-suite) against a real GGUF, confirming the Task-internal KV clear still fixes #390.

Issue #398, opened 2026-04-15
Investigate iPad crash when loading long-context GGUF models


roryford:
## Summary

Loading some local models on iPad appears to crash during model load, with the strongest current evidence pointing to the GGUF/llama load path over-allocating context/KV cache.

## Current findings

1. `ChatViewModel+ModelLoading` currently passes `model.detectedContextLength ?? 2048` into `InferenceService.loadModel(...)`, so GGUF models can be loaded at their full trained context by default.
2. `LlamaBackend.initializeModel(...)` clamps using a `ramSafeCap` derived from total physical memory with a hard ceiling of 128k, which does not meaningfully constrain long-context models on modern iPads.
3. `MemoryGate` preflight checks model-file cost, but for `.mappable` backends it does not account for the large KV/context allocation implied by the requested context size.
4. This makes GGUF a much stronger suspect than MLX for the iPad crash: a large-context GGUF can pass preflight and then fail during llama initialization/allocation.

## Likely affected files

- `Sources/BaseChatUI/ViewModels/ChatViewModel+ModelLoading.swift`
- `Sources/BaseChatBackends/LlamaBackend.swift`
- `Sources/BaseChatInference/Services/MemoryGate.swift`
- `Sources/BaseChatInference/Services/ModelLifecycleCoordinator.swift`

## Investigation goals

1. Reproduce with a long-context GGUF on an iPad-class device or simulator-supported diagnostic path.
2. Determine whether the crash occurs during llama model init, KV-cache allocation, or another backend allocation step.
3. Decide whether the primary fix should be:
   - a safer default GGUF load context,
   - a stronger llama-side context clamp based on available memory/headroom,
   - and/or a better preflight estimate that includes requested context size.
4. Add regression coverage for the chosen guardrails.

## Notes

- No current upstream `main` changes appear to have altered this load path.
- We do not yet have a device crash log attached to confirm the exact failing allocation.

roryford:
Fixed by merged PR #399 ('fix: clamp GGUF context size based on available memory to prevent iPad crashes'). The PR title referenced this issue informally rather than via a Closes keyword, so GitHub did not auto-close it.

PR #399, opened 2026-04-16
fix: clamp GGUF context size based on available memory to prevent iPad crashes (#398)

roryford:
## Summary

- Introduces `DeviceCapabilityService.safeContextSize(for:availableMemoryBytes:)` — a pure, testable helper that derives a safe GGUF context ceiling from available memory (not physical RAM) on iOS/iPadOS, using `os_proc_available_memory()` for the per-app jetsam budget.
- `ChatViewModel+ModelLoading` now calls this helper instead of passing `model.detectedContextLength ?? 2048` straight through to the backend.
- `LlamaBackend` gains `computeRamSafeCap()` (also `os_proc_available_memory`-aware on iOS) and a 3-attempt retry loop that halves `n_ctx` on each `llama_init_from_model` nil return, with a 1 024-token floor.

## Problem

Loading long-context GGUF models (32K–128K) on iPad was crashing. The UI layer sent the model's full trained context size to llama.cpp. `LlamaBackend` capped via `physicalMemory / 8 192`, which is `>= 128 000` on any modern iPad regardless of the per-app jetsam limit (~3 GB on an 8 GB device). A 128K GGUF at 8 KB/token needs ~1 GB of KV cache — pushing the process over its jetsam limit and triggering a SIGKILL during `llama_init_from_model`.

## What changed

| Layer | Before | After |
|---|---|---|
| `ChatViewModel+ModelLoading` | `Int32(model.detectedContextLength ?? 2048)` | `DeviceCapabilityService.safeContextSize(for:)` |
| `LlamaBackend` `ramSafeCap` | `physicalMemory / 8 192` (always ≥ 128 000 on modern devices) | `os_proc_available_memory() / 8 192` on iOS |
| `LlamaBackend` context init | Throw on nil from `llama_init_from_model` | Retry up to 3× with halved `n_ctx`, floor 1 024 |
| Default when `detectedContextLength` is nil | 2 048 | 8 192 (conservative but usable) |

## Testing

- 6 new unit tests for `safeContextSize` in `DeviceCapabilityServiceTests`: clamp below trained length, respect model trained length, nil fallback (8 192 not 2 048), absolute ceiling, available-memory constraint, near-zero floor.
- 2 new unit tests in `LlamaBackendTests`: updated `ramSafeCap` test uses `computeRamSafeCap()` directly; new `retryContextHalving_progressionStopsAtFloor` validates the halving sequence.
- All four CI suites pass with zero failures.

## Deferred

- KV-size-aware `MemoryGate` (threading context size into `MemoryGate.check` for per-architecture KV math).
- Per-architecture KV calculation using `llama_model_n_layer` / `llama_model_n_embd_kv`.

## Release Note

**iPad crash loading long-context GGUF models** — loading a 32K–128K context GGUF on iPad was silently over-allocating KV cache, pushing the app past its jetsam limit and triggering a SIGKILL. The fix makes context sizing jetsam-aware at two layers: `DeviceCapabilityService.safeContextSize()` now queries `os_proc_available_memory()` (the real per-app budget) rather than physical RAM when computing how many context tokens the device can safely support, and `LlamaBackend` does the same for its own RAM-safe cap. A new 3-attempt retry loop in `LlamaBackend` also halves `n_ctx` on clean init failures, providing defense-in-depth for near-threshold cases that return nil rather than being killed outright.

Refs #398

### DIFF
--- a/Sources/BaseChatBackends/LlamaBackend.swift
+++ b/Sources/BaseChatBackends/LlamaBackend.swift
@@ -226,6 +226,32 @@ public final class LlamaBackend: InferenceBackend, @unchecked Sendable {
         }
     }
 
+    /// Returns the maximum context size (in tokens) that is safe to allocate given
+    /// the memory available to this process.
+    ///
+    /// On iOS/iPadOS, `os_proc_available_memory()` returns the per-app jetsam budget —
+    /// the number of bytes the process can allocate before the system kills it. This is
+    /// typically ~3 GB on an 8 GB iPad, regardless of how much physical RAM exists.
+    /// Using physical memory here would produce a meaningless cap (e.g., 128 000 on
+    /// every modern device) and defeat the purpose of the guard.
+    ///
+    /// On macOS there is no per-app jetsam limit, so physical memory is the right ceiling.
+    ///
+    /// 1 token ≈ 8 KB of KV cache; the result is clamped to 128 000 as an absolute ceiling.
+    static func computeRamSafeCap() -> Int32 {
+        let kvBytesPerToken: Int64 = 8_192  // 8 KB/token — matches DeviceCapabilityService
+        #if os(iOS) || os(tvOS) || os(watchOS)
+        let available = os_proc_available_memory()
+        // os_proc_available_memory() can return 0 or negative in edge cases (simulator boot);
+        // fall through to the physical-memory path if that happens.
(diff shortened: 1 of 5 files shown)


roryford:
Follow-ups from review filed: #400 (KV-aware MemoryGate) and #401 (per-architecture KV math).

PR #429, opened 2026-04-18
test: add in-memory SwiftData harness and direct provider tests

roryford:
## Summary

PR 1 of 4 in the [test-coverage plan](../blob/test/w3-inmemory-persistence-harness/.claude/plans/peppy-sprouting-fairy.md). Enforces CLAUDE.md's "do not mock the persistence layer to make tests faster" rule and adds the direct `SwiftDataPersistenceProvider` unit coverage that was previously only transitive through integration tests.

- `InMemoryPersistenceHarness` builds a fresh in-memory `ModelContainer` + real `SwiftDataPersistenceProvider` per call — one line in `setUp`, honest CRUD semantics.
- `ErrorInjectingPersistenceProvider` wraps any real provider to add per-method error injection and call counting. Replaces the features that kept `MockPersistenceProvider` alive.
- 20 new `SwiftDataPersistenceProviderTests` covering CRUD round-trips across every `@Model` in `BaseChatSchemaV3`, pagination ordering (`fetchRecentMessages`, `fetchMessages(for:before:limit:)`), the cascade-scope invariant (`test_deleteMessages_doesNotCascadeAcrossSessions` — the bug shape that silently nukes user data), and the malformed-CSV tolerance on `pinnedMessageIDsRaw`.
- Migrates `PaginationTests`, `ViewModelEdgeCaseTests`, `SessionAutoRenameTests` to the harness + wrapper; deletes `MockPersistenceProvider`.

## Release Note

**Honest persistence tests by default** — `BaseChatTestSupport` now provides `InMemoryPersistenceHarness`, a fresh in-memory `SwiftDataPersistenceProvider` stack that tests can use in place of a hand-rolled mock. Paired with `ErrorInjectingPersistenceProvider` for failure-injection needs, it closes a long-standing test/production divergence where mocked storage could silently drift from the real SwiftData predicates, ordering, and cascade semantics. `MockPersistenceProvider` has been removed; all internal tests have been migrated.

## Test plan

- [x] `swift test --filter BaseChatCoreTests --disable-default-traits` — 196 tests, 0 failures (includes 20 new)
- [x] `swift test --filter BaseChatInferenceTests --disable-default-traits` — 713 tests, 0 failures
- [x] `swift test --filter BaseChatUITests --disable-default-traits` — 436 tests, 0 failures (migration unaffected)
- [x] `swift test --filter BaseChatBackendsTests --disable-default-traits` — 138 tests, 0 failures
- [x] Sabotage check: removing the `sessionID` predicate in `SwiftDataPersistenceProvider.deleteMessages` flips `test_deleteMessages_doesNotCascadeAcrossSessions` to red; reverting restores green.

### DIFF
--- a/Sources/BaseChatCore/Protocols/ChatPersistenceProvider.swift
+++ b/Sources/BaseChatCore/Protocols/ChatPersistenceProvider.swift
@@ -21,8 +21,10 @@ public enum ChatPersistenceError: Error, LocalizedError, Sendable, Equatable {
 
 /// Abstraction over chat persistence, decoupling view models from SwiftData.
 ///
-/// The default implementation is ``SwiftDataPersistenceProvider``. Tests can
-/// substitute ``MockPersistenceProvider`` from `BaseChatTestSupport`.
+/// The default implementation is ``SwiftDataPersistenceProvider``. Tests
+/// should build a real provider over an in-memory container via
+/// ``InMemoryPersistenceHarness`` in `BaseChatTestSupport`, optionally wrapped
+/// in ``ErrorInjectingPersistenceProvider`` when failure injection is needed.
 @MainActor
 public protocol ChatPersistenceProvider: AnyObject, Sendable {
 
--- a/Sources/BaseChatTestSupport/ErrorInjectingPersistenceProvider.swift
+++ b/Sources/BaseChatTestSupport/ErrorInjectingPersistenceProvider.swift
@@ -2,25 +2,19 @@ import Foundation
 import BaseChatCore
 import BaseChatInference
 
-/// In-memory mock of ``ChatPersistenceProvider`` for testing.
+/// Test double that wraps any ``ChatPersistenceProvider`` and adds two
+/// facilities the real provider cannot offer: per-method error injection
+/// and call counting.
 ///
-/// Stores sessions and messages in plain arrays. Thread-safe via `@MainActor`.
+/// Pair it with ``InMemoryPersistenceHarness`` when a test needs to assert
+/// that a view model handles persistence failures correctly, or that it
+/// routed through the persistence layer the expected number of times.
+/// All other persistence behaviour (CRUD, ordering, pagination) is the
+/// real provider's, so tests written against this wrapper stay honest.
 @MainActor
-public final class MockPersistenceProvider: ChatPersistenceProvider {
+public final class ErrorInjectingPersistenceProvider: ChatPersistenceProvider {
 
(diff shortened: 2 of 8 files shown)


roryford:
Review pass — LGTM, no changes needed.

**What I checked:**
- Diff against `main` (only the 8 claimed files; no scope creep or leaked debug edits)
- `InMemoryPersistenceHarness.swift` + `ErrorInjectingPersistenceProvider.swift` surface vs. the deleted `MockPersistenceProvider.swift`
- All three migrated UI tests (`PaginationTests`, `ViewModelEdgeCaseTests`, `SessionAutoRenameTests`) for semantic equivalence
- All 20 new `SwiftDataPersistenceProviderTests` for logic, ordering expectations, and duplication
- Residual `MockPersistenceProvider` references across `Sources/`, `Tests/`, and root-level docs — zero remain (ignoring `.worktrees/`)
- All four CI suites with `--disable-default-traits` locally: **Core 196 / Inference 69 / UI 436 / Backends 84 — all green**

**Notes (none blocking):**
- `ErrorInjectingPersistenceProvider` omits `shouldThrowOnUpdateMessage`, `shouldThrowOnDeleteMessage`, and `shouldThrowOnDeleteSession` slots. The deleted `MockPersistenceProvider` also lacked these, so this is parity, not a regression. Call counts are complete. Add the slots if a future test needs them.
- `shouldThrowOnFetchMessages` intentionally collapses the three fetch variants into a single slot — same as the old mock. Fine.
- `isInMemoryStore(_:)` is exposed as public API but only consumed by the single self-test `test_harness_isInMemoryStore`. The `/dev/null` detection correctly matches the private helper in `ModelContainerFactory`. Harmless, kept as an explicit failure signal rather than a `precondition` per the author's doc comment.
- `ViewModelEdgeCaseTests.persistenceStack` retention pattern (to keep `ModelContainer` alive past the helper's return) is the right fix for the SIGTRAP footnote in the harness — the earlier failure mode documented in the doc comment is real and the workaround lands correctly.
- `test_clearChat_whenPersistenceDeleteFails_reloadsPersistedMessages` now inserts the session explicitly and re-fetches via `provider.fetchMessages(for:)` instead of reading a `messages` property. The re-fetch correctly reads whichever session ended up active (in case `sendMessage` auto-created one), and the post-assertion still verifies persistence retained the pre-clear state. Semantics preserved.
- Sabotage check claimed for `test_deleteMessages_doesNotCascadeAcrossSessions` is the right priority — that's the highest-consequence invariant in the new file.

Ready to merge from my side.

Issue #492, opened 2026-04-18
fuzz: multi-turn / session-replay sequences


roryford:
## Why

Today's fuzzer drives one chat turn against a fresh-ish backend. Many real bugs surface only across turns: edit-message + regenerate, branch + retry, stop-and-resume. \`InferenceService\`'s sequential-FIFO queue, latest-wins load tokens, and per-priority preemption are entirely unexercised by the v1 single-turn fuzzer.

## What

Add a \`SessionScript\` corpus type alongside \`CorpusEntry\`:

\`\`\`json
{ "id": "edit-then-regenerate", "steps": [
  { "kind": "send", "text": "What is 17*23?" },
  { "kind": "stop" },
  { "kind": "edit", "messageIndex": 0, "newText": "What is 18*23?" },
  { "kind": "regenerate" }
]}
\`\`\`

Drive scripts through \`InferenceService.enqueue\` (not the bare backend) so the queueing/cancellation semantics are exercised. Capture per-step records and run detectors over the joined sequence.

## Why this matters

QA review flagged this as the largest coverage gap. Turn-1 fuzzing cannot reach KV-cache state corruption, race conditions in stop-then-regenerate, or session-scoped \`discardRequests(notMatching:)\` paths.

## Order

Independent of \`--replay\` and the calibration corpus, but probably worth landing after both so multi-turn findings can be reproduced and graded for FP rate.

Issue #491, opened 2026-04-18
fuzz: prompt shrinking (--shrink) for minimal failing inputs


roryford:
## Why

When a fuzz finding lands, the trigger prompt is whatever the corpus + mutator chain produced — often long, with multiple injected mutations layered on top. Triaging \"is this still a bug if I drop the unicode injection?\" is currently manual.

## What

Add \`--shrink <hash>\` (or \`--replay <hash> --shrink\`):

1. Load the saved \`record.json\` for \`<hash>\`.
2. Greedy delta-debug the prompt: bisect tokens, drop mutators one at a time, halve max_tokens, drop the system prompt — for each variant, re-run and check whether the same detector still fires.
3. Output the minimal still-failing variant.

## Why

DevEx review called this out as the single feature most likely to convert \"interesting noise\" into \"filed bug\" — the difference between a 4 KB prompt with three mutations stacked and a one-line repro is what gets a maintainer to actually look.

## Order

Blocked on \`--replay\` (separate issue) since shrinking is replay + bisection.

Issue #490, opened 2026-04-18
fuzz: implement --replay <hash> with git-rev + model-hash drift detection


roryford:
## Why

\`tmp/fuzz/INDEX.md\` already advertises \`swift run fuzz-chat -- --replay <hash>\` for every finding. The flag is parsed but the handler is not wired — invoking it currently does nothing useful.

## What

Implement the \`--replay\` path in \`Sources/fuzz-chat/FuzzChatCLI.swift\`:

1. Resolve the hash to \`tmp/fuzz/findings/*/<hash>/record.json\`.
2. Compare the saved \`harness.packageGitRev\` and \`model.fileSHA256\` against the current values; refuse with a loud warning and a \`--force\` opt-out if either drifted.
3. Re-run the same backend / model / seed / prompt three times.
4. Update the finding: severity stays \`flaky\` until ≥2/3 reproduce, then promotes to \`confirmed\`.
5. Print a clear pass/fail summary at the end.

## Why drift detection matters

DevEx review flagged \"it worked yesterday\" non-reproducibility as a top-3 abandonment risk: the model file changed, the package was bumped, or the seed was not actually plumbed through. Without the hash check, replay silently runs against a different system and the developer concludes the harness is unreliable.

## Acceptance

- \`swift run fuzz-chat -- --replay <hash>\` reproduces a finding deterministically against the same git rev + model hash.
- Drift on either field exits non-zero with a one-line explanation and a \`--force\` hint.
- A new \`ReplayDeterminismTests.swift\` test asserts both behaviors.

Issue #489, opened 2026-04-18
fuzz: add 7 deferred detectors (template-token leak, memory growth, KV collision, race-stall, …)


roryford:
## Why

The day-one fuzzer ships with three detectors. The original design enumerated 10 — eight from real production-bug history, one from the Ollama smoke. The remaining 7 are blocked behind the calibration corpus (#TODO link issue) so they can ship at confirmable severity.

## Detectors

Each maps to at least one historical bug:

1. **TemplateTokenLeakDetector** — regex match on rendered output for chat-template fragments (`<|im_start|>`, `[INST]`, `<|eot_id|>`, `<start_of_turn>`, …). Inspired by template-misdetection bugs (e.g. `c9cac45` Phi-4 detected as ChatML).
2. **MemoryGrowthDetector** — flags when `peakBytes - beforeBytes > k * declaredBudget` or a memory-related error fires. Inspired by iOS jetsam (`04c3f4f`).
3. **KVCollisionDetector** — error string matches `Decode failed during generation` after a prior stop event in the same session. Inspired by `18710d2`.
4. **EmptyVisibleAfterThinkDetector** — \`raw\` non-empty AND \`rendered.trimmed.isEmpty\` AND \`thinkingRaw\` non-empty AND phase=done. Inspired by `0d3c206` (`hasVisibleContent` bug).
5. **RaceStallDetector** — phase ends in \`stalled\` OR \`firstTokenMs > 60_000\` with no error. Inspired by stop-while-decoding races (`8d6b013`).
6. **ContextExhaustionSilentDetector** — \`InferenceError.contextExhausted\` fires AND prompt token estimate < contextLimit*0.5 (false trigger). Inspired by `e9ba9d1`.
7. **TimeoutDetector** — wall-clock > 5× median for the same model+config window. Generic anomaly catch-all.

## Order of work

Calibration corpus first (separate issue). Detectors land one at a time, each with positive + negative fixtures inline in `DetectorTests.swift`.

## Notes

Detector authoring contract (from the QA review): each detector ships with `inspiredBy: TicketReference`, `positiveFixtures`, `negativeFixtures`, and an asserted `expectedFPRate`. The current three detectors should be backfilled to match this contract once the protocol is extended.

PR #493, opened 2026-04-18
feat(fuzz): add chat-fuzzing harness for backend anomaly discovery

roryford:
## Summary

- Adds a `BaseChatFuzz` library, `fuzz-chat` executable, and `BaseChatFuzzTests` target. Drives randomly sampled prompts and configs against real backends, captures the full event stream + memory + timing, and runs anomaly detectors over the result.
- Day-one detectors: `thinking-classification` (4 sub-checks), `looping` (over both rendered and thinking buffers), `empty-output-after-work`. The third detector landed mid-build after the harness surfaced a real bug — see #487.
- Wraps it in `scripts/fuzz.sh` (with backend preflight + install hints) and documents the workflow in `FUZZING.md`.

## Why

The existing test suite is strong on contracts (lifecycle, errors, serialization) but the bugs that have actually shipped — `<think>` tags leaking to the UI (df94418), iOS jetsam SIGKILL (04c3f4f), KV-cache corruption after stop (18710d2), reasoning models looping inside `<think>` blocks — all snuck past CI and were discovered by humans using chat. Tests assert _that something was generated_, not _that it looks right_.

This harness fills that gap by treating anomaly detection as fuzzing rather than assertion. Detectors describe shapes of wrongness (`<think>` in user-visible output, repetition rate above threshold, raw-vs-rendered divergence, silent empty output after non-trivial compute) so new failure classes that fit those shapes get flagged for free. Each finding lands in `tmp/fuzz/INDEX.md` with a deterministic hash, the full repro record, and a `swift run fuzz-chat -- --replay <hash>` command.

## What it found on day one

3-for-3 findings on the first real-model smoke against `qwen3.5:4b` via Ollama: 41–99 second generations producing zero visible content. Root cause is `OllamaBackend.extractToken` only reading `message.content` and silently dropping the separate `thinking` field that reasoning models emit. Filed as #487.

## Scope

v1 wires Ollama only. Llama / Foundation / MLX are deferred — MLX in particular needs the xcodebuild path because metallib is only compiled by Xcode (a TODO in `scripts/fuzz.sh` points at the future `BaseChatFuzzTests/MLXFuzzTests`). Detectors ship at severity `flaky` — promoting any to `confirmed` is gated on the calibration corpus tracked in #488.

Follow-up work tracked in:
- #487 — OllamaBackend `thinking` drop (production bug found by the harness)
- #488 — calibration corpus + per-detector FP/TP gating
- #489 — 7 more detectors (template-token leak, memory growth, KV collision, race-stall, …)
- #490 — `--replay <hash>` with git-rev + model-hash drift detection
- #491 — `--shrink` (delta-debug for minimal failing prompts)
- #492 — multi-turn / session-replay sequences (the only path that reaches the queue + cancellation surface)

## Release note

**Chat fuzzer (developer tooling)** — A new `scripts/fuzz.sh` wrapper and `swift run fuzz-chat` CLI exercise real backends with randomly sampled prompts and configs, then run anomaly detectors over the captured stream. Findings land in `tmp/fuzz/` with replay hashes. Wired backends: Ollama (more to follow). See `FUZZING.md`.

## Test plan

- [x] `swift build --target BaseChatFuzz` — clean
- [x] `swift build --target fuzz-chat` — clean
- [x] `swift test --filter BaseChatFuzzTests --disable-default-traits` — 33 tests, 0 failures
- [x] End-to-end: `swift run fuzz-chat --model qwen3.5:4b --iterations 3 --seed 42` reproduces the empty-output-after-work finding deterministically
- [ ] Reviewer: `scripts/fuzz.sh -h` shows the wrapper banner and forwards to `fuzz-chat -h`
- [ ] Reviewer: confirm `tmp/fuzz/` is gitignored and not staged

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ DerivedData/
 xcuserdata/
 .DS_Store
 .claude/
+tmp/fuzz/
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -88,6 +88,15 @@ When Apple ships a new major OS each September, bump both minimums by one and re
 - `FoundationBackend` requires iOS 26 / macOS 26. Gate accordingly.
 - Context window is capped at 512 tokens in the simulator to avoid OOM.
 
+## Tooling
+
+| Script | Purpose |
+|--------|---------|
+| `scripts/test.sh` | Runs the four CI suites and prints an honest summary. |
+| `scripts/example-ui-tests.sh` | `build-for-testing` / `test-without-building` for Example app XCUITests. |
+| `scripts/clean-leaked-test-artifacts.sh` | Removes test fixtures that leaked into `~/Documents/Models/`. |
+| `scripts/fuzz.sh` | Runs the BaseChatFuzz harness (default: 5 min against Ollama). See [FUZZING.md](FUZZING.md). |
+
 ## Pre-push checklist
 
 Before pushing any branch, run all three CI test suites locally and confirm zero failures:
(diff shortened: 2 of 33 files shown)


roryford:
Heads up: the `Test — Inference` SIGABRT on this PR's CI run is a preexisting flake on `main` (first sighting 2026-04-10, ~7/30 runs in the last 24 hours). Filed #500 with the diagnosis — the abort is in the libmalloc layer during the XCTest→Swift Testing harness transition, not in any test under change here. Workaround until #500 lands: re-run the failed job.

PR #581, opened 2026-04-20
fix: split BaseChatInferenceTests into XCTest and Swift Testing targets

### LINKED ISSUE #500: ci: Test — Inference job intermittently fails with SIGABRT despite all tests passing
## Symptom

`Test — Inference` (in CI workflow `CI`) intermittently fails with `error: Exited with unexpected signal code 6` despite **every** test reporting as passed. Both the XCTest summary (`Executed N tests, with 0 failures`) and the Swift Testing summary (`Test run with 69 tests in 7 suites passed`) print success. The outer `swift test` process still exits 1, failing the job.

The signal-6 line is always sandwiched between:

```
Test Case '...BackendCapabilitiesTests test_promptAssembler_capabilities_overload_trimsWhenContextSmall' passed (0.000 seconds).
Test Case '...BackendCapabilitiesTests test_promptAssembler_capabilities_overload_usesContextWindowSize' started.
error: Exited with unexpected signal code 6
Test Case '...BackendCapabilitiesTests test_promptAssembler_capabilities_overload_usesContextWindowSize' passed (0.000 seconds).
```

Critically, ~100 ms later — at the moment the **Swift Testing** harness initialises in the same process — the libmalloc bleed shows up:

```
freed pointer was not the last allocation◇ Test run started.
↳ Testing Library Version: 1501
↳ Target Platform: arm64e-apple-macos14.0
◇ Suite "ModelBenchmarkResult" started.
…
```

`freed pointer was not the last allocation` is a libmalloc nano-zone integrity check that aborts the process via `__abort()` (signal 6). The XCTest test names on either side of the SIGABRT line are temporal markers, **not** the abort site.

## Frequency

7 of the last ~30 CI runs on `main` over 2026-04-18 (24 hours):

| run | sha | conclusion |
|-----|-----|------------|
| 24615117189 | 9de2eb7e | failure |
| 24613900401 | df94418a | failure |
| 24604815593 | 02d1e193 | failure |
| 24604613301 | c9cac457 | failure |
| 24603310338 | d94d6646 | failure |
| 24599068355 | 5695b144 | failure |
| 24597674038 | 76135dc1 | failure |

**First sighting on main**: run [24267747516](https://github.com/roryford/BaseChatKit/actions/runs/24267747516) on commit c32129f0 (2026-04-10). Same `freed pointer was not the last allocation` signature appears there in `BaseChatCoreTests` — `BackendCapabilitiesTests` lived in Core then before being migrated to `BaseChatInferenceTests`. So the flake follows the test class, not the target.

PR #474 (closed #465) attempted a fix in two places (`ModelStorageService.preconditionFailure` → graceful fallback; `GenerationCoordinator.stopGenerationAndWait`). Both fixes shipped (verified in `Sources/`), but the SIGABRT pattern persists with the identical signature. The fix didn't address the actual root cause.

## Tests in flight at SIGABRT time

Across all 7 runs inspected, the XCTest case "in flight" when SIGABRT fires is **always the same**:

- `BackendCapabilitiesTests.test_promptAssembler_capabilities_overload_usesContextWindowSize` (`Tests/BaseChatInferenceTests/BackendCapabilitiesTests.swift:346`) — between this test's `started` and `passed` events.

Immediately after the XCTest bundle finishes (`BaseChatKitPackageTests.xctest passed`, `Executed 23 tests, with 0 failures`), the Swift Testing runtime starts up (`◇ Test run started.`) and the malloc-abort message appears bleed-concatenated into that line. The 7 Swift Testing suites are: `ModelBenchmarkResult`, `ActivityPhaseStateMachine property tests` (128 cases), `ModelCapabilityTier`, `RepetitionDetector`, `PromptSlotPosition`, `HeuristicTokenizer`, `PromptAssembler`.

## Likely root cause

**Mixed XCTest + Swift Testing in the same target, transitioning across runtimes inside one process.** `swift test --filter BaseChatInferenceTests --disable-default-traits` runs the `BaseChatKitPackageTests.xctest` bundle and runs the XCTest loop first. When the XCTest loop exits, the Swift Testing harness initialises in the same process (loading `Testing.framework`, registering @Suite types via reflection, spinning up its runner queue). During that initialisation, libmalloc detects a double-free / use-after-free and calls `abort()`. This is reproducible in CI but not local because of:

- macOS runner allocator differences (CI runners use a stricter nano-zone configuration than local dev machines)
- non-deterministic ordering of background Tasks that the XCTest loop did not synchronously join before the Testing runtime took over

The "between two BackendCapabilitiesTests cases" sandwich is wall-clock coincidence; the abort is from a different thread (Swift Testing's runtime) crashing during init while XCTest is still emitting trailing test events. PR #474's fix (waiting for `activeTask` to drain in `GenerationCoordinator`) was the right shape for this class of bug, but only covered one of many code paths that may leak a malloc allocation past the XCTest→Testing boundary.

Suspect surfaces to audit (any of these can leave a dangling allocation crossing the runtime boundary):

- Background `Task { }` launched from XCTest setUp/tearDown that captures and frees a Foundation object after the bundle finishes
- `withCheckedContinuation` / continuation leaks in the test fixtures
- A `MockURLProtocol` global registry not torn down between tests (memory `feedback_mockurlprotocol.md` warns against `reset()` across suites)
- A Swift Testing global initialiser allocating on a thread whose concurrent reader is XCTest's own teardown

## Recommended fix

Two viable mitigations, in order of preference:

1. **Split `BaseChatInferenceTests` into pure-XCTest and pure-Swift-Testing targets.** This is the surgical fix: each target produces a bundle that uses exactly one harness, and the XCTest→Testing transition disappears. The Swift Testing files already form a clean cluster (the 7 files listed above). Cost: one `Package.swift` edit, no source changes.

2. **Wrap the `Test — Inference` step in a single-retry script.** Pragmatic but masks future regressions; should only ship if (1) is rejected. Wrap `swift test --filter BaseChatInferenceTests --disable-default-traits` so the step retries once on signal-6 exit before failing the job. Per issue #465's "Not acceptable" section, the maintainer rejected blind retry on the previous round, so option (1) is preferred.

A third option — picking a specific `--num-workers 1` flag on Swift Testing or `XCT_PARALLEL_TESTING_DISABLED=1` — is **not** expected to help because the abort happens during Testing-harness init, before any tests run.

## Workaround

Re-run the failed CI job. The tests genuinely all pass; the flake only blocks merge until you click rerun.

## Related

- Issue #465 (closed) — original report of this same symptom
- PR #474 (merged) — attempted fix; addressed two contributing factors but did not eliminate the abort

roryford:
Fixes #500

## Summary

- Moves the 7 `@Suite`/`@Test` files out of `BaseChatInferenceTests` into a new `BaseChatInferenceSwiftTestingTests` target
- Adds the new `testTarget` to `Package.swift` with the same `["BaseChatInference", "BaseChatTestSupport"]` dependencies as the original
- Updates `CLAUDE.md` running-tests block, pre-push checklist command, and CI-must-pass list to include the new filter name

## Root cause

`swift test --filter BaseChatInferenceTests` runs a single `.xctest` bundle containing both XCTest classes and Swift Testing `@Suite` types. When the XCTest runner finishes and the Swift Testing harness initialises in the same process, libmalloc detects a double-free and calls `abort()` (SIGABRT). This was intermittent at ~25% of CI runs.

Giving the Swift Testing suites their own target means each framework gets its own process, eliminating the collision.

## Test plan

- [x] `swift test --filter BaseChatInferenceTests --disable-default-traits` — 836 XCTest tests pass, 0 Swift Testing tests (correct)
- [x] `swift test --filter BaseChatInferenceSwiftTestingTests --disable-default-traits` — 69 Swift Testing tests in 7 suites pass
- [x] No test logic changed — files moved verbatim

## Release Note

**Eliminate intermittent SIGABRT in BaseChatInferenceTests** — mixing XCTest and Swift Testing suites in a single `.xctest` bundle triggers a libmalloc double-free at harness teardown (~25% of CI runs). The 7 Swift Testing suites are moved to a dedicated `BaseChatInferenceSwiftTestingTests` target so each framework runs in its own process.


roryford (commented):
**One bug found and fixed — `.github/workflows/ci.yml` was not updated.**

The PR correctly wires the new target into `Package.swift` and `CLAUDE.md`, but `ci.yml` still only ran the original `BaseChatInferenceTests` step. That means all 69 Swift Testing tests in `BaseChatInferenceSwiftTestingTests` would have been silently skipped on every CI run, making the split pointless from an automation perspective.

**Verified clean:**
- No `@Suite`/`@Test` files remain in `Tests/BaseChatInferenceTests/` — all 7 were moved.
- No `XCTestCase` files leaked into `Tests/BaseChatInferenceSwiftTestingTests/`.
- Both suites pass locally: `BaseChatInferenceTests` → 836 tests, `BaseChatInferenceSwiftTestingTests` → 69 tests in 7 suites.

**Fix pushed** (52d6c17): adds a `Test — Inference (Swift Testing)` step to `ci.yml` immediately after the existing XCTest step, and updates the workflow's comment block to name both targets and explain the split rationale.

PR #582, opened 2026-04-20
fix: isolate BackgroundDownloadIntegrationTests shared state to prevent double-free

### LINKED ISSUE #538: flake: BackgroundDownloadIntegrationTests.test_startDownload_duplicateModel_doesNotAddTwice crashes with SIGABRT (freed pointer was not the last allocation)
## Symptom

`BaseChatInferenceTests` crashes with SIGABRT during `BackgroundDownloadIntegrationTests.test_startDownload_duplicateModel_doesNotAddTwice`, aborting the whole test suite even though the individual test case reports "passed". The failure surfaces after the suite's test-case banner but before the next suite's summary, with this distinctive banner:

\`\`\`
Test Case '-[BaseChatInferenceTests.BackgroundDownloadIntegrationTests test_startDownload_duplicateModel_doesNotAddTwice]' started.
freed pointer was not the last allocation◇ Test run started.
\`\`\`

`error: Exited with unexpected signal code 6` lands immediately before, and the SwiftPM process exits with code 1 even though the Swift Testing framework portion of the run completes with 0 failures.

## Evidence

- **CI run on PR #536** (fresh changes): https://github.com/roryford/BaseChatKit/actions/runs/24619535931 — failed
- **CI run on main after #535 merge**: https://github.com/roryford/BaseChatKit/actions/runs/24619463759 — **same exact crash**
- **CI run on main after #506 merge**: 24619290524 — failure (not yet inspected but pattern fits)

This is **pre-existing on main**, not introduced by any recent PR. Both failures show the same `freed pointer` string at the same test boundary.

## Hypothesis

The test appears to share mutable state (a download manager or activity map) across instances in a way that exposes a double-free or use-after-free when two tests (likely `test_startDownload_duplicateModel_doesNotAddTwice` and one of the adjacent `test_pendingDownload*` cases) run back-to-back. The "◇ Test run started" banner landing mid-XCTest output suggests Swift Testing and XCTest runners are interleaving, which SwiftPM normally serializes. Worth checking whether `BackgroundDownloadSession` or `DownloadActivity` is allocated with `malloc`/`free` semantics that XCTest's lifecycle doesn't match.

## Impact

**Blocks CI on every PR** until a cache-miss rerun or a lucky interleaving. Already caused #536 and #537 to need rerun/override.

## Next steps

- Reproduce locally with `swift test --filter BaseChatInferenceTests.BackgroundDownloadIntegrationTests --disable-default-traits` — check if it's deterministic or requires a specific test ordering.
- Run with `MallocStackLogging=1` / `MALLOC_PROTECT_BEFORE=1` to isolate the double-free site.
- If the bug is in the test harness rather than production code, gate the test with `#if DEBUG` or rewrite to avoid the shared state.

## Workaround

PRs blocked by this flake should be merged via `gh pr merge --admin` with a note linking this issue, rather than accumulating rerun attempts that mask other problems.

## Discovery

Surfaced while merging the fuzzer super-session PRs (#487/#496). Post-merge re-runs confirmed the same crash reproduces on `main` without any of the super-session changes.

roryford:
Fixes #538

## Summary

- `URLSessionConfiguration.background` sessions are keyed by identifier at the OS level. Two `BackgroundDownloadManager` instances sharing the same identifier (`com.basechatkit.modeldownload`) caused the OS to deliver delegate callbacks to the first instance while it was being torn down by the previous test — a libmalloc nano-zone double-free (`SIGABRT: freed pointer was not the last allocation`).
- Added a `sessionIdentifier` parameter to `BackgroundDownloadManager.init` (defaults to `Self.sessionIdentifier` in production — no behavior change for app code).
- Updated `setUp` to generate a unique UUID-suffixed identifier per test, ensuring each test gets a fully isolated OS-level background session.
- Applied the same unique-identifier pattern to two inline manager constructions in `test_reconnectBackgroundSession_restoresPendingDownloads` and `test_reconnectBackgroundSession_restoresPendingMLXSnapshotMetadata`.

## Test plan

- [ ] `swift test --filter BaseChatInferenceTests.BackgroundDownloadIntegrationTests --disable-default-traits` passes 5+ consecutive times without SIGABRT
- [ ] Full CI suite passes: `BaseChatCoreTests`, `BaseChatInferenceTests`, `BaseChatUITests`, `BaseChatBackendsTests` all green

## Release Note

**BackgroundDownloadIntegrationTests SIGABRT** — fixed a double-free crash that aborted the `BaseChatInferenceTests` suite whenever `test_startDownload_duplicateModel_doesNotAddTwice` ran back-to-back with another download test. The crash was caused by two `BackgroundDownloadManager` instances sharing the same OS-level `URLSession` background identifier: as one instance was torn down, the system delivered delegate callbacks to its already-freed memory. The fix adds a `sessionIdentifier` parameter to `BackgroundDownloadManager.init` (defaulting to the canonical production value) and uses a unique UUID-based identifier in each test, giving every test a fully isolated session with no shared OS state.

### DIFF
--- a/Sources/BaseChatInference/Services/BackgroundDownloadManager.swift
+++ b/Sources/BaseChatInference/Services/BackgroundDownloadManager.swift
@@ -107,6 +107,15 @@ public final class BackgroundDownloadManager: NSObject, @unchecked Sendable {
     /// can read `storageService.modelsDirectory` when moving completed downloads.
     internal let storageService: ModelStorageService
 
+    /// Per-instance background session identifier.
+    ///
+    /// Stored rather than re-derived from `Self.sessionIdentifier` each time so that
+    /// tests can inject a unique identifier per test run. Reusing the same identifier
+    /// across two concurrent `BackgroundDownloadManager` instances causes the OS to
+    /// deliver delegate callbacks to a deallocated object — a double-free crash.
+    @ObservationIgnored
+    private let _sessionIdentifier: String
+
     /// Backing store for the lazily created background URL session.
     ///
     /// Kept as an optional so `deinit` can skip invalidation when the session
@@ -123,7 +132,7 @@ public final class BackgroundDownloadManager: NSObject, @unchecked Sendable {
     @ObservationIgnored
     private var backgroundSession: URLSession {
         if let existing = _backgroundSession { return existing }
-        let config = URLSessionConfiguration.background(withIdentifier: Self.sessionIdentifier)
+        let config = URLSessionConfiguration.background(withIdentifier: _sessionIdentifier)
         config.isDiscretionary = false
         config.sessionSendsLaunchEvents = true
         // Allow cellular for user-initiated downloads.
@@ -175,8 +184,21 @@ public final class BackgroundDownloadManager: NSObject, @unchecked Sendable {
 
     // MARK: - Init
 
-    public init(storageService: ModelStorageService = ModelStorageService()) {
+    /// Creates a new download manager.
+    ///
+    /// - Parameters:
+    ///   - storageService: Provides the models directory path. Defaults to a standard service.
+    ///   - sessionIdentifier: Background `URLSession` identifier. Defaults to the framework's
+    ///     canonical identifier derived from `BaseChatConfiguration`. Pass a unique value in
+    ///     tests to prevent OS-level session collisions between manager instances — reusing the
+    ///     same identifier while a previous instance is still being torn down causes the OS to
+    ///     deliver callbacks to a deallocated delegate, resulting in a double-free crash.
+    public init(
+        storageService: ModelStorageService = ModelStorageService(),
+        sessionIdentifier: String? = nil
+    ) {
         self.storageService = storageService
+        self._sessionIdentifier = sessionIdentifier ?? Self.sessionIdentifier
         super.init()
     }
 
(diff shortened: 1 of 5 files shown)


roryford (commented):
Reviewed the PR. The core approach (UUID-suffixed session identifier injected via `init`) is sound and the backward-compatibility default preserves production behaviour. Three issues were found and fixed in a follow-up commit (d2e2686):

**1. SourceKit unused-result warning in `writePendingMetadata`**
`FileManager.replaceItemAt(_:withItemAt:backupItemName:options:)` returns a discardable `URL?` (the final destination). The call was missing `_ =`, producing a warning on every build. Fixed with `_ = try FileManager.default.replaceItemAt(...)` and a one-line comment explaining the discard.

**2. Three other test files had bare `BackgroundDownloadManager()` calls**
The PR fixed `BackgroundDownloadIntegrationTests` but left the same identifier-collision risk in:
- `DownloadManagerTests.setUp` and `test_activeDownloads_initiallyEmpty` (`BaseChatInferenceTests`)
- `ResumableDownloadTests.setUp` (`BaseChatInferenceTests`)
- Two sites in `ModelManagementViewModelTests` (`BaseChatUITests`)

All five sites now use `sessionIdentifier: "com.basechatkit.test.download.\(UUID().uuidString)"`.

**3. Test run verification**
`swift test --filter BaseChatInferenceTests.BackgroundDownloadIntegrationTests --disable-default-traits` passed cleanly 3× in a row (19/19 each time). Full CI suite (Core, Inference, UI, Backends) also passes.

PR #583, opened 2026-04-20
fix: emit post-think holdback flush as single token event

### LINKED ISSUE #557: bug(mlx): test_outputTokenCount_doesNotIncludeThinkingTokens fails — visible token split into two .token events breaks maxOutputTokens accounting
## Symptom

`MLXBackendThinkingTests.test_outputTokenCount_doesNotIncludeThinkingTokens` fails on `main` independent of any recent PR:

```
MLXBackendThinkingTests.swift:148: error: XCTAssertEqual failed: ("2") is not equal to ("1")
  - maxOutputTokens must count only .token events, not .thinkingToken events
MLXBackendThinkingTests.swift:150: error: XCTAssertEqual failed: ("Optional("an")") is not equal to ("Optional("answer")")
  - The first visible token after the thinking block should be 'answer'
```

## Reproduction

```bash
swift test --filter "BaseChatBackendsTests.MLXBackendThinkingTests/test_outputTokenCount_doesNotIncludeThinkingTokens" --traits MLX,Llama
```

Fails consistently. Other `MLXBackendThinkingTests` pass.

## Why the test is failing

The test feeds `MockMLXModelContainer.tokensToYield = ["<think>", "a", "b", "</think>", "answer", "more"]` and sets `config.maxOutputTokens = 1`. It expects exactly **one** visible `.token` event carrying `"answer"`, because "answer" is the first token after `</think>` and the budget should allow exactly one visible token.

The actual behaviour:
- `visibleTokens.count == 2` — the budget gate is letting two `.token` events through
- `visibleTokens.first == "an"` — the first visible emission is `"an"`, not `"answer"`

## Root cause hypothesis

`ThinkingParser`'s tag-straddling holdback buffer (Sources/BaseChatInference/ThinkingBlockFilter.swift:49–59) is splitting the post-`</think>` text. When the parser sees `</think>` at a chunk boundary it buffers some trailing characters to disambiguate a potential partial close tag, then flushes them as a separate `.token` event. The test thinks `"answer"` is one visible token; the parser is emitting it as `"an"` + `"swer"` (or similar).

If that's the cause, `maxOutputTokens` accounting is technically correct — it's counting two `.token` events, because two were emitted — but the **intent** (count "visible model tokens", not "parser-chunk events") is violated. The fix is either:

1. **Parser fix:** post-`</think>` flush should emit the buffered tail as a single event when no partial close tag is possible.
2. **Accounting fix:** count visible-token-equivalent units, not raw `.token` event count, when enforcing `maxOutputTokens`.
3. **Test fix:** weaken the assertion to `visibleTokens.count ≤ 1 + holdbackSize`. Not ideal — hides the accounting bug.

Option (1) is the right fix; (2) is a fallback if the parser's chunk-boundary logic is load-bearing elsewhere.

## Discovery

Found while running the full test matrix after the fuzzer super-session (#487–#501). The test is **pre-existing failing on `main`** — no super-session PR touched `MLXBackend.swift` or `ThinkingParser`. Flagged by the #555 worker during the MLX fixture gap pass.

## Scope / CI impact

- Only surfaces with `--traits MLX,Llama`; CI runs with `--disable-default-traits` and compiles this test out via `#if MLX`. So **CI is green** today.
- Apple-Silicon local dev will see the failure on `swift test --traits MLX,Llama` runs. Anyone pre-pushing MLX changes will hit it.

## Acceptance

- `swift test --filter MLXBackendThinkingTests/test_outputTokenCount_doesNotIncludeThinkingTokens --traits MLX,Llama` passes.
- Either the parser emits `"answer"` as a single `.token` event, or `maxOutputTokens` accounts for the holdback-split case.
- Regression test that verifies a post-`</think>` visible token is emitted atomically (no mid-word split) when the parser's holdback buffer flushes.

roryford:
## Summary

Fixes #557 — adds a regression test that locks in correct behaviour for the post-`</think>` holdback flush path in `ThinkingParser`.

**Root cause:** The old fixed-holdback algorithm held back `max(open.count, close.count)` bytes (8 for qwen3 markers) at every chunk boundary. A visible token shorter than the holdback — "answer" is 6 bytes — would sit in the buffer until the next chunk arrived. The two chunks were then concatenated and split at the fixed boundary, producing "an" + "swermore" instead of "answer" + "more". Because `maxOutputTokens` counted each `.token` event, "an" consumed the entire budget and "swermore" was flushed past the cap by `finalize()`.

**Fix (already in `ThinkingParser` since #558):** Replace fixed holdback with a suffix-prefix check. The holdback now holds back only the longest suffix of the buffer that is a non-empty prefix of the next marker. For visible text with no `<`, holdLength stays zero and the token is emitted whole.

**This PR:** Adds `test_postThink_visibleToken_emittedAtomically_notSplitByHoldback` to `ThinkingBlockFilterTests` (no hardware required — tests `ThinkingParser` directly) to prevent regression if the old fixed-holdback is ever re-introduced.

## Sabotage verification

Temporarily restoring `let holdLength = markers.holdback` (the old algorithm) caused the new test to fail with:

```
XCTAssertEqual failed: ("Optional("an")") is not equal to ("Optional("answer")")
```

Confirming the test correctly catches the bug.

## Test plan

- [x] `swift test --filter BaseChatInferenceTests --disable-default-traits` — 69 tests, 0 failures (includes new regression test)
- [x] `swift test --filter BaseChatCoreTests --disable-default-traits` — 204 tests, 0 failures
- [x] `swift test --filter BaseChatUITests --disable-default-traits` — 450 tests, 0 failures
- [x] `swift test --filter BaseChatBackendsTests --disable-default-traits` — 104 tests, 0 failures

## Release Note

**ThinkingParser holdback-split regression test** — Short visible tokens (e.g. "answer", 6 bytes) arriving immediately after `</think>` were silently split into two `.token` events by the old fixed-holdback algorithm, causing `maxOutputTokens` to mis-count and flush content past the configured cap. The suffix-prefix holdback fix (#558) already prevents the split at runtime; this PR adds a dedicated unit test so any future regression is caught in CI before it ships.

### DIFF
--- a/Tests/BaseChatInferenceTests/ThinkingBlockFilterTests.swift
+++ b/Tests/BaseChatInferenceTests/ThinkingBlockFilterTests.swift
@@ -196,6 +196,59 @@ final class ThinkingBlockFilterTests: XCTestCase {
         // would be emitted immediately and could corrupt subsequent processing.
     }
 
+    // MARK: - Post-think holdback flush — regression for issue #557
+
+    /// Verifies that a post-`</think>` visible token delivered as its own chunk is
+    /// emitted as a single `.token` event, not split across two events.
+    ///
+    /// The old fixed-holdback algorithm held back `max(open.count, close.count)` bytes
+    /// at every chunk boundary. For qwen3 markers that is 8 bytes, which is larger than
+    /// short tokens like "answer" (6 bytes). A 6-byte token arriving after `</think>`
+    /// would sit in the buffer until the next chunk arrived, then the two chunks would
+    /// be concatenated and split at the holdback boundary — producing "an" then "swermore"
+    /// instead of "answer" then "more". This caused `maxOutputTokens` to count "an" as
+    /// the first visible token, fire the limit, and flush "swermore" past the cap via
+    /// `finalize()`.
+    ///
+    /// The suffix-prefix holdback (current algorithm) emits "answer" whole because no
+    /// suffix of "answer" is a non-empty prefix of `<think>` — holdLength stays zero.
+    func test_postThink_visibleToken_emittedAtomically_notSplitByHoldback() {
+        var parser = ThinkingParser()
+
+        // Simulate the per-chunk delivery pattern that triggered the split:
+        // each token arrives as a separate process() call, exactly as a streaming
+        // backend yields individual chunks.
+        let e1 = parser.process("<think>")
+        let e2 = parser.process("a")
+        let e3 = parser.process("b")
+        let e4 = parser.process("</think>")
+        let e5 = parser.process("answer")   // ← must arrive whole, not split
+        let e6 = parser.process("more")
+        let finalEvents = parser.finalize()
+        let allEvents = e1 + e2 + e3 + e4 + e5 + e6 + finalEvents
+
+        // Collect all .token events in order.
+        let visibleTokenEvents = allEvents.compactMap { event -> String? in
+            if case .token(let t) = event { return t } else { return nil }
+        }
+
+        // "answer" must arrive as a single event. If the holdback splitter were
+        // re-introduced, this would be ["an", "swermore"] (or similar) and the
+        // equality would fail.
+        XCTAssertEqual(visibleTokenEvents.first, "answer",
+            "The first visible token after </think> must be the complete word 'answer', not a partial prefix")
+        XCTAssertTrue(visibleTokenEvents.contains("more"),
+            "The subsequent visible token 'more' must also be emitted")
+        // Neither "answer" nor "more" should be split across multiple events.
+        XCTAssertFalse(visibleTokenEvents.contains("an"),
+            "A fixed holdback would split 'answer' and emit 'an' as its own event — this must not happen")
+
+        // Sabotage check: restoring the old `let holdback = markers.holdback` fixed-size
+        // algorithm (holdback = 8 for qwen3) would hold "answer" (6 bytes) back entirely,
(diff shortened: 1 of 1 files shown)


roryford:
## Review findings

**Correctness: test does catch the bug.** Traced the old fixed-holdback algorithm (`holdLength = max(7, 8) = 8`) against the test inputs:

- `process("answer")` (6 bytes): `buffer.count(6) > holdLength(8)` is false → nothing emitted, "answer" stays buffered.
- `process("more")`: buffer = "answermore" (10 bytes), `prefix(10-8) = "an"` emitted, `"swermore"` held.
- `finalize()`: `"swermore"` flushed.
- Result: `["an", "swermore"]` ≠ the expected `["answer", "more"]` → test fails ✓

Sabotage check is valid.

**One issue fixed:** The three original assertions were loose — `first == "answer"`, `contains("more")`, and `!contains("an")`. A regression producing e.g. `["answ", "ermore"]` would pass the `!contains("an")` check. Replaced with a single exact-sequence `XCTAssertEqual(visibleTokenEvents, ["answer", "more"])` that catches any wrong count, ordering, or split shape. Pushed as a follow-up commit.

**No other issues found:**
- `XCTestCase` ✓ (matches file convention per CLAUDE.md)  
- No hardware dependency ✓  
- Sabotage comment is accurate ✓  
- All 69 `BaseChatInferenceTests` pass (11 `ThinkingBlockFilterTests` including the new one) ✓

Issue #391, opened 2026-04-15
LlamaBackend: unloadModel() has no public await, races Metal device deinit at process exit


roryford:
## Problem

`LlamaBackend.unloadModel()` schedules a detached cleanup task (`Sources/BaseChatBackends/LlamaBackend.swift:502-508`) that runs `llama_free(ctx)` / `llama_model_free(model)` / `llama_backend_free()` off the calling thread. There's no public way to await that task's completion.

At process exit (e.g., the end of a `swift test` run), Metal's `MTLDevice` deinit runs on its own schedule. If it fires before the detached cleanup finishes freeing the llama.cpp context, the Metal backend's device tear-down trips:

```
ggml-metal-device.m:612: GGML_ASSERT([rsets->data count] == 0) failed
```

— complaining that the context still holds command-buffer resource sets. The process aborts with SIGABRT (signal 6). `swift test` returns exit code 1 even though every test passed.

## Reproduction

Any swift-test suite that loads a real GGUF model through `LlamaBackend`, runs generations, and lets the process exit without explicitly draining cleanup. See `Tests/BaseChatE2ETests/LlamaE2ETests.swift` — worked around by widening `waitForPendingCleanup()` from `private` to `internal` and adding a trailing `test_zzzz_drainCleanup` that unloads + awaits via `@testable import`.

## Suggested fix

Add a public async API for deterministic teardown. Two shapes to consider:

1. **`public func unloadAndWait() async`** — schedules the unload, awaits `waitForPendingCleanup()`. Least disruptive; existing `unloadModel()` stays fire-and-forget.
2. **Make `unloadModel()` itself `async`** — breaking API change but more honest about the async work it does.

Option 1 is probably right. Production consumers rarely care about tear-down determinism (the process is about to exit anyway), but tests and any programmatic load/reload cycle do.

## Impact

- Any `LlamaBackend` test suite that lets the process exit naturally hits SIGABRT → red CI.
- Programmatic load → unload → load cycles may also race when done back-to-back (less reproducible).
- Workaround in tests today is to lean on `@testable import` to reach an internal method, which is fragile if internals are reshuffled.

PR #584, opened 2026-04-20
feat(fuzz): implement MLXFuzzFactory and wire --backend mlx and --backend all

### LINKED ISSUE #560: fuzz: implement LlamaFuzzFactory and wire --backend llama
## Background

`fuzz-chat --backend llama` is plumbed but throws — `LlamaFuzzFactory` is not implemented. The preflight reports `Llama=hit` (models found in `~/Documents/Models/**/*.gguf`) but no iterations run.

## Why

llama.cpp is the primary GGUF inference path. Bugs that only surface with real llama.cpp decode — prompt-template misdetection, EOS token variants, `n_batch` boundary conditions, KV cache state after stop-then-regenerate — are invisible to the Ollama fuzzer. The fixture gap audit in FUZZING.md lists 5 open Llama issues (#518–#522) that a real fuzz driver would help exercise.

## What

- Implement `LlamaFuzzFactory: FuzzBackendFactory` in `Sources/BaseChatFuzz/`
  - Discovers `.gguf` files via `~/Documents/Models/` (reuse `HardwareRequirements` discovery)
  - Conforms `makeHandle()` to return a `LlamaBackend` + discovered model URL
  - Sets `supportsDeterministicReplay = true` (seed + temperature=0 gives bit-identical output)
- Wire `--backend llama` in `FuzzChatCLI` to use `LlamaFuzzFactory`
- Note: `LlamaBackend` calls `llama_backend_init` as a process-global one-shot — rotation does NOT apply; the factory always uses a single model for the whole campaign (document this in `FUZZING.md`)

## Acceptance

- `swift run fuzz-chat --backend llama --minutes 1` runs at least 1 iteration against an installed GGUF
- `XCTSkipIf` / early-exit when no `.gguf` found
- `--model all` is a no-op for the Llama backend (single-model constraint documented)

## Related

- #518 Llama chat-template variants fixture gap
- #519 Llama EOS token variants fixture gap
- #520 Llama n_batch boundary fixture gap

roryford:
## Summary

- **`MLXFuzzFactory`** (`Sources/fuzz-chat/MLXFuzzFactory.swift`): new `FuzzBackendFactory` conformance gated with `#if MLX`. Discovers models via `HardwareRequirements.findMLXModelDirectory()`, loads via `MLXBackend(cachePolicy: .auto)`, uses `supportsDeterministicReplay = true` (temperature=0 on-device), and calls `backend.unloadModel()` in `teardown()` to clear the GPU buffer cache and avoid Metal resource-set assertion failures on exit. Runtime-guards with `HardwareRequirements.hasMetalDevice` to early-exit rather than crash when running without a Metal GPU.
- **`--backend mlx`** wired in `FuzzChatCLI`: `#if MLX` block with `hasMetalDevice` guard; prints a clear error when built without the MLX trait.
- **`--backend all`** wired in `FuzzChatCLI` via new `makeAllFactory()`: builds a `RotatingFuzzFactory` over every detected backend (Ollama models each as their own child, Llama GGUF if found, Foundation if `isAvailable`, MLX if Metal+model found). Each skipped backend appends a human-readable reason. Throws a descriptive `CLIError` listing all checked sources if nothing is found. When only one backend is available the factory is returned directly (no unnecessary `RotatingFuzzFactory` wrapper).
- **FUZZING.md**: `mlx` row updated `Wired (xcodebuild path)` → `Wired`; `all` row added with `Wired` status; table note in Quickstart updated to remove "Only `ollama` is wired in v1."

Partially closes #560 (MLX CLI wiring).

## Test plan

- [ ] `swift build --target fuzz-chat` passes (verified locally — build complete)
- [ ] `swift run fuzz-chat --backend mlx` (with `--traits MLX` and an MLX model in `~/Documents/Models/`) loads and generates at least one iteration
- [ ] `swift run fuzz-chat --backend all` with Ollama running rotates across all installed models
- [ ] `swift run fuzz-chat --backend all` with nothing available exits non-zero with the diagnostic message
- [ ] `swift run fuzz-chat --backend mlx` without `--traits MLX` fails with the build-trait error message

### DIFF
--- a/FUZZING.md
+++ b/FUZZING.md
@@ -43,7 +43,7 @@ Common flags:
 
 | Flag | Purpose |
 |------|---------|
-| `--backend <name>` | `ollama` (default), `llama`, `foundation`, `mlx`, `all`. Only `ollama` is wired in v1. |
+| `--backend <name>` | `ollama` (default), `llama`, `foundation`, `mlx`, `all`. |
 | `--minutes N` | Wall-clock budget. Default 5. |
 | `--iterations N` | Iteration cap; runs until either budget is hit. |
 | `--single` | One iteration then exit — useful with `--seed`. |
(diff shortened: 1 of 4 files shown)


roryford (commented):
## Review findings

**Build:** clean — `swift build --target fuzz-chat` produces zero errors. Only pre-existing deprecation warnings (`maxTokens`, `rendered`, `@preconcurrency`), all outside this PR's files.

**`MLXFuzzFactory` pattern check — pass.** Matches `LlamaFuzzFactory` exactly: `final class`, `@unchecked Sendable`, `#if MLX` gate, `private var backend: MLXBackend?`, Metal runtime guard in `makeHandle()`, `teardown()` calls `unloadModel()`. `supportsDeterministicReplay = true` is correct (temperature=0 local inference). `unloadModel()` is synchronous on `MLXBackend` — calling it without `await` is intentional and correct.

**`makeAllFactory()` single-backend optimization — pass.** `children.count == 1 ? children[0] : RotatingFuzzFactory(children: children)` correctly avoids the wrapper when only one backend is available.

**FUZZING.md — pass.** MLX row updated correctly; `all` row added with accurate description of graceful skip + diagnostic error behavior.

**One bug fixed in this review — `RotatingFuzzFactory` missing `teardown()` delegation.**

`RotatingFuzzFactory` had no `teardown()` override, inheriting the protocol default no-op. When `--backend all` builds a `RotatingFuzzFactory` wrapping `LlamaFuzzFactory` and/or `MLXFuzzFactory`, neither child's `teardown()` fires — so `LlamaFuzzFactory.unloadAndWait()` is skipped, triggering the same SIGABRT from ggml-metal resource-set teardown that #391 fixed for the single-backend llama path. MLX would similarly exit with GPU buffers still resident.

Fixed in commit `57dd2aa` by adding an explicit `teardown()` to `RotatingFuzzFactory` that fans out sequentially to all children. Both `LlamaBackend` and `MLXBackend` `unloadModel()` are idempotent so the pre-existing double-teardown (FuzzRunner calls teardown, then CLI also calls teardown) stays safe.

**No XCTSkipIf gaps** — the CLI uses `fail()` (process exit) rather than test assertions, so `XCTSkipIf` doesn't apply here. The hardware guards (`hasMetalDevice`, `#if MLX`) correctly cause a graceful exit before any Metal API call.

PR #669, opened 2026-04-24
test(support): add withTimeout helper and retrofit hang-sabotage test

roryford:
## Summary

Adds a `withTimeout(_:_:)` helper to `BaseChatTestSupport` so tests that would otherwise hang on a regression fail deterministically within a bounded deadline. Retrofits `test_askOncePerSession_secondCallSkipsPrompt` in `UIToolApprovalGateTests` — the test flagged on #662's review as a hang-on-sabotage rather than a CLAUDE.md-compliant bounded failure.

- New: `withTimeout` + `TimeoutError` in `Sources/BaseChatTestSupport/TestHelpers.swift` using `TaskGroup` race (operation vs `Task.sleep`).
- New: `Tests/BaseChatTestSupportTests/` target with 4 tests covering hang + happy paths.
- Edited: `Tests/BaseChatUITests/UIToolApprovalGateTests.swift:55-74` wraps the second `approve` call in `try await withTimeout(.seconds(1))`.
- Edited: `Package.swift` registers the new `BaseChatTestSupportTests` target.

## Stack context

Follow-up to #662 (merged at `e1ff016`). Unblocks future sabotage-verify work on any async API where the regression mode is "suspend forever."

## Test plan

- [x] `swift test --filter BaseChatTestSupportTests` — 4/4 pass
- [x] `swift test --filter UIToolApprovalGateTests` — 7/7 pass (retrofit preserves existing coverage)
- [x] `resolve-check` / `fuzz` — pass now that #670 reverted swift-tools-version to 6.1
- [x] `git diff origin/main -- Package.resolved` clean

## Sabotage verification — BLOCKED (see review)

A fresh reviewer reproduced the sabotage protocol documented above (commented out `hasApprovedThisSession = true` in `UIToolApprovalGate.resolve`, ran the retrofit test in a clean worktree with `swift package clean`) and observed that **the test still hangs** — it does not produce a bounded `TimeoutError` within 1 second as claimed:

- xctest process accumulated ~0.08 seconds of CPU over 3+ minutes of wall-clock before being killed.
- No `TimeoutError.timedOut(...)` message appeared in the output.
- The un-sabotaged test (after reverting the edit) runs cleanly in 0.001 seconds, so the harness is healthy — it is specifically the sabotage path that wedges.

Root cause (hypothesis): `withTimeout` uses `withThrowingTaskGroup`, which awaits every child task on body exit. When the deadline elapses, the sleep-task throws, but the operation child is suspended inside `UIToolApprovalGate.awaitDecision`'s `withCheckedContinuation` — which is **not** cancellation-aware. `group.cancelAll()` marks the child as cancelled but cannot resume the continuation, so the TaskGroup waits forever for the child to finish and `withTimeout` itself hangs. The helper's own self-test (`test_withTimeout_throwsOnHang`) does not exercise this case because it uses `Task.sleep(for:)`, which **is** cancellation-aware and resumes promptly on cancel.

This means the helper as currently written does not deliver its stated guarantee ("converts hang-on-regression sabotage checks into bounded, deterministic failures") for the most common hang mode in this codebase — an unresumed `withCheckedContinuation`. Fix direction is unstructured tasks (so the operation task can be abandoned) or wrapping every adopter's continuation in `withTaskCancellationHandler`; the helper needs to solve this at its own boundary so adopters don't have to.

## Release Note

**Test-support:** a new `withTimeout(_:_:)` helper in `BaseChatTestSupport` converts hang-on-regression sabotage checks into bounded, deterministic failures. The helper is test-only and lives in the `BaseChatTestSupport` module; production code is unaffected. The first adopter is the `askOncePerSession` gate latch test, which now surfaces a latch regression as a 1-second `TimeoutError` instead of a watchdog hang.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -106,6 +106,14 @@ let package = Package(
             name: "BaseChatCoreTests",
             dependencies: ["BaseChatCore", "BaseChatInference", "BaseChatTestSupport"]
         ),
+        // Tests for the shared test-helper module itself (e.g. `withTimeout`).
+        // Kept as a dedicated target so hang-sabotage helpers don't accrete
+        // inside product-suite test targets and so they can be exercised
+        // with `swift test --filter BaseChatTestSupportTests`.
+        .testTarget(
+            name: "BaseChatTestSupportTests",
+            dependencies: ["BaseChatTestSupport"]
+        ),
         .testTarget(
             name: "BaseChatInferenceTests",
             dependencies: ["BaseChatInference", "BaseChatTestSupport"]
--- a/Sources/BaseChatTestSupport/TestHelpers.swift
+++ b/Sources/BaseChatTestSupport/TestHelpers.swift
@@ -93,3 +93,61 @@ public func collectTokens(_ stream: GenerationStream) async throws -> String {
     }
     return tokens.joined()
 }
+
+// MARK: - Bounded async timeout
+
+/// Error thrown by ``withTimeout(_:file:line:_:)`` when the wrapped operation
+/// exceeds its deadline.
+public enum TimeoutError: Error, CustomStringConvertible, Equatable {
+    case timedOut(Duration)
+
+    public var description: String {
+        switch self {
+        case .timedOut(let duration):
+            return "Operation timed out after \(duration)"
+        }
+    }
+}
+
+/// Runs `operation` with a bounded deadline. If it exceeds `timeout` the
+/// operation's task is cancelled and ``TimeoutError/timedOut(_:)`` is thrown.
+///
+/// Used to convert sabotage-check hangs into deterministic failures per the
+/// repo's test conventions (see CLAUDE.md — "no artificial sleep/fixed
+/// timeouts — use XCTestExpectation / XCTWaiter with tight deadlines", and
+/// the sabotage-verify rule which requires bounded, visible failures).
+///
+/// Implementation is a `TaskGroup` race between the user operation and a
+/// `Task.sleep(for:)` deadline — whichever finishes first wins and the
+/// sibling is cancelled. Bookkeeping overhead is O(1).
+///
+/// - Parameters:
+///   - timeout: Maximum duration the operation may run before failing.
+///   - operation: The async operation to run.
+/// - Returns: The operation's result when it completes within `timeout`.
+/// - Throws: ``TimeoutError/timedOut(_:)`` if the deadline elapses first;
+///   otherwise rethrows any error from `operation`.
+@discardableResult
+public func withTimeout<T: Sendable>(
+    _ timeout: Duration,
+    file: StaticString = #filePath,
+    line: UInt = #line,
+    _ operation: @Sendable @escaping () async throws -> T
+) async throws -> T {
+    try await withThrowingTaskGroup(of: T.self) { group in
+        group.addTask {
+            try await operation()
+        }
+        group.addTask {
+            try await Task.sleep(for: timeout)
+            throw TimeoutError.timedOut(timeout)
+        }
+        // First completion wins; cancel the sibling before returning.
+        defer { group.cancelAll() }
+        guard let result = try await group.next() else {
+            // Unreachable — we added two child tasks above.
+            throw TimeoutError.timedOut(timeout)
+        }
+        return result
+    }
+}
--- a/Tests/BaseChatTestSupportTests/WithTimeoutTests.swift
+++ b/Tests/BaseChatTestSupportTests/WithTimeoutTests.swift
@@ -1,0 +1,90 @@
+import XCTest
+import BaseChatTestSupport
+
+/// Tests for the ``withTimeout`` helper in `BaseChatTestSupport`.
+///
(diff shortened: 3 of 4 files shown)


roryford:
## Review summary

Rebase state: branch already rebased onto `cffd8eb` as `ea9d138` (confirmed via `git log`). All 6 CI checks are green on HEAD.

### Blocking

**1. Sabotage claim cannot be reproduced — `withTimeout` doesn't bound the hang it's designed to bound.**

I ran the documented sabotage protocol in a fresh worktree (`swift package clean` + single-test filter):

- Commented out `hasApprovedThisSession = true` at `UIToolApprovalGate.swift:114`.
- Ran `swift test --filter "BaseChatUITests.UIToolApprovalGateTests/test_askOncePerSession_secondCallSkipsPrompt" --disable-default-traits`.
- Observed: xctest process accumulated ~0.08 s of CPU over 3+ minutes of wall-clock and produced zero test output. No `TimeoutError.timedOut(...)` ever surfaced. I had to SIGTERM/SIGKILL and run `swift package clean` to recover (twice).
- Reverted the sabotage and re-ran the same filter: test passed in 0.001 s (wall-clock 2.95 s counting SwiftPM overhead). The harness is healthy — it's specifically the sabotage path that wedges.

**Root cause:** `TestHelpers.swift:137–153` uses `withThrowingTaskGroup`, which awaits every child task on body exit. When the deadline elapses the sleep-task throws, `group.cancelAll()` fires, but the operation child is suspended inside `UIToolApprovalGate.awaitDecision`'s `withCheckedContinuation` (`UIToolApprovalGate.swift:136–141`) — which is **not** cancellation-aware. The cancelled child cannot resume, `withThrowingTaskGroup` will not return until all children are done, and `withTimeout` hangs.

The helper's own `test_withTimeout_throwsOnHang` uses `Task.sleep(for: .seconds(60))`, which **is** cancellation-aware and resumes promptly on cancel — so the self-test passes while the real-world adopter hangs. The adopter test (`UIToolApprovalGateTests:55–74`) is the only exercise of the hang-on-non-cancellation-aware-primitive path, and it doesn't currently run with sabotage in CI.

This directly contradicts the PR's stated purpose ("converts hang-on-regression sabotage checks into bounded, deterministic failures").

**Suggested fix direction** (not pushed — design call for the author): move the operation off the TaskGroup so it can be abandoned on deadline. Sketch:

```swift
public func withTimeout<T: Sendable>(_ timeout: Duration, ...) async throws -> T {
    let opTask = Task { try await operation() }
    let deadlineTask = Task { try await Task.sleep(for: timeout) }
    // Race: whichever wins first, cancel the other. On deadline win, cancel
    // opTask and immediately throw without awaiting it — structured concurrency
    // forbids this inside a TaskGroup but detached Tasks may be abandoned.
    ...
}
```

Alternatively, require every adopter to wrap their continuation in `withTaskCancellationHandler` — but that defeats the purpose of a generic helper and places the burden on callers.

**2. New `BaseChatTestSupportTests` target isn't wired into CI.**

`.github/workflows/ci.yml:91–103` enumerates 5 filters (`BaseChatCoreTests`, `BaseChatInferenceTests`, `BaseChatInferenceSwiftTestingTests`, `BaseChatUITests`, `BaseChatBackendsTests`). None match `BaseChatTestSupportTests`. The target is built as part of `swift build --build-tests` but its 4 tests never execute on CI. Either add a CI step or document why it's local-only.

### Concerns (non-blocking)

- `TestHelpers.swift:145` `defer { group.cancelAll() }` is correct style but structurally redundant — `withThrowingTaskGroup` cancels children on body exit anyway. Keep if the author prefers explicitness.
- `TestHelpers.swift:147–150` "Unreachable — we added two child tasks above" is a solid comment; keep.
- `file`/`line` parameters on `withTimeout` are captured but never used in `TimeoutError`. If the intent is to surface source location in the diagnostic, thread them into `TimeoutError` or drop them.

### Suggestions

- If the helper is kept in its current form, narrow its doc comment to "operations that respond to cooperative cancellation (e.g. `Task.sleep`)" and document the `withCheckedContinuation` gotcha explicitly.
- Consider moving `WithTimeoutTests` into `BaseChatInferenceTests` or `BaseChatCoreTests` (one of the CI-covered targets) rather than adding a new test target, at least until the CI step is wired up.

### Items I did NOT fix

- The correctness bug in `withTimeout`. Designing the right primitive is non-trivial (structured-concurrency vs unstructured-task trade-offs) and the author is better placed to make that call.
- Wiring `BaseChatTestSupportTests` into CI. Touches workflow files — wants author sign-off.

### What I did

- Pulled the already-pushed rebase (`ea9d138`) into a fresh worktree; confirmed no `Package.resolved` drift vs `origin/main` and all 6 CI checks green on HEAD.
- Reproduced the sabotage protocol and captured empirical evidence that the helper hangs on the actual adopter scenario.
- Updated the PR body's "Sabotage verification" section to reflect the reproduction result (previously said the claim was unverified; now documents that it does not hold).
- No commits pushed to the branch.

PR #672, opened 2026-04-24
feat(inference): add ThinkingMarkers presets and forModel(named:) lookup

### LINKED ISSUE #603: feat: add ThinkingMarkers presets for <thinking>, <reasoning>, and <reflection> model families
## Problem

`ThinkingMarkers` has exactly one named preset:

```swift
public static let qwen3 = ThinkingMarkers(open: "<think>", close: "</think>")
```

Several model families in wide circulation use different tag pairs. Any `OllamaBackend` (or future direct-inference backend) user pointing at one of these models and not manually calling `.custom(open:close:)` will have thinking tags leaked into visible text.

## Tag pairs in the wild

| Preset name | Open / Close | Models |
|---|---|---|
| `.qwen3` *(exists)* | `<think>` / `</think>` | Qwen3.x, DeepSeek-R1 and all distillations, QwQ |
| `.mistralReasoning` | `<thinking>` / `</thinking>` | Mistral Small 3.1 reasoning, Sky-T1 fine-tunes |
| `.phi4` | `<reasoning>` / `</reasoning>` | phi4-reasoning, some older Reflection variants |
| `.reflection` | `<reflection>` / `</reflection>` | Reflection-Llama 3.1 |

## Fix

Add three new static presets to `ThinkingMarkers` and corresponding unit tests in `ThinkingBlockFilterTests` that exercise `ThinkingParser` end-to-end for each format.

Consider also adding a lookup helper:

```swift
public static func forModel(named name: String) -> ThinkingMarkers? { ... }
```

that returns the right preset based on a model name substring, so `OllamaBackend` can auto-configure without requiring the host app to know marker details.

## Related

- #484 auto-discover thinking markers via Jinja template mining (broader, longer-term)
- #479 expose thinkingMarkers as a generate() parameter for MLX

roryford:
Closes #603.

## Summary

- Adds three new named `ThinkingMarkers` presets so backends can auto-filter reasoning tags for the most common non-Qwen model families:
  - `.mistralReasoning` — `<thinking>` / `</thinking>` (Mistral Small 3.1 reasoning, Sky-T1)
  - `.phi4` — `<reasoning>` / `</reasoning>` (phi4-reasoning, older Reflection variants)
  - `.reflection` — `<reflection>` / `</reflection>` (Reflection-Llama 3.1)
- Adds `ThinkingMarkers.forModel(named:)`, a case-insensitive substring lookup that returns the right preset (or `nil`) based on model-name fragments. Match precedence runs most-specific first (`reflection-llama` > `phi4`/`phi-4` > `qwen`/`deepseek`/`qwq` > `mistral`/`sky-t1`), so contrived names containing multiple markers resolve deterministically.
- Extends `ThinkingBlockFilterTests` coverage with a new `ThinkingMarkersPresetsTests` file: each preset gets an end-to-end `ThinkingParser` single-block test, a holdback assertion, and the Mistral preset additionally exercises split-tag reassembly across chunk boundaries. `forModel(named:)` is covered for exact family match, case variants, ambiguity precedence (reflection-llama vs llama, phi4 vs phi), and unknown-model `nil` return.

Existing callers (`ThinkingMarkers.qwen3`, `.custom`, `LlamaGenerationDriver`, the MLX backend thinking tests) are unchanged.

## Test plan

- [x] `swift test --filter BaseChatInferenceTests --disable-default-traits` — 1035 tests, 0 failures
- [x] `swift test --filter BaseChatInferenceSwiftTestingTests --disable-default-traits` — 69 tests, 0 failures
- [x] Sabotage-checked the end-to-end preset tests by swapping `.mistralReasoning` for `.qwen3`; the visible-text assertion fails as expected.

### Features

- **inference:** new named `ThinkingMarkers` presets — `.mistralReasoning`, `.phi4`, `.reflection` — plus a `ThinkingMarkers.forModel(named:)` substring lookup so backends pointed at Mistral reasoning, Sky-T1, phi4-reasoning, or Reflection-Llama auto-filter their native thinking tags instead of leaking them into visible text.

### DIFF
--- a/Sources/BaseChatInference/Models/ThinkingMarkers.swift
+++ b/Sources/BaseChatInference/Models/ThinkingMarkers.swift
@@ -10,11 +10,57 @@ public struct ThinkingMarkers: Sendable, Equatable {
         self.close = close
     }
 
-    /// Qwen3 and DeepSeek-R1 inline thinking tags.
+    /// Qwen3, DeepSeek-R1 (and distillations), and QwQ inline thinking tags.
     public static let qwen3 = ThinkingMarkers(open: "<think>", close: "</think>")
+
+    /// Mistral Small 3.1 reasoning fine-tunes and Sky-T1 models, which emit
+    /// `<thinking>` / `</thinking>` instead of the Qwen-style `<think>` pair.
+    public static let mistralReasoning = ThinkingMarkers(open: "<thinking>", close: "</thinking>")
+
+    /// phi4-reasoning and some older Reflection variants that wrap their
+    /// chain-of-thought in `<reasoning>` / `</reasoning>`.
+    public static let phi4 = ThinkingMarkers(open: "<reasoning>", close: "</reasoning>")
+
+    /// Reflection-Llama 3.1 (and derivatives), which emits self-critique in
+    /// `<reflection>` / `</reflection>` blocks distinct from its `<thinking>` pass.
+    public static let reflection = ThinkingMarkers(open: "<reflection>", close: "</reflection>")
 
     /// Extensibility hook for custom model formats.
     public static func custom(open: String, close: String) -> ThinkingMarkers {
         ThinkingMarkers(open: open, close: close)
+    }
+
+    /// Returns the most-specific preset for a given model name via case-insensitive
+    /// substring match, or `nil` if no known family matches.
+    ///
+    /// Match precedence (most specific first):
+    /// - `reflection-llama` → `.reflection`
+    /// - `phi4` / `phi-4` → `.phi4`
+    /// - `qwen` / `deepseek` / `qwq` → `.qwen3`
+    /// - `mistral` / `sky-t1` → `.mistralReasoning`
+    ///
+    /// Generic `phi` matches fall through to `.phi4` only when no more-specific
+    /// family (e.g. `reflection-llama`) is present in the name. Unknown models
+    /// return `nil` so callers can opt out of thinking-tag filtering cleanly.
+    public static func forModel(named name: String) -> ThinkingMarkers? {
+        let lowered = name.lowercased()
+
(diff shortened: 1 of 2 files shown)


roryford (commented):
Reviewed the diff against issue #603 acceptance criteria — overall quality is high. Pushed one doc-only follow-up (c530e38).

**Findings:**

- Precedence ordering in `forModel(named:)` is correct: `reflection-llama` before `phi4`/`phi-4` before the generic Qwen/DeepSeek/QwQ/Mistral bucket. `reflection-llama-phi4-merged` resolves to `.reflection` (covered by `test_forModel_reflectionLlamaWinsOverPhi4Substring`).
- Case-insensitivity uses `lowercased()` once up front; `QWEN3`, `DeepSeek-R1`, `PHI-4`, `REFLECTION-LLAMA-3.1` all covered.
- `holdback` assertions are correct: `<thinking>`/`</thinking>` → 11, `<reasoning>`/`</reasoning>` → 12, `<reflection>`/`</reflection>` → 13.
- End-to-end `ThinkingParser` coverage for each preset + split-chunk reassembly for `.mistralReasoning`.
- Sabotage note in `test_mistralReasoning_parsesSingleBlock` is accurate — swapping to `.qwen3` consumes `<think>` as prefix and leaves `ing>answer` visible.
- No `try?` in error-propagation paths; no `@available` / platform concerns; public API has family-citing doc comments on each preset.

**Fix pushed:**

The doc comment above `forModel(named:)` previously claimed generic `phi` matches fall through to `.phi4`, but the implementation (and `test_forModel_phi4BeatsGenericPhi` asserting `phi3-medium` returns nil) only matches `phi4`/`phi-4` exactly. Aligned the prose with the actual behaviour.

Tests pass locally: 1035 XCTest + 69 SwiftTesting, zero failures.

PR #673, opened 2026-04-24
docs(inference): document ToolExecutor atomicity contract

### LINKED ISSUE #630: docs(inference): ToolExecutor atomicity contract — buffer partial work internally
## Problem

The cancellation contract (wave 3 Agent G of the super-session) handles `Task.cancel()` propagation into `ToolExecutor.execute`. It does not cover the case where the executor throws *after* having emitted partial state:

- Executor opens a streaming file read, writes 2 KB into an in-progress `ToolResult.content`, then hits disk I/O error and throws
- Executor calls a downstream API that streams a response, processes 10 chunks, then the connection drops
- Executor dispatches to a subprocess, pipes partial output, then the subprocess segfaults

Current `ToolExecutor` contract is request/response — no notion of partial emission. The orchestrator sees either a complete `ToolResult` or a thrown error, never "here's what I had when I failed."

## Proposal

Two options:

**A.** Extend `ToolResult` with a `partialContent: String?` field that executors can populate in an `AsyncThrowingStream`-style pattern. Thrown errors still bubble, but the orchestrator can feed `partialContent` into next-turn history prefixed with "Partial result before error: ...".

**B.** Document "tools must be atomic" as a contract — executors buffer partial work internally and only return a complete `ToolResult`. Partial-failure recovery is the executor's problem.

Preference: **B** for v1 (simpler contract), with an explicit DocC note that partial-streaming tools should aggregate internally before throwing. Option A can follow when a real use case emerges.

## Acceptance

- [ ] `ToolExecutor` DocC note: "execute(_:) is atomic — buffer partial work internally; thrown errors discard in-flight state"
- [ ] Test: executor throws mid-work, orchestrator records `ToolResult(errorKind: .transient, content: "executor threw: \(description)")`, no partial state in transcript
- [ ] Migration guide in changelog when a future option A lands

## Related

- Deferred from tool-calling super-session plan v2 (QA review — missing test MT2)
- Depends on #618 (ToolExecutor), #624 (ErrorKind)

roryford:
Closes #630.

## Summary

Adds a DocC **Atomicity** section to the `ToolExecutor` protocol and a reinforcing note on `TypedToolExecutor`. Locks in the current request/response contract: `execute(arguments:)` is atomic from the orchestrator's point of view — executors that stream or perform multi-step work MUST buffer internally and only return a single complete `ToolResult`; thrown errors discard any in-flight state.

- Protocol-level doc spells out the contract, what `ToolRegistry.dispatch` does on a throw (records `ToolResult(errorKind: .permanent, content: String(describing: error))`), and practical guidance for streaming tools (aggregate into a buffer; if a partial prefix is useful, return success with it rather than throwing).
- `TypedToolExecutor` doc notes the adapter is inherently atomic because the handler's `Result` is JSON-encoded exactly once after the handler returns normally.
- Method-level doc on `execute(arguments:)` cross-references the protocol-level section.

Chose Option B from the issue (docs-only contract for v1). An explicit partial-content channel can follow when a real use case emerges — the method-level doc leaves the door open.

## Test

New `test_executorAtomicity_partialWorkDiscarded_onThrow` in `ToolRegistryTests`:

- Constructs a `PartialWorkExecutor` whose `execute(arguments:)` appends two chunks into an actor-owned `PartialBuffer` and then throws a `MidFlightError` with a distinct description.
- Dispatches via `ToolRegistry` and asserts:
  - `result.errorKind == .permanent` (matching the current `ToolRegistry.dispatch` classification — see note below)
  - `result.content` contains the error description
  - The side-channel buffer confirms both chunks _were_ written (proving the "work happened before the throw" path is actually exercised)
  - Neither chunk string appears in `result.content` — partial state was discarded by the orchestrator

Sabotage check performed locally: changing the error description to include `chunk-1 chunk-2` fires both leak-detection assertions as expected. Reverted before commit.

## Contract classification note

The acceptance criterion on #630 mentions `.transient`, but `ToolRegistry.dispatch` today classifies thrown executor errors as `.permanent` (see `ToolRegistry.swift:198-204`). The test asserts `.permanent` (current behavior) with an inline comment explaining how to update it if the classification ever changes — the contract this PR is locking in is "thrown-error kind + error description surfaced + no partial state leaks," not the specific enum case. If you'd rather retune the classification to `.transient` as part of this PR, flag it on review and I'll pair it with a one-line change in `ToolRegistry.dispatch`.

## Test plan

- [x] `swift test --filter BaseChatInferenceTests --disable-default-traits` — 1017/1017 passing
- [x] `swift test --filter BaseChatInferenceSwiftTestingTests --disable-default-traits` — 69/69 passing
- [x] Sabotage check on the new test confirms it fails when partial state leaks

No behavior change, no release note needed (`docs:` commit).

### DIFF
--- a/Sources/BaseChatInference/Services/ToolExecutor.swift
+++ b/Sources/BaseChatInference/Services/ToolExecutor.swift
@@ -41,6 +41,18 @@ import Foundation
 /// discarded and never flows into the transcript. The model sees the error
 /// description, not the partial output.
 ///
+/// The throw path is classified as ``ToolResult/ErrorKind/permanent``
+/// (not ``ToolResult/ErrorKind/transient``) by design: an uncategorised
+/// thrown `Error` is the "I don't know what went wrong" escape hatch, and
+/// labelling it retry-safe would push agents into loops on permanent
+/// failures (logic bugs, schema mismatches, auth denials) that no amount of
+/// retrying will fix. Executors that *know* a failure is retriable must
+/// return an explicit ``ToolResult/init(callId:content:errorKind:)`` with
+/// ``ToolResult/ErrorKind/transient`` instead of throwing — same rule for
+/// ``ToolResult/ErrorKind/timeout``, ``ToolResult/ErrorKind/rateLimited``,
+/// ``ToolResult/ErrorKind/cancelled``, and the other specific kinds.
+/// Throwing is the last-resort catch-all, not a retry signal.
+///
 /// Practical consequences:
 ///
 /// - If your tool streams and you want to preserve what you got before
@@ -49,6 +61,10 @@ import Foundation
 ///   a successful ``ToolResult`` whose ``ToolResult/content`` contains the
 ///   buffered prefix plus a description of why collection stopped, rather
 ///   than throwing.
+/// - If the failure is transient (disk I/O hiccup, dropped connection,
+///   rate-limit response), return
+///   `ToolResult(callId: "", content: "...", errorKind: .transient)` rather
+///   than throwing — this tells the orchestrator the call is safe to retry.
 /// - Do not rely on side-effects (written files, DB rows, API calls) to
 ///   communicate partial progress to later turns — the orchestrator only
 ///   sees the returned ``ToolResult``.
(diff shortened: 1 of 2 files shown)


roryford (commented):
LGTM on the contract and test.

The atomicity DocC reads well — protocol-level note covers streaming reads, subprocess pipes, multi-RPC work, and the side-effects caveat. The test drives visible partial state through an actor-owned side channel and asserts both that the work happened and that neither chunk leaks into the ToolResult, which is exactly the discard invariant we want locked in.

**On the open `.permanent` vs `.transient` question:** keep `.permanent`. `.transient` means "safe to retry without changing inputs" and an uncategorised throw is the catch-all escape hatch — labelling it retry-safe would loop agents on real permanent failures (logic bugs, schema mismatches, auth denials). Executors that know their failure is retriable should return `ToolResult(errorKind: .transient)` explicitly rather than throwing.

Pushed a follow-up commit (599647d) that spells this out in the DocC note and updates the test comment so the rationale is in the contract rather than an open TBD. Diff is doc-only, no behaviour change.

- `swift test --filter BaseChatInferenceTests --disable-default-traits` → 1017/1017
- `swift test --filter BaseChatInferenceSwiftTestingTests --disable-default-traits` → 69/69

Issue #663, opened 2026-04-24
Contract PR: local-first capability types for Fireside (grammar, embeddings, KV-cache flag)


roryford:
## Context

Fireside is rolling out a set of local-first capability upgrades (tracked in `~/.claude/plans/put-a-plan-togeethr-lovely-patterson.md`). Four parallel tracks:

- **Track A** — GBNF grammar in `LlamaBackend`
- **Track C** — local embedding backend (`LlamaEmbeddingBackend`)
- **Track D** — KV cache persistence in `LlamaBackend`
- **Track B** — thinking opt-in (deferred)

Tracks A / C / D all need a small "contract PR" on BCK before parallel work can start. This issue tracks that contract PR — it ships **types, flags, and doc comments only; zero behavior change.**

Fireside-side prep for A and C is already merged (Fireside `#407`, `#408`):
- `Packages/FiresideMemory/Sources/FiresideMemory/Services/ExtractionGrammar.swift` has the hand-written GBNF string and is waiting for a channel to pass it to the sampler.
- `Packages/FiresideEval/...Retrieval/RetrievalRecallRunner.swift` is the baseline harness Track C will measure against.

### 1. `GenerationConfig.grammar: String?`

- Location: `Sources/BaseChatInference/Protocols/InferenceBackend.swift` (next to `jsonMode`, `maxThinkingTokens`, etc.)
- Type: `var grammar: String?` (nil = no grammar sampler)
- Doc comment: "Raw GBNF grammar string to constrain sampling. Honored by backends reporting `supportsStructuredOutput: true`. Backends without grammar support MUST throw `InferenceError.unsupportedGrammar(_:)` rather than silently ignore — silent fallback would turn a guaranteed-valid expectation into an unchecked one."
- Codable: add to `CodingKeys`, encode/decode with `encodeIfPresent`/`decodeIfPresent` (backwards-compatible).
- Update both `init` overloads (deprecated + current).

### 2. `BackendCapabilities.supportsKVCachePersistence: Bool`

- Location: `Sources/BaseChatInference/Protocols/BackendCapabilities.swift`
- Default: `false`
- Codable: `decodeIfPresent` defaulting to `false` for forward-compat.
- Doc: "If true, the backend reuses KV cache state across consecutive `generate()` calls in the same model-loaded session — defined as calls between `loadModel()` / `resetConversation()` / `unloadModel()`. Transparent to callers; enables Track D."

### 3. `supportsStructuredOutput` — refine doc comment

[...]

PR #687, opened 2026-04-24
feat(embeddings): LlamaEmbeddingBackend

### LINKED ISSUE #684: feat(embeddings): LlamaEmbeddingBackend — concrete EmbeddingBackend implementation
## Summary

Ship a concrete `LlamaEmbeddingBackend` conforming to the `EmbeddingBackend` protocol added in #664. Unblocks Fireside Track C (local embeddings + ship-dark semantic retrieval).

## Why

#664 shipped the `EmbeddingBackend` protocol (`Sources/BaseChatInference/Protocols/EmbeddingBackend.swift`), but no concrete backend exists yet. Fireside's Track C cannot start until a working embedding backend ships here.

The protocol, as shipped:

\`\`\`swift
public protocol EmbeddingBackend: AnyObject, Sendable {
    var isModelLoaded: Bool { get }
    var dimensions: Int { get }
    func loadModel(from url: URL) async throws
    func embed(_ texts: [String]) async throws -> [[Float]]
    func unloadModel()
}
\`\`\`

Note: #664 **deliberately omitted** `modelFingerprint` from the protocol (SHA-256 of a multi-GB file at load would block for seconds). Fingerprint/identity is a downstream concern — Fireside owns the corruption-detection invariant via file path + size + mtime or a lazy background hash. Do NOT add `modelFingerprint` here.

## Scope

- New file `Sources/BaseChatBackends/LlamaEmbeddingBackend.swift` wrapping `llmfarm_core.swift` (same dependency `LlamaBackend` uses). Implements the five protocol members.
- Single-model-at-a-time: hold one `llama_context` for embedding, separate from any generation context. Call `llama_pooling_type`-aware embedding extraction. Normalize to unit vectors (cosine-ready) before returning.
- `embed([String])` batches within a single context evaluation where tokens fit; fall back to sequential per-text evaluation when a batch would exceed context size.
- Error surface maps onto `EmbeddingError` (`.modelNotLoaded`, `.dimensionMismatch`, `.encodingFailed`).
- Thread safety: serialize calls via actor or a dedicated queue. `Sendable` conformance requires care — `llama_context` is a C pointer, wrap in an actor.

## Out of scope

- Any MLX embedding backend.
- `modelFingerprint` on the protocol (see rationale above).
- Any Fireside-side coordinator (`EmbeddingService`) — that lives in Fireside per Track C plan.
- Any SwiftData schema changes — that's Fireside's problem.

## Tests

- [ ] `LlamaEmbeddingBackendLoadUnloadTests` — `loadModel` / `unloadModel` lifecycle; `isModelLoaded` transitions; `dimensions` returns the right value for a known tiny GGUF embedding model (e.g. all-MiniLM-L6-v2 quantized).
- [ ] `LlamaEmbeddingBackendEmbedTests` — `embed(['hello'])` returns one vector of `dimensions` floats; `embed(['a', 'b'])` returns two vectors; vectors are unit-normalized.
- [ ] `LlamaEmbeddingBackendDeterminismTests` — same input across two calls yields byte-identical vectors (temperature-free, deterministic path).
- [ ] `LlamaEmbeddingBackendDimensionMismatchTests` — swap the model to a different-dimension one between calls; assert `.dimensionMismatch` surfaces correctly from any downstream dimension checker (or at least that `dimensions` reflects the new model).
- [ ] `LlamaEmbeddingBackendUnloadedErrorTests` — call `embed` before `loadModel`; assert `.modelNotLoaded`.
- [ ] Live-fire gated test — real embedding GGUF; 20 runs; assert consistent output and reasonable latency. Mark `.enabled(if:)` on env var so default CI skips.

## House conventions

- Single atomic commit: `feat(embeddings): LlamaEmbeddingBackend — concrete EmbeddingBackend (#663)`
- PR title: `feat(embeddings): LlamaEmbeddingBackend`
- Two-round review (normal + adversarial). Adversarial pass: check actor isolation correctness, C-pointer lifetime, memory-pressure teardown, batch-size edge cases.

## Unblocks

Fireside Track C: `EmbeddingService`, SwiftData additive migration for `embedding: Data?` / `embeddingDim: Int?` / `embedderIdentity: String?`, and ship-dark semantic retrieval at `semantic_weight: 0.0` default. Carry-over prompt at `~/.claude/plans/post-contract-carry-over-prompt.md` Part B / Task B2.

Track C is also gated (PM-side) on the `FIRESIDE-onboarding-embedder-download` activation issue being scheduled — not a BCK concern, noted for the reader.

roryford:
## Summary

Ships the concrete `LlamaEmbeddingBackend` that #664 left as the next milestone — a GGUF-backed implementation of the `EmbeddingBackend` protocol. Unblocks Fireside Track C (local embeddings + ship-dark semantic retrieval).

Closes #684. Tied to epic #663.

## Design notes

**Actor isolation.** `llama_context` is an opaque C pointer with no ARC story. The new backend wraps all C state in a private `actor Storage` so the context is never read or freed from more than one task at a time. The outer class is `final class … @unchecked Sendable` — the `@unchecked` is paid for by the actor confining every C-touching call, not by trusting raw pointer access. The synchronous protocol mirrors (`isModelLoaded`, `dimensions`) are guarded by `NSLock.withLock`, so callers see immediate, consistent state without forcing a hop through the actor for every read.

**Process lifecycle.** `LlamaBackendProcessLifecycle.retain()` / `release()` is shared with `LlamaBackend`. `llama_backend_init` is global per-process, and we MUST NOT call it twice — the lifecycle helper refcounts that init across both backends.

**Architecture handling.** `LlamaModelLoader.unsupportedArchitectures` (the generation-side denylist) explicitly rejects `bert`, `nomic-bert`, and `jina-bert-v2` — those are the *correct* embedding architectures, but not chat models. The embedding loader checks both lists: an architecture is allowed if it is on the embedder allowlist OR not on the generation-side denylist. Anything else is rejected, so a Whisper / SD GGUF cannot be silently routed through here.

**Pooling-aware extraction.** After `llama_encode`:
- `pooling_type != NONE` → read pooled vector via `llama_get_embeddings_seq(ctx, 0)`
- `pooling_type == NONE` → mean-pool per-token rows from `llama_get_embeddings_ith`

In both cases the result is L2-normalized to unit length so downstream cosine retrieval reduces to a dot product.

**Batching.** `embed([texts])` runs one `llama_encode` per text against `seq_id = 0`. The KV memory is cleared between texts so back-to-back calls don't poison each other. Texts longer than the embedding context are truncated head-first (CLS + leading content wins). True multi-sequence packing into a single batch is deferred until measurement justifies the extra complexity — see deviation note below.

**Failure shape.** Every C-level failure is wrapped into `EmbeddingError.encodingFailed(_:)` with a meaningful `userInfo` message. `embed` thrown before load surfaces `EmbeddingError.modelNotLoaded`. Dimension changes between back-to-back loads are observable via `dimensions`; the actual `.dimensionMismatch` enforcement lives downstream where it belongs.

**Drive-by build fix.** `LlamaGenerationDriver.swift:148` carried a bare `return` in a `Bool`-returning function — left over from the merge of #667 (grammar parsing) and #668 (Bool-return signature change). The package was un-buildable under `--traits Llama`, so this PR includes the trivial fix (`return true`, with a comment explaining why true is correct here — no decode has run, the KV cache is still coherent).

## Tests

All gated behind `#if Llama` so CI's `--disable-default-traits` run keeps skipping them. Suites:

- `LlamaEmbeddingBackendLoadUnloadTests` — initial state, `loadModel` sets `dimensions`, `unloadModel` clears it, invalid path → `encodingFailed`.
- `LlamaEmbeddingBackendEmbedTests` — single text → 1 vector, batch of 3 → 3 vectors, empty array → empty result, all unit-normalized.
- `LlamaEmbeddingBackendDeterminismTests` — identical input across two calls produces matching vectors (1e-4 per-component tolerance for Metal reduction-order drift).
- `LlamaEmbeddingBackendUnloadedErrorTests` — `embed` before load and after unload both throw `modelNotLoaded`. **No hardware needed.**
- `LlamaEmbeddingBackendDimensionMismatchTests` — load A then B, observe `dimensions` updates and new vector matches new dim.
- `LlamaEmbeddingBackendLiveFireTests` — 20 iterations on the same input; consistent vectors and median ≤500 ms per call. Opt-in via `BCK_EMBEDDING_MODEL_PATH`.

GGUF-dependent tests skip cleanly when no embedding model is reachable (env var unset and none on disk under `~/Documents/Models/`). The dimension-mismatch test additionally requires `BCK_EMBEDDING_MODEL_PATH_ALT` so it can verify the dim swap with a second model.

## What I tested locally

- `swift build --traits Llama` — clean (zero errors).
- `swift build --disable-default-traits` — clean.
- All five CI suites green: `BaseChatCoreTests` (213), `BaseChatInferenceTests` (1055), `BaseChatInferenceSwiftTestingTests` (69), `BaseChatUITests` (476), `BaseChatBackendsTests` (225 XCT + 136 ST). 0 failures, 4–8 skipped per suite (hardware/env-gated, expected).
- `swift test --filter "BaseChatBackendsTests.LlamaEmbeddingBackend" --traits Llama` — 12 tests; 4 ran (the 2 unloaded-error tests and `loadModel_invalidPath` and `initialState_isUnloaded`), 8 skipped because no embedding GGUF was on this machine. Skip messages name the env vars to set.
- The two pre-existing failures observable under `--traits Llama` (`LlamaBackendTests.test_fixture_stopGeneration_midDecode_…_regression522` and `LlamaGrammarSamplerTests.test_grammar_cancelCleansTeardown`) are independent of this PR and reproduce on `main` once the bare-`return` build error is fixed. They're filed elsewhere.

## Deviation from the issue spec

- The issue body refers to `llmfarm_core.swift` as the upstream wrapper. The repo actually depends on `mattt/llama.swift` 2.8772.0 (`Package.swift`); I followed the in-repo dependency. The C API contract — `llama_set_embeddings`, `llama_pooling_type`, `llama_get_embeddings_seq` — is the same either way.
- `embed([texts])` does NOT pack multiple texts into a single `llama_encode` call. Each text gets its own evaluation. The issue mentions batching when tokens fit; I traded that micro-optimisation for code that's straightforward to audit. Adding multi-`seq_id` packing is a follow-up if Track C measurement shows latency is dominated by per-encode setup rather than the encode itself.

## Adversarial-review self-check

- **Pointer lifetime.** `Storage.unload()` zeroes `model`/`context`/`vocab` before calling `llama_free` / `llama_model_free`, and the actor's serial executor guarantees no concurrent `encode` is reading them. `deinit` blocks on the unload via a semaphore so the process refcount release strictly follows pointer free.
- **Memory pressure / teardown.** `unloadModel()` (the public path) flips the synchronous flags first and runs the C cleanup on a detached task — the protocol caller is never blocked. Any in-flight `embed` finishes on the actor before the unload runs (serial executor again). No use-after-free window.
- **Batch-size edge cases.** Empty input array → returns `[]`. Single empty string after tokenization → returns a zeroed vector of `dimensions` floats (chosen over throwing because chat UIs frequently embed whitespace inputs). Texts that tokenize beyond the context window are truncated head-first; we keep CLS + prefix because that's what BERT-style pooling cares about.
- **Concurrent loads on different instances.** `embeddingLoadLock` serialises embedding-load `llama_model_load_from_file` calls against each other; the underlying GGML init lock serialises against `LlamaBackend`'s loads implicitly.

## Test plan

- [x] `swift build --traits Llama` succeeds
- [x] `swift build --disable-default-traits` succeeds
- [x] All five CI suites pass locally (BaseChatCoreTests + Inference + InferenceSwiftTesting + UI + Backends)
- [x] `--traits Llama` Backends suite includes the new tests (12 visible, gated skips when no GGUF)
- [ ] CI reruns the same five suites post-push
- [ ] Reviewer (or downstream Fireside) runs the live-fire suite with a real embedding GGUF before relying on the latency budget

### DIFF
--- a/Sources/BaseChatBackends/LlamaEmbeddingBackend.swift
+++ b/Sources/BaseChatBackends/LlamaEmbeddingBackend.swift
@@ -1,0 +1,491 @@
+#if Llama
+import Foundation
+import LlamaSwift
+import os
+import BaseChatInference
+
+/// llama.cpp embedding backend for GGUF embedding models (BERT, Nomic, Jina, etc.).
+///
+/// Holds a single `llama_context *` dedicated to embedding extraction, separate
+/// from any generation context owned by ``LlamaBackend``. Pointer ownership and
+/// all C-level state (model + context + vocab) are confined to a private actor
+/// so the public API can be `Sendable` while the underlying C resources are
+/// never touched from more than one task at a time.
+///
+/// ## Design notes
+///
+/// - **Actor isolation:** `llama_context` is an opaque C pointer with no ARC
+///   guarantees. Wrapping all access in an actor (`Storage`) means we never need
+///   `@unchecked Sendable` for the pointer itself — the actor's serial executor
+///   guarantees there is no concurrent decode or free. The outer
+///   `LlamaEmbeddingBackend` is a `final class` so it can be shared by reference
+///   while still satisfying `EmbeddingBackend: AnyObject, Sendable`.
+///
+/// - **Process lifecycle:** Shares ``LlamaBackendProcessLifecycle`` with
+///   ``LlamaBackend``. `llama_backend_init` is global and must only be called
+///   once per process; the lifecycle helper refcounts that init across both
+///   backends.
+///
+/// - **Architecture allowlist for embedders:** ``LlamaModelLoader``'s denylist
+///   rejects `bert`, `nomic-bert`, and `jina-bert-v2` — those are *valid*
+///   embedding architectures, but the loader treats them as unsupported because
+///   they are not causal LMs. This backend uses its own loader path that
+///   explicitly allows the embedder-family architectures.
+///
+/// - **Pooling-aware extraction:** After `llama_encode`, embeddings are
+///   read via either `llama_get_embeddings_seq` (when the model declares a
+///   pooling type other than `NONE`) or by averaging per-token outputs from
+///   `llama_get_embeddings_ith` (NONE). Vectors are L2-normalized to unit
+///   length before being returned, matching cosine-retrieval expectations.
+///
+/// - **Batching:** `embed([texts])` evaluates each text in its own
+///   `llama_encode` call against `seq_id = 0` to keep the pooled-embedding
+///   read paths simple. Texts are truncated to the context window when
+///   tokenization exceeds capacity. A future optimisation can pack multiple
+///   short texts into a single batch by assigning distinct `seq_id`s; we
+///   defer that until measurement justifies the added complexity.
+public final class LlamaEmbeddingBackend: EmbeddingBackend, @unchecked Sendable {
+
+    // MARK: - Logging
+
+    private static let logger = Logger(
+        subsystem: BaseChatConfiguration.shared.logSubsystem,
+        category: "embedding"
+    )
+
+    // MARK: - Storage
+
+    /// Owns all C-level resources. Confined to an actor so the C pointers are
+    /// never read from more than one task concurrently.
+    fileprivate actor Storage {
+        var model: OpaquePointer?
+        var context: OpaquePointer?
+        var vocab: OpaquePointer?
+        var dimensions: Int = 0
+        var contextSize: Int32 = 0
+        /// Captured pooling type. `LLAMA_POOLING_TYPE_NONE` means we manually
+        /// average per-token embeddings; any other value means the C side has
+        /// already pooled and we read via `llama_get_embeddings_seq`.
+        var poolingType: llama_pooling_type = LLAMA_POOLING_TYPE_UNSPECIFIED
+
+        var isLoaded: Bool { model != nil && context != nil }
+
+        func install(model: OpaquePointer, context: OpaquePointer, vocab: OpaquePointer, dimensions: Int, contextSize: Int32, pooling: llama_pooling_type) {
+            self.model = model
+            self.context = context
+            self.vocab = vocab
+            self.dimensions = dimensions
+            self.contextSize = contextSize
+            self.poolingType = pooling
+        }
+
+        /// Drops references and frees the C resources in the order
+        /// `llama_free` (context) → `llama_model_free` (model). Safe to call
+        /// when nothing is loaded.
+        func unload() {
+            let ctx = context
+            let mdl = model
+            context = nil
+            model = nil
+            vocab = nil
+            dimensions = 0
+            contextSize = 0
+            poolingType = LLAMA_POOLING_TYPE_UNSPECIFIED
+            if let ctx { llama_free(ctx) }
+            if let mdl { llama_model_free(mdl) }
+        }
+
+        /// Tokenizes `text`, runs `llama_encode`, reads the pooled (or manually
+        /// averaged) embedding, and returns it normalized to unit length.
+        ///
+        /// Pointers are read from `self` under the actor's serial executor —
+        /// no caller-side snapshot is needed because `unload()` cannot
+        /// interleave with this method.
+        func encode(text: String) throws -> [Float] {
+            guard let context, let vocab else {
+                throw EmbeddingError.modelNotLoaded
+            }
+            let dim = dimensions
+            let pooling = poolingType
+            let maxTokens = Int(contextSize)
+
+            // Tokenize via the existing helper. `addBos: true` matches what the
+            // BERT-family GGUFs expect (CLS at position 0); the helper is also
+            // safe for non-BERT embedders.
+            var tokens = LlamaTokenization.tokenize(text, vocab: vocab, addBos: true)
+            if tokens.isEmpty {
+                // Empty input produces a zero vector of the right shape rather
+                // than throwing — callers commonly embed user input that may be
+                // pure whitespace, and a typed exception there would be noisy.
+                return [Float](repeating: 0, count: dim)
+            }
+
+            // Truncate to context capacity. We deliberately keep the head of the
+            // sequence (CLS + leading content) rather than the tail — for BERT
+            // pooling the CLS token's representation is what gets returned, and
+            // most embedding workloads expect the document's prefix to dominate.
+            if tokens.count > maxTokens {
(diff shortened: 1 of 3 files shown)


roryford:
## Review — shipped one fix

**Critical issue found & fixed (9e16ad1):** `deinit` was blocking on a `DispatchSemaphore` to await the actor-confined unload. Two problems:

1. When the backend is owned by an `@MainActor` `InferenceService` (the expected deployment) deinit runs on the main actor. `sem.wait()` freezes the UI for the duration of C cleanup — easily 500ms+ on a 1GB embedding model.
2. Any main-actor hop inside `Storage.unload()` would deadlock outright because the main thread is parked on the semaphore.

Switched to `LlamaBackend`'s pattern: retain the process refcount, fire the unload on a detached task, release inside that task. The sync release at the end of deinit balances init's retain; the extra retain/release pair keeps `llama_backend_free` from racing with an in-flight unload. All 2174 CI-suite tests still green; `--traits Llama` embedding tests run (12 visible, 4 pass, 8 env-gated skips — no local embedding GGUF here).

**Non-blocking observations (leaving as-is or filing follow-ups):**

- **No memory-pressure integration.** `LlamaBackend.init` wires `MemoryPressureHandler` to auto-unload under pressure; `LlamaEmbeddingBackend` doesn't. Embedding GGUFs are often 1GB+, so this is worth a follow-up. Happy to open one.
- **Concurrent `loadModel` on the same instance** can in principle interleave `unload` → `install` → `unload` → `install` across two callers and leak the first model. `InferenceService` already serializes loads so it's not actively reachable; not worth a Swift-side lock without evidence.
- **`embed([])` test requires a GGUF it doesn't need** (tests line 697) — the function short-circuits before touching the model. Would let this run on any CI. Minor.
- **Drive-by `LlamaGenerationDriver.swift:148` fix** looks correct: at that point no decode has run, the KV cache is coherent, so `return true` preserves `sessionKVState` for the caller.

LGTM after the deinit fix — ready for your merge decision.

PR #723, opened 2026-04-25
fix(inference): reject GGUF files without magic-bytes header in discovery

roryford:
## Summary

`ModelInfo(ggufURL:)` accepted any file with a `.gguf` extension, so 1 KB stub files surfaced in the model picker as fake entries. This came up as a live user observation ("many model entries of 1KB" in the picker on dev machines) — the same leak `scripts/clean-leaked-test-artifacts.sh` has been hammering manually.

The infrastructure to validate GGUF files already exists: `GGUFMetadataReader.isValidGGUF(at:)` reads the first 4 bytes and checks against `[0x47, 0x47, 0x55, 0x46]` ("GGUF"). The fix is one extra guard in the GGUF initializer — `ModelStorageService.discoverModels()` inherits the filter for free, so the picker now rejects leaked test stubs, truncated downloads, and any `.gguf`-extensioned companion file without a real header.

## What changed

- `Sources/BaseChatInference/Models/ModelInfo.swift`: add `guard GGUFMetadataReader.isValidGGUF(at: url) else { return nil }` to `init?(ggufURL:)`, after the extension check and before reading file attributes.
- Test fixtures updated: `ModelStorageServiceTests.createGgufFile` and `ModelInfoTests` now emit the real magic header (`0x47 0x47 0x55 0x46`) padded to the requested size so they continue to pass the gate.
- Two new tests in `ModelInfoTests` lock the contract: `test_ggufInit_rejectsFileWithoutMagicBytes` (15-byte ASCII placeholder → nil, mirroring the real leak shape) and `test_ggufInit_acceptsFileWithMagicBytes` (real header → non-nil).

Scope is deliberately tight: only the GGUF init changes. `discoverModels`, `DownloadFileValidator`, and the cleanup script are untouched.

#### Discovery rejects non-GGUF `.gguf` files

Leaked test fixtures and truncated downloads that carried a `.gguf` extension but no real header used to appear in the model picker as 1 KB phantom entries. `ModelInfo(ggufURL:)` now requires the GGUF magic header (`0x47 0x47 0x55 0x46`) before constructing a model, so `refreshModels()` becomes self-healing — the picker filters these files out automatically instead of relying on `scripts/clean-leaked-test-artifacts.sh`.

` ` `swift
// Rejected: empty .gguf stubs, partial downloads, companion files
ModelInfo(ggufURL: stubURL) // → nil

// Accepted: files whose first 4 bytes are "GGUF"
ModelInfo(ggufURL: realURL) // → ModelInfo(modelType: .gguf, ...)
` ` `

No API change; purely a behaviour tightening on the discovery path.

## Test plan

- [x] `swift test --filter BaseChatInferenceTests --disable-default-traits` → 1076 tests, 0 failures
- [x] New `test_ggufInit_rejectsFileWithoutMagicBytes` asserts the fix
- [x] New `test_ggufInit_acceptsFileWithMagicBytes` confirms the happy path still works

### DIFF
--- a/Sources/BaseChatInference/Models/ModelInfo.swift
+++ b/Sources/BaseChatInference/Models/ModelInfo.swift
@@ -82,6 +82,14 @@ public struct ModelInfo: Identifiable, Hashable, Sendable {
     public init?(ggufURL url: URL) {
         guard url.pathExtension.lowercased() == "gguf" else { return nil }
 
+        // Reject files that don't carry the GGUF magic header — covers:
+        //   - leaked test-fixture stubs (e.g. 15-byte ASCII placeholders, see
+        //     `scripts/clean-leaked-test-artifacts.sh`)
+        //   - downloads truncated before the first 4 bytes were written
+        //   - companion files some HF repos ship with a .gguf extension but no
+        //     header (rare, but cheap to filter)
+        guard GGUFMetadataReader.isValidGGUF(at: url) else { return nil }
+
         let fileManager = FileManager.default
         guard let attributes = try? fileManager.attributesOfItem(atPath: url.path),
               let size = attributes[.size] as? UInt64 else {
--- a/Tests/BaseChatInferenceTests/ModelInfoTests.swift
+++ b/Tests/BaseChatInferenceTests/ModelInfoTests.swift
@@ -18,12 +18,25 @@ final class ModelInfoTests: XCTestCase {
         super.tearDown()
     }
 
+    // MARK: - Helpers
+
+    /// Writes a minimal fixture with a real GGUF magic-bytes header padded to the requested size.
+    /// The first 4 bytes are `[0x47, 0x47, 0x55, 0x46]` so the file passes
+    /// `GGUFMetadataReader.isValidGGUF(at:)` without needing a full GGUF structure.
+    @discardableResult
+    private func writeGGUFFixture(at url: URL, totalSize: Int = 1024) throws -> URL {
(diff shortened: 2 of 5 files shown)


roryford (commented):
Reviewed. The magic-bytes guard is the right fix for discovery — both production callers (`ModelStorageService.discoverModels` and `ModelManagementViewModel.importedModel`) now correctly reject leaked placeholder fixtures and pre-magic truncated downloads. Fail-closed on unreadable files is the right policy for discovery. Perf impact is trivial (4-byte read per `.gguf` file, called on-demand not per-render).

**Pushed a follow-up (71dcce7):** CI surfaced two UI test fixture helpers the original PR missed — `GenerationViewModelTests.createFakeGGUF` and `ModelManagementViewModelTests.test_importModel_withGGUF_copiesFileIntoModelsDirectory` were writing zero/0xAB-filled placeholders that now fail discovery. Same magic-bytes prefix pattern as `ModelInfoTests.writeGGUFFixture` and `ModelStorageServiceTests.createGgufFile`. Local `BaseChatUITests` now 479/479 passing; `BaseChatInferenceTests` 1076/1076; `BaseChatInferenceSwiftTestingTests` 69/69; `BaseChatBackendsTests` 136/136; `BaseChatCoreTests` 214/214; `BaseChatToolsTests` 36/36; `BaseChatTestSupportTests` 13/13 — all green.

**Minor (won't block):** the comment block added to `ModelInfo.swift` is slightly long for the project's "explain why, not what" convention — the bullet list describes what's covered rather than the underlying rationale. Not worth a round-trip, just a style note.

Issue #719, opened 2026-04-25
Phase 2B: TrafficBoundaryAuditTest rules 5+7 — closes part of #714


roryford:
**Phase 2B of #714.** Adds the two audit rules that depend on the trait split.

Blocked by Phase 2A (traits must exist).

## Scope

Extend `Tests/BaseChatInferenceTests/TrafficBoundaryAuditTest.swift` with two new rules:

### Rule 5 — `Package.swift` hygiene

Banned in `Package.swift`:
- `linkedFramework("Network")`
- `linkedFramework("CFNetwork")`
- `unsafeFlags`
- `.buildToolPlugin`
- `.commandPlugin` (unless allowlisted)

Catches future PRs that re-introduce networking via build settings (defeating source-grep audit) or add SwiftPM plugins (which can run arbitrary code at build time).

### Rule 7 — Trait gate sanity

For every `#if <Identifier>` in `Sources/`, where `<Identifier>` looks trait-named (TitleCase or UPPERCASE), assert it references a trait declared in `Package.swift`. Catches stale `#if` directives after a trait rename.

Implementation: regex-parse `Package.swift` for trait names, then scan source files for `#if`/`#elseif` guards.

### Verification

- New sabotage tests for rules 5 and 7 (positive + negative cases)
- Audit passes against trait split landed in Phase 2A
- Sabotage: add `linkedFramework("Network")` to `Package.swift` → rule 5 fails
- Sabotage: add `#if Bogus` to a source file → rule 7 fails

Plan reference: `.claude/plans/shifting-to-plan-mode-shiny-sutherland.md` §F.

roryford:
Already shipped via PR #740 (commit on main).

What's on `origin/main` covering all acceptance criteria:
- Rule 5 `test_rule5_packageManifestHygiene` at `Tests/BaseChatInferenceTests/TrafficBoundaryAuditTest.swift:384` — bans `linkedFramework("Network")`, `linkedFramework("CFNetwork")`, `unsafeFlags(`, `.buildToolPlugin(`, `.commandPlugin(`. Empty allowlist.
- Rule 7 `test_rule7_traitGateSanity` at line 416 — parses `.trait(name: "X", ...)` from `Package.swift` via `parseDeclaredTraits`, walks `Sources/`, extracts trait-shaped `#if` identifiers, asserts each is declared.
[...]

Issue #714, opened 2026-04-25
feat(security): local-only build modes + privacy validation infra


roryford:
## Context

Strategic opportunity to position BCK for regulated verticals (healthcare, legal, finance, defense, EU residency) by making local-only and datacenter-only builds **provably free** of SaaS cloud backend code, plus mechanical CI-enforced guarantees that future code changes don't silently regress the privacy posture.

Plan refined through five persona reviews (security architect, PM, QA, tin-foil-hat, DevEx) — the full plan is at `.claude/plans/shifting-to-plan-mode-shiny-sutherland.md` (local).

## Outcome

Four supported build modes — `offline` / `ollama` / `saas` / `full` — with multi-layered validation:

1. **Runtime network-isolation** test using `DenyAllURLProtocol` registered on `URLSessionConfiguration.default` + `.ephemeral`, exercised across Foundation/MLX/Llama backends with a coverage-asserted matrix.
2. **Source audit** (`TrafficBoundaryAuditTest`) covering 7 rule classes — network I/O imports, C interop / dynamic dispatch, hostname literals + obfuscation, privacy-sensitive Apple APIs (CloudKit/Handoff/iCloud/pasteboard/`os_log` `.public`), `Package.swift` hygiene, import-graph boundary, `#if` trait-name validity. Path-only allowlist hard-capped at 15 with mandatory justification.
3. **Build-mode CI matrix** (cost-optimized: Linux compile-only for offline/ollama, macOS test-only for saas/full).
4. **Binary symbol audit** (defense in depth — explicitly *not* sold as procurement evidence; string obfuscation defeats it).
5. **Runtime kill-switch** at `URLSessionProvider.networkDisabled` for belt-and-suspenders runtime locking.
6. `SECURITY.md` + `docs/THREAT_MODEL.md` with explicit known-non-mitigations section (macro plugin sandbox, xcframework checksum, FIPS, key zeroization, KV-cache residue).

## Phasing

- **Phase 1** (4-5 days): validation infra against current codebase. Ships value without trait split.
- **Phase 2** (5-7 days): trait split (`Ollama`, `CloudSaaS`), `#warning` stubs for gated types, `availableInBuild`, kill-switch.
- **Phase 3** (2-3 days): build-mode CI matrix + binary audit script.
- **Phase 4** (2 days): docs (SECURITY.md, THREAT_MODEL.md, CONTRIBUTING.md, README/CLAUDE updates).
[...]

PR #740, opened 2026-04-25
feat(test): add Package.swift hygiene + #if trait sanity audit rules (Phase 2B of #714)

roryford:
## Summary

Extends `TrafficBoundaryAuditTest` with two new source-level guard rules promised in the file's header comment.

**Rule 5 — `Package.swift` hygiene.** Bans `linkedFramework(\"Network\")`, `linkedFramework(\"CFNetwork\")`, `unsafeFlags`, `.buildToolPlugin`, and `.commandPlugin` in the manifest. Linking `Network.framework` or `CFNetwork` via build settings would re-introduce networking surface that the source-grep audit cannot see; SwiftPM plugins can run arbitrary code at build time. `packageHygieneAllowlist` starts empty (cap 3).

**Rule 7 — Trait gate sanity.** Regex-parses `.trait(name: \"X\", ...)` out of `Package.swift` to build the allowed set, then walks every `#if`/`#elseif` directive in `Sources/` and validates each trait-shaped (TitleCase, bare) identifier. Function-form conditionals (`os(...)`, `canImport(...)`, `swift(...)`, `compiler(...)`, `arch(...)`, `targetEnvironment(...)`) are skipped automatically. `DEBUG`/`RELEASE`/`NDEBUG`/`TESTING` are whitelisted. Compound expressions (`Ollama || CloudSaaS`, `Llama && Fuzz`) and negations (`!Ollama`) are split and validated piecewise.

The header doc-comment is updated to reflect the new rule set; the placeholder line "Rules 5 and 7 … land in Phase 2" is removed.

Closes part of #714 — see #719. Phase 2B of #714.

## Sabotage verification

- **Rule 5**: temporarily added `linkerSettings: [.linkedFramework(\"Network\")]` to `BaseChatInference` target — `test_rule5_packageManifestHygiene` failed with the expected offender output. Reverted before commit.
- **Rule 7**: temporarily added `#if Bogus … #endif` to `Sources/BaseChatInference/Models/ActivityPhaseStateMachine.swift` — `test_rule7_traitGateSanity` failed naming `Bogus` as the unknown identifier. Reverted before commit.
- Inline matcher-level sabotage tests (`test_sabotage_rule5_*` / `test_sabotage_rule7_*`) cover positive + negative cases for each new helper without requiring fixture files under `Sources/`.

## Test plan

- [x] `swift test --filter BaseChatCoreTests --disable-default-traits`
- [x] `swift test --filter BaseChatInferenceTests --disable-default-traits` — all new tests pass (one pre-existing `SilentCatchAuditTest` failure on `main` re `MLXToolDialect.swift` `try?` — unrelated to this PR)
- [x] `swift test --filter BaseChatInferenceSwiftTestingTests --disable-default-traits`
- [x] `swift test --filter BaseChatUITests --disable-default-traits`
- [x] `swift test --filter BaseChatBackendsTests --disable-default-traits`
- [x] `swift test --filter BaseChatTestSupportTests --disable-default-traits`

## Release Note

**Audit rules 5 + 7 — `Package.swift` hygiene and `#if` trait sanity** — the existing `TrafficBoundaryAuditTest` now also rejects re-introduction of networking via SwiftPM `linkedFramework`/`unsafeFlags` settings and SwiftPM build/command plugins, and validates that every `#if <Trait>`/`#elseif <Trait>` identifier under `Sources/` matches a trait declared in `Package.swift`. Catches stale gates after trait renames and prevents the source-grep audit from being silently bypassed via build settings.

### DIFF
--- a/Tests/BaseChatInferenceTests/TrafficBoundaryAuditTest.swift
+++ b/Tests/BaseChatInferenceTests/TrafficBoundaryAuditTest.swift
@@ -33,12 +33,23 @@ import XCTest
 ///    (`os_log` `.public` deliberately not covered — see the
 ///    `privacyAPIPattern` doc comment for rationale.)
 ///
+/// 5. **`Package.swift` hygiene** — bans `linkedFramework("Network")`,
+///    `linkedFramework("CFNetwork")`, `unsafeFlags`, and SwiftPM
+///    `.buildToolPlugin` / `.commandPlugin` declarations. Catches future
+///    PRs that re-introduce networking via build settings (defeating the
+///    source-grep audit) or add SwiftPM plugins (which can run arbitrary
+///    code at build time).
+///
 /// 6. **Import-graph boundary** — locks in the layered architecture from
 ///    `CLAUDE.md`. UI must not depend on Backends; Inference must not
 ///    depend on Core or Backends.
 ///
-/// (Rules 5 and 7 — Package.swift hygiene and `#if` trait-name validity —
-/// land in Phase 2 alongside the trait split.)
+/// 7. **Trait gate sanity** — every `#if`/`#elseif` identifier in
+///    `Sources/` that looks trait-named (TitleCase or UPPERCASE, not a
+///    well-known compiler conditional like `os(...)`, `canImport(...)`,
+///    `swift(...)`, `DEBUG`, etc.) must match a trait declared in
+///    `Package.swift`. Catches stale `#if` directives left behind after
+///    a trait rename.
 ///
 /// ## Allowlist policy
 ///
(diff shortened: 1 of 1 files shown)


roryford:
## Code review

**LGTM with one nit fixed in [a25d16d](https://github.com/roryford/BaseChatKit/commit/a25d16d).**

### Strengths

- Both rules follow the existing fingerprint/allowlist/cap pattern from rules 1–6 — minimal cognitive overhead for reviewers.
- Sabotage tests are thorough: positive matches for every banned token, negatives that prove the matcher doesn't false-fire on prose, and a parser round-trip for `parseDeclaredTraits` against a fixture manifest.
- The empty allowlist + cap of 3 for Rule 5 is the right starting posture — `unsafeFlags` and SwiftPM plugins are exactly the kind of escape hatches that warrant individual review when added.
- The `traitGateCompilerConditionals` set is explicit and small (`DEBUG`, `RELEASE`, `NDEBUG`, `TESTING`); adding to it would be visible in code review.

### Nit (fixed in fix-up commit)

- **Rule 7 parser** silently drops parenthesised groupings like `#if (Ollama || Bogus)` because the contains-`(` filter treats them as function-form. No `Sources/` file uses parenthesised `#if` groups today (`grep -rn '#if (' Sources/` returns nothing), so this is a safe simplification rather than a real hole — but I added a doc-comment caveat under `extractTraitLikeIdentifiers` so the next contributor who reaches for that style knows the audit may not catch typos in that form. Optional: a future PR could rewrite the tokeniser to flatten `(...)` groups recursively if the codebase adopts parenthesised gates.

- The header-comment update reflects the new state clearly. Rules are listed 1, 2, 3, 4, 5, 6, 7 with 5 inserted between 4 and 6 — consistent with the existing numbering.
- Rule 5's `shouldScan(line:)` filter inherits the comment-skip logic, so prose like `/// Don't use unsafeFlags` won't self-trip the audit. Good defensive layering.

Local: `swift test --filter TrafficBoundaryAuditTest --disable-default-traits` → 14/14 pass on `a25d16d`.

PR #724, opened 2026-04-25
refactor(traits): split Ollama and CloudSaaS into opt-in traits (Phase 2A of #714)

roryford:
Closes #718. Phase 2A of the local-only-build initiative (#714) — the foundational trait split that every other Phase 2 ticket depends on.

## What's in this PR

**Trait split** (`Package.swift`)
- Adds `Ollama` and `CloudSaaS` traits.
- `Ollama` stays in the default trait set for one minor release (deprecation cycle); the next major moves it out. `CloudSaaS` is opt-in immediately.
- Adds `.define(...)` for both to every target that already has the MLX/Llama defines.

**Source gating**
- `#if CloudSaaS` (whole file): `ClaudeBackend`, `OpenAIBackend`, `SSECloudBackend`, `CloudErrorSanitizer`, `DNSRebindingGuard`, `PinnedSessionDelegate`.
- `#if Ollama`: `OllamaBackend`, `OllamaModelListService`.
- `#if Ollama || CloudSaaS`: `URLSessionProvider` factories (kill-switch outside the gate), settings views.
- `DefaultBackends` gates each cloud factory case behind the matching trait; iterates `APIProvider.availableInBuild` instead of `allCases`.

**`#warning` stubs** for `ClaudeBackend` / `OpenAIBackend` / `OllamaBackend` — turn missing-symbol errors into self-explanatory diagnostics with the exact `.package(...)` syntax to add the trait.

**`APIProvider.availableInBuild`** — static computed under `#if`. Enum cases remain always-present (data-only) so `ConversationRecords.selectedEndpointID` schema is unchanged. Replaces `allCases` at the two non-test call sites.

**`URLSessionProvider.networkDisabled` runtime kill-switch** — always available outside trait gates. Both ergonomic non-throwing accessors (preconditions) and recoverable `throwingPinned()` / `throwingUnpinned()` variants. 7-test suite covers each factory's enabled/disabled paths and reset semantics, sabotage-verified.

## Verification

| Build mode | Command | Result |
|------------|---------|--------|
| offline | `swift build --disable-default-traits` | ✓ |
| ollama | `swift build --disable-default-traits --traits Ollama` | ✓ |
| saas | `swift build --disable-default-traits --traits CloudSaaS` | ✓ |
| full | `swift build --traits MLX,Llama,Ollama,CloudSaaS` | ✓ |

All 6 CI test suites pass under `--disable-default-traits`:

- BaseChatCoreTests — 213 tests
[...]

Issue #720, opened 2026-04-25
Phase 2C: deprecation + CHANGELOG for trait-gated backends — closes part of #714


roryford:
**Phase 2C of #714.** Migration artifacts so existing consumers aren't blindsided by the trait split.

Blocked by Phase 2A.

## Scope

1. **`@available(*, deprecated, message: ...)`** on `OllamaBackend.init`, referencing the trait requirement. Helps consumers using `OllamaBackend` directly (rather than via `DefaultBackends.register`) see a warning before the next major release moves Ollama out of the default trait set.
2. **`CHANGELOG.md` migration section.** Major version bump. Document:
   - `Ollama` remains in default traits for this release (deprecation cycle)
   - Next major: `Ollama` moves to opt-in; `CloudSaaS` already opt-in
   - How to add the trait to a consumer `.package(...)` entry

3. **`README.md`** build-mode decision table keyed by use case (consumer app → default, regulated vertical → `--disable-default-traits --traits MLX,Llama`, datacenter → `+ Ollama`, full → all).

### Verification

- Building a consumer app that uses `OllamaBackend()` directly with default traits succeeds with a deprecation warning (not a hard error).
- CHANGELOG renders cleanly via Release Please.
- README example commands actually compile.

Plan reference: `.claude/plans/shifting-to-plan-mode-shiny-sutherland.md` §L.

roryford:
Shipped via #741 (commit baedf35).

Verified all three deliverables landed on main:

1. `@available(*, deprecated, ...)` annotation on `OllamaBackend.init(urlSession:)` and `OllamaBackend.makeChecked(urlSession:)` at `Sources/BaseChatBackends/OllamaBackend.swift:70` and `:81` with the message pointing at #714.
2. README "Build modes" decision table at `README.md:85` keyed by use case (default consumer / regulated vertical / datacenter / full).
3. Migration prose written into PR #741's `## Release Note` block — will land in CHANGELOG.md via the next Release Please rewrite per documented workflow.

Issue #721, opened 2026-04-25
Phase 2D: trait-gate existing tests — closes part of #714


roryford:
**Phase 2D of #714.** Mechanical: gate existing cloud-backend tests behind the new traits so they only run in builds that include the relevant code.

Blocked by Phase 2A.

## Scope

Add `#if CloudSaaS` and `#if Ollama` guards (matching the `#if MLX` / `#if Llama` convention) to test files that exercise gated backend code:

- `Tests/BaseChatBackendsTests/ClaudeBackend*.swift` → `#if CloudSaaS`
- `Tests/BaseChatBackendsTests/OpenAIBackend*.swift` → `#if CloudSaaS`
- `Tests/BaseChatBackendsTests/SSECloud*.swift` → `#if CloudSaaS`
- `Tests/BaseChatBackendsTests/CloudErrorSanitizer*.swift` → `#if CloudSaaS`
- `Tests/BaseChatBackendsTests/RetryPolicyIntegrationTests.swift` → `#if CloudSaaS`
- `Tests/BaseChatBackendsTests/OllamaBackend*.swift` → `#if Ollama`
- `Tests/BaseChatBackendsTests/OllamaToolCallingTests.swift` → `#if Ollama`
- `Tests/BaseChatE2ETests/OllamaThinkingE2ETests.swift` → `#if Ollama`

Tests that exercise enum integrity (e.g. `APIConfigurationLogicTests`, `ModelCompatibilityTests`) keep iterating `APIProvider.allCases` — they test enum shape, not backend behaviour, and stay trait-independent.

### Verification

- `swift test --disable-default-traits` runs only trait-independent tests; cloud tests skipped (not failed).
- `swift test --traits CloudSaaS` runs the gated CloudSaaS tests.
- `swift test --traits Ollama` runs Ollama tests.
- Existing `swift test --filter BaseChatBackendsTests` passes with default traits (no behaviour change for current consumers).

Plan reference: `.claude/plans/shifting-to-plan-mode-shiny-sutherland.md` §B step 10.

roryford:
Closing as already complete: PR #724 (Phase 2A trait-split foundation) gated every test file listed in this issue's scope.

Verified on `main` @ d9e7c00:
[...]

PR #741, opened 2026-04-25
docs(traits): document trait-split migration (Phase 2C of #714)

roryford:
## Summary

Phase 2C of #714 — see #720. Adds the migration artifacts for the Ollama / CloudSaaS trait split that landed in #724:

- **Deprecation annotations** on `OllamaBackend.init(urlSession:)` and `OllamaBackend.makeChecked(urlSession:)` warning consumers that `Ollama` will move to opt-in in the next major. The file is `#if Ollama`-gated, so the warning only fires for the audience that actually has the trait enabled.
- **README "Build modes" section** with a use-case-keyed decision table (default / regulated-local / datacenter / full) and a concrete `.package(... traits: ...)` example.
- **Internal-caller `// FIXME(#714):` markers** in `DefaultBackends.register(_:)`, `bck-tools/main.swift`, and `fuzz-chat/OllamaFuzzFactory.swift` — these are infrastructure that intentionally exercises the trait-gated init and the warning is expected until Phase 2D replaces it with trait-gated unavailability.

Closes part of #714. Full migration prose lives in the Release Note block below for the next release-rewrite step to pick up.

## Test plan

- [x] `swift build` (defaults) — succeeds; three expected deprecation warnings on tagged internal callers.
- [x] `swift build --disable-default-traits --traits MLX,Llama` — succeeds.
- [x] `swift build --disable-default-traits --traits MLX,Llama,Ollama` — succeeds.
- [x] `swift build --traits MLX,Llama,Ollama,CloudSaaS` — succeeds.
- [x] `BaseChatCoreTests` — pass (214 tests, 4 skipped).
- [x] `BaseChatInferenceSwiftTestingTests` — pass (69 tests).
- [x] `BaseChatUITests` — pass.
- [x] `BaseChatBackendsTests` — pass (75 tests, 1 skipped).
- [x] `BaseChatTestSupportTests` — pass.
- [ ] `BaseChatInferenceTests` — pre-existing failure in `SilentCatchAuditTest` flagging MLX `try?` fingerprints unrelated to this PR (verified by stashing the PR's changes; same failure on the parent commit). Not introduced here.

#### Trait-split migration: Ollama deprecation cycle starts now

The Ollama and Claude / OpenAI backends are now split behind Swift package traits so consumers in regulated or air-gapped environments can compile out everything that touches the network. `Ollama` ships in the default trait set this minor for backwards compatibility; **in the next major it moves to opt-in** alongside the already-opt-in `CloudSaaS` trait. Direct call sites of `OllamaBackend()` now emit a deprecation warning pointing at the new README "Build modes" table.

```swift
.package(
    url: "https://github.com/roryford/BaseChatKit.git",
    from: "0.11.0",
    traits: [
        .trait(name: "MLX"),
        .trait(name: "Llama"),
        .trait(name: "Ollama"),       // remove for local-only builds
        .trait(name: "CloudSaaS"),    // add to enable Claude / OpenAI
    ]
)
```

Routing backend registration through `DefaultBackends.register(_:)` continues to work and will be the supported migration path: it gates Ollama on the trait so consumers do not need to touch their existing call sites today. See README "Build modes" for the four pre-blessed configurations (default / regulated-local / datacenter / full) and the build commands that match each one.

See [#714], [#720], [#724].

### Features

- **traits:** add README "Build modes" section with use-case-keyed decision table and consumer `.package(...)` example ([#720])

### Deprecations

- **traits:** `OllamaBackend.init(urlSession:)` and `OllamaBackend.makeChecked(urlSession:)` are now deprecated — Ollama moves to opt-in in the next major, route through `DefaultBackends.register(_:)` or add the `Ollama` trait explicitly to your manifest ([#720])

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,35 @@ Add the targets you need:
 ])
 ```
 
-### 2. Configure at app startup
+### 2. Build modes
+
+BaseChatKit ships four pre-blessed build configurations keyed by use case. Backends are gated behind Swift package traits so consumers in regulated or air-gapped environments can compile out everything that touches the network.
+
+| Use case | Build command | Backends available |
+|----------|---------------|--------------------|
+| Default consumer app | `swift build` | MLX, Llama, Ollama |
+| Regulated vertical (local-only, no networking) | `swift build --disable-default-traits --traits MLX,Llama` | MLX, Llama |
+| Self-hosted / private datacenter | `swift build --disable-default-traits --traits MLX,Llama,Ollama` | MLX, Llama, Ollama |
+| Full / SaaS-enabled | `swift build --traits MLX,Llama,Ollama,CloudSaaS` | MLX, Llama, Ollama, Claude, OpenAI |
+
+Pass the matching set as `traits:` on your `.package(...)` entry to lock the configuration in a consumer manifest:
+
+```swift
+.package(
+    url: "https://github.com/roryford/BaseChatKit.git",
+    from: "0.11.0",
+    traits: [
+        .trait(name: "MLX"),
+        .trait(name: "Llama"),
+        .trait(name: "Ollama"),       // remove for local-only builds
+        .trait(name: "CloudSaaS"),    // add to enable Claude / OpenAI
+    ]
+)
+```
+
+`CloudSaaS` is opt-in today. **`Ollama` is in the default trait set for the current minor release but moves to opt-in in the next major** — call sites that construct `OllamaBackend()` directly emit a deprecation warning pointing at this table. Route through `DefaultBackends.register(_:)` (which gates Ollama on the trait) or add `Ollama` explicitly to your manifest to silence it. See [#714](https://github.com/roryford/BaseChatKit/issues/714).
+
+### 3. Configure at app startup
 
 ```swift
 import BaseChatCore
@@ -140,7 +168,7 @@ struct MyApp: App {
 }
 ```
 
-### 3. Wire up the UI
+### 4. Wire up the UI
 
 ```swift
 struct ContentView: View {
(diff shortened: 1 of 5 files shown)


roryford:
## Code review

**LGTM with one nit fixed in [5159c8e](https://github.com/roryford/BaseChatKit/commit/5159c8e).**

### Strengths

- Deprecation message text is actionable ("Add the `Ollama` trait to your .package(...) entry, or register via DefaultBackends.register(_:)") and points at both the README table and #714 — consumers hitting the warning have a one-click path to the fix.
- Annotation lives on **both** init surfaces (`init(urlSession:)` and `makeChecked(urlSession:)`), so callers using either path get the warning.
- README placement (between Quick Start steps 1 and 3, with subsequent steps renumbered) is the right home — anyone copy-pasting the install snippet will see the build-mode table next.
- Internal callers (`DefaultBackends.register`, `bck-tools`, `fuzz-chat`) tagged with `// FIXME(#714):` rather than suppressed via `@available` workarounds — preserves the warning signal for direct callers of `OllamaBackend()` while documenting why we're tolerating noise here.
- Migration content placed in PR-body `## Release Note` per repo convention rather than in a freestanding `MIGRATION.md` — matches how recent releases (e.g. v0.11.2 Highlights format) handle migrations.

### Nit (fixed in fix-up commit)

- All three `// FIXME(#714):` comments referenced **"Phase 2D"** as the moment Ollama leaves the default trait set. Phase 2D is #721 (test gating) — already merged into the trait split via #724 — and **does not** flip Ollama's default-trait status. That happens at the next major version bump. Updated the wording to "the next major release" so the next reader doesn't go looking for a 2D commit that never made the flip.

- Deprecation message length (~234 chars) shows on two lines in the Xcode tooltip but stays scannable. Trimming would lose the actionable trait/factory pointer; keeping it.
- `from: "0.11.0"` in the README `.package(...)` example is fine — SemVer admits all 0.11.x releases including the current 0.11.4.
- 3 expected deprecation warnings firing under default `swift build` are documented in the PR body and verified locally; no `warnings-as-errors` in `swiftSettings` so they don't fail the build.

PR #743, opened 2026-04-25
fix(test): allowlist MLX tool-calling try? probes in SilentCatchAuditTest

roryford:
## Summary

`SilentCatchAuditTest` has been failing on `main` since #725 (MLX tool calling) with 9 unapproved silent swallows. Each is a legitimate best-effort optional conversion — same shape as existing allowlisted entries — so adding them to the allowlist with per-site rationale rather than rewriting to `do/catch`.

## Why allowlist (not rewrite)

Per the audit's own header comment: \"if the \`try?\` is a legitimate optional conversion (e.g., \`guard let x = try? Decoder.decode(...)\`) or an intentional best-effort cleanup, append its fingerprint to allowlist.\" All 9 fingerprints fit that profile:

- **MLXToolDialect.detect** — explicit best-effort \`config.json\` probe per its doc contract (\"missing or unreadable config maps to \`.unknown\`\"), structurally identical to the already-allowlisted \`MLXBackend.validateArchitecture\` pair.
- **MLXBackend.effectiveSystemPrompt** — Encoder→JSONObject round-trip used as a shape-conversion probe so tool parameters can merge into a \`[String: Any]\` dict; explicit fallback to empty schema, which Qwen tolerates.
- **MLXBackend tool-call history** — re-serialise model-emitted \`<tool_call>\` JSON for the next chat turn; if the round-trip fails we drop the single block rather than failing the whole prompt.
- **MLXToolCallParser** — defensive parse of buffered payloads between \`<tool_call>…</tool_call>\` (models routinely emit truncated/malformed JSON mid-stream), same as SSE \`parseToken\` family.

Each entry carries a comment explaining what the failure mode is and why the fallback is safe.

## Test plan

- [x] \`swift test --filter SilentCatchAuditTest --disable-default-traits\` passes (was failing on \`main\` @ d9e7c00).
- [x] All 6 trait-disabled CI suites green locally: Core 214, Inference 1076, InferenceSwiftTesting 69, UI 479, Backends + TestSupport clean.
- [ ] CI green.

## Release Note

Internal-only fix. Not user-visible — the audit gates source-tree hygiene. No runtime change.

### DIFF
--- a/Tests/BaseChatInferenceTests/SilentCatchAuditTest.swift
+++ b/Tests/BaseChatInferenceTests/SilentCatchAuditTest.swift
@@ -104,6 +104,56 @@ final class SilentCatchAuditTest: XCTestCase {
         // than masking it with a false architecture error.
         "BaseChatBackends/MLXBackend.swift:guard let data = try? Data(contentsOf: configURL),",
         "BaseChatBackends/MLXBackend.swift:let json = try? JSONSerialization.jsonObject(with: data) as? [String: Any] else {",
+        // MLXBackend.effectiveSystemPrompt (Qwen tool-block builder): the
+        // `tool.parameters` round-trip through Encoder→JSONObject is a
+        // shape-conversion probe so the parameters merge into the surrounding
+        // [String: Any] tools array. The else-branch falls back to an empty
+        // schema, which is the documented Qwen behaviour for parameterless
+        // tools — failure here is functionally indistinguishable from a tool
+        // declaring no parameters.
+        "BaseChatBackends/MLXBackend.swift:if let paramsData = try? JSONEncoder().encode(tool.parameters),",
+        "BaseChatBackends/MLXBackend.swift:let paramsObj = try? JSONSerialization.jsonObject(with: paramsData) {",
+        // Same builder, outer wrap: serialising the assembled tools array into
+        // pretty-printed JSON for the <tools> block. On failure we emit "[]",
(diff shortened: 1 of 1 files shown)


roryford:
## Self-review

Submitting this so #740 and #741 unblock — `SilentCatchAuditTest` was failing on `main` since #725 (MLX tool calling), which adds 9 `try?` call sites that the audit treats as unapproved silent swallows.

### Why allowlist (not rewrite to do/catch)

Each fingerprint matches the existing pattern in `MLXBackend.validateArchitecture` (config.json best-effort probe), the SSE backends' `parseToken` family (defensive payload parses), or `JSONSchemaValue`'s try-each-type-in-order decoder. The audit's own header explicitly allows these:

> if the `try?` is a legitimate optional conversion (e.g., `guard let x = try? Decoder.decode(...)`) or an intentional best-effort cleanup, append its fingerprint to allowlist.

Each new entry carries a per-block comment describing what the failure mode is and why the fallback is safe. Reviewers should sanity-check the rationale rather than the line count.

### Verification

- `swift test --filter SilentCatchAuditTest --disable-default-traits` was failing on `d9e7c00` with 9 offenders, now passes.
- All 6 trait-disabled CI suites green locally: Core 214, Inference 1076, InferenceSwiftTesting 69, UI 479, Backends + TestSupport clean.
- `Package.resolved` drift reset before commit; not in the diff.

Issue #758, opened 2026-04-25
track(mlx): runtime tuning, observability, and config parity


roryford:
Umbrella tracking 6 small MLX-runtime-side tweaks. Each lands independently — no shipping order, no umbrella PR.

- [x] #745 thermal pause — shipped in #764
- [ ] #747 yield every N tokens — tiny, MLXBackend tight loop (~3 LOC + test)
- [x] #746 prefill_progress SSE events — shipped in #804 (OpenAI passthrough)
- [ ] #748 HF config.json vision/audio probe — medium, ModelLoader + capability gating
- [ ] #749 KV-cache prefix reuse — large, dedicated design pass before PR
- [ ] #750 GenerationConfig audit — research issue; may produce 0–N follow-ups

Note: These are MLX-only, runtime-side tweaks. They land independently when convenient — no shipping order required, no umbrella PR. The umbrella exists so they aren't 6 noisy backlog rows.

Issue #748, opened 2026-04-25
feat(inference): probe HF config.json for vision/audio capability before loading


roryford:
## Goal

Add a probe that reads a model's `config.json` (and `preprocessor_config.json` if present) and exposes capability flags — `supportsVision`, `supportsAudio` — without loading the model.

## Why

- Lets the UI gate "attach image" affordances per-model without trial-and-error.
- Lets the inference layer pre-route to `VLMModelFactory` / `LLMModelFactory` correctly.
- Required to drive the FitStatus / model-browser UX (#310).

## Approach

**Durable signal** for vision: `config["vision_config"] != nil` (and similarly `audio_config` for ALMs). This works across model families without a hardcoded list.

**Avoid** the SwiftLM approach of hardcoding `model_type in {paligemma, qwen2_vl, gemma3, …}` — that list rots quickly as new VLMs ship.

## Scope

- New type in `BaseChatInference`: `ModelCapabilityProbe`.
- Reads `config.json` from a model directory (or via HF API if remote).
- Returns a `ModelCapabilities { supportsVision: Bool, supportsAudio: Bool, contextLength: Int?, ... }`.
- Backends consume the probe to decide which factory to use.

## SwiftLM reference

`Sources/MLXInferenceCore/ModelArchitectureProbe.swift` (132 lines) — has the structure right but uses the brittle hardcoded list. Borrow the file shape; replace the detection logic with the durable `config["vision_config"] != nil` check.

_From SwiftLM research pass._

roryford:
Tracked in umbrella #758.

roryford:
Shipped. `ModelCapabilityProbe` is on main at `Sources/BaseChatInference/Services/ModelCapabilityProbe.swift` with the durable JSON-key approach (`vision_config`/`audio_config` as objects, NSNull-safe). Tests in `Tests/BaseChatInferenceTests/ModelCapabilityProbeTests.swift` cover vision-only / audio-only / plain LLM / malformed JSON / missing files / preprocessor_config.

Closing as already-implemented. The probe is not yet *consumed* by backends to gate routing / UI affordances — that consumption work is tracked under #20.

PR #762, opened 2026-04-25
feat(inference): probe HF config.json for vision/audio capability before loading

roryford:
## Summary

Adds `ModelCapabilityProbe` and `ModelCapabilities` in `BaseChatInference`. The probe reads a downloaded HF model's `config.json` and reports `supportsVision` / `supportsAudio` / `contextLength` **without** loading the model. This lets backends pre-route to `LLMModelFactory` vs `VLMModelFactory` and lets the UI gate "attach image" affordances per-model.

## Approach

Detection keys off **durable JSON shape**, not a hardcoded `model_type` allowlist:

- `supportsVision = config["vision_config"] != nil`
- `supportsAudio = config["audio_config"] != nil`
- `contextLength = config["max_position_embeddings"] ?? config["n_ctx"]`

This is the deliberate divergence from SwiftLM's `ModelArchitectureProbe.swift`, which enumerates `paligemma | qwen2_vl | gemma3 | …` and rots every time a new VLM ships. The `vision_config` / `audio_config` keys are emitted by the transformers library for any multimodal architecture, so detection extends to new models for free — a test fixture uses a deliberately fake `model_type` (`made_up_vlm_2099`) to lock that contract in.

`preprocessor_config.json` is documented as reserved for a future revision (image-size / sample-rate UI hints) and intentionally not read today — config.json is authoritative for the flags we expose.

```swift
let caps = try ModelCapabilityProbe.probe(modelDirectory: dir)
if caps.supportsVision { /* show image attach button */ }
```

## Scope

- **In:** the probe + 8 unit tests (LLM, VLM, ALM, n_ctx fallback, missing context key, preprocessor coexistence, missing config, malformed config).
- **Out:** wiring `MLXBackend` / `LlamaBackend` to actually consume the probe — intentional follow-up on the umbrella, keeps this PR small enough to review on its own.

## Test plan

- [x] `swift test --filter BaseChatInferenceTests --disable-default-traits` — all 1093 inference tests pass (including the new 8).
- [x] Sabotage check: forced `supportsVision = false`, confirmed `testProbe_VLM_reportsVision` fails; restored.
- [x] Silent-catch audit (`SilentCatchAuditTest`) passes — no `try?` swallows introduced.

Refs #748, #758.

### DIFF
--- a/Sources/BaseChatInference/Services/ModelCapabilityProbe.swift
+++ b/Sources/BaseChatInference/Services/ModelCapabilityProbe.swift
@@ -9,10 +9,14 @@ import Foundation
 /// are detected without a code change.
 public struct ModelCapabilities: Sendable, Equatable {
     /// True when `config.json` contains a top-level `vision_config` object.
+    /// An explicit JSON `null` does not count — the key must map to an object.
     public let supportsVision: Bool
     /// True when `config.json` contains a top-level `audio_config` object.
+    /// An explicit JSON `null` does not count — the key must map to an object.
     public let supportsAudio: Bool
-    /// `max_position_embeddings` (preferred) or `n_ctx` if present, else `nil`.
+    /// First non-nil of `max_position_embeddings`, `n_ctx`,
+    /// `text_config.max_position_embeddings`, or `text_config.n_ctx`.
+    /// `nil` when none of those keys are present (e.g. embedding-only models).
     public let contextLength: Int?
 
     public init(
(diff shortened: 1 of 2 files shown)


roryford (commented):
**Verdict**: approve with fixes pushed

**Pushed**

- `189998f` fix(inference): treat null/non-object multimodal keys as absent in probe
  - `vision_config: null` (NSNull after JSONSerialization) was satisfying `!= nil`, falsely flagging vision support. Switched to `is [String: Any]` so only an actual object counts. Same for `audio_config`. Verified via a quick `swift -e` repro that NSNull is `!= nil`.
  - Added a `text_config.max_position_embeddings` (and `text_config.n_ctx`) fallback. PaliGemma- and LLaVA-NeXT-style VLMs nest their language config there and don't always expose `max_position_embeddings` at the top level — without the fallback, `contextLength` would be `nil` for some of the most common VLMs the probe is meant to support. Top-level still wins when both are set.
  - Added 4 tests: explicit-null, scalar-typed (defensive: malformed-but-parseable), nested `text_config`, and top-level-wins precedence.

**Flagged not fixed**

- *None.* The PR is genuinely small and self-contained — `preprocessor_config.json` is documented as future scope, the network-related concerns in the review prompt don't apply (this is purely an on-disk read; nothing fetches `huggingface.co`), and trait/module placement is correct (`BaseChatInference`, no ML deps).
- One nit not worth a follow-up: `errorDescription` says "is not a JSON object" for `invalidConfigJSON`, which also fires for actual JSON-parse errors thrown by `JSONSerialization`. The thrown `Data(contentsOf:)` / `JSONSerialization` errors propagate as their own native errors, so this is fine — just worth knowing if a future caller tries to render a uniform error.

**Tested**

- `swift test --filter ModelCapabilityProbeTests --disable-default-traits` — 12/12 pass (8 original + 4 new).
- `swift test --filter BaseChatInferenceTests --disable-default-traits` — 1097/1097 pass, no regressions, no SilentCatchAuditTest hits.

The probe is a good shape: keying off durable JSON structure rather than a `model_type` allowlist is the right call, and the fixture using a fake `made_up_vlm_2099` model_type is a great regression guard.

Issue #441, opened 2026-04-18
docs(example): Action Extension recipe — 'summarize selection' from any app


roryford:
## Motivation

Companion to the Share Extension recipe. Action Extensions let users invoke a BCK-powered capability on selected text from any app (iOS) or via Services menu (macOS). Demonstrating the pattern unblocks consumers building keyboard / system-wide text tools.

## Proposal

Add to the demo app:

- Action Extension target (iOS) and Services provider (macOS)
- One canonical action: "summarize selection"
- Same App Group handoff pattern as the Share Extension recipe
- Document the macOS shortcut key path

## Depends on

- Share Extension recipe (companion issue) — shares the handoff mechanism
- `ChatViewModel.ingestPendingPayload` API (companion issue)

## Acceptance criteria

- Works on both iOS (Action Extension) and macOS (Services)
- Returns result to the invoking app where possible, otherwise opens host app with result

roryford:
Folded into #440 — both demo extensions share the App Group handoff and the PendingSharePayload plumbing; one demo PR can ship both targets. #442 (ingestPendingPayload API) remains an independent prerequisite. Closing as duplicate.

PR #763, opened 2026-04-25
feat(ui): ChatViewModel.ingestPendingPayload API for extension-driven sessions

### LINKED ISSUE #442: feat(ui): ChatViewModel.ingestPendingPayload API for extension-driven sessions
## Motivation

Share / Action / URL-scheme / App Intent entry points all need a way to hand a payload to `ChatViewModel` and have it open a new session (or append to active) cleanly. Without a public API, host apps poke at internal state.

## Proposal

Add to `ChatViewModel`:

\`\`\`swift
public func ingestPendingPayload(_ payload: PendingPayload, intent: IngestionIntent) async
\`\`\`

- `PendingPayload` — `.text(String)`, `.url(URL)`, `.image(Data)`, `.file(URL)`
- `IngestionIntent` — `.newSession(preset: IngestionPreset?)`, `.appendToActive`, `.draft`

Handles session creation, model selection, message seeding, and any configured auto-run.

## Acceptance criteria

- Public API on `ChatViewModel`
- Does not widen `inferenceService` visibility
- Used by Share and Action Extension recipes (companion issues)
- Tested with in-memory SwiftData + `MockInferenceBackend`

roryford:
## Summary

Adds a public seam on `ChatViewModel` so Share Extensions, Action Extensions, and AppIntents can hand off content without poking at internal state. The new API composes the existing public surface — `inferenceService` stays internal per CLAUDE.md.

```swift
public enum PendingPayload: Sendable {
    case text(String)
    case url(URL)
    case image(Data, mimeType: String = "image/png")
    case file(URL)
}

public struct IngestionPreset: Sendable {
    public let modelID: String?           // matches ModelInfo.id (UUID string) or .name
    public let systemPrompt: String?
    public let temperature: Float?
    public let topP: Float?
    public let repeatPenalty: Float?
}

public enum IngestionIntent: Sendable {
    case newSession(preset: IngestionPreset?)
    case appendToActive
    case draft
}

extension ChatViewModel {
    @MainActor
    public func ingestPendingPayload(_ payload: PendingPayload, intent: IngestionIntent) async
}
```

## Behaviour by intent

- **`.newSession`** — inserts a fresh `ChatSessionRecord` via the configured persistence provider, switches to it, applies the preset (model match-by-UUID-then-name, system prompt, sampler), seeds the payload as the next user message, and runs `sendMessage()` when a model is loaded. When no model is loaded the payload remains as `inputText` — matching the compose-bar contract.
- **`.appendToActive`** — concatenates the payload onto the current `inputText` (newline-separated when the draft is non-empty). No persistence write, no send.
- **`.draft`** — replaces `inputText` with the payload. Works without persistence so hosts can wire it up before the SwiftData container opens.

URL payloads serialise to their absolute string. File payloads render as their path (no `.file` `MessagePart` exists yet — issue #441 will introduce one). Image payloads ride as a `MessagePart.image` attachment alongside the user message; for `.draft` the input text is cleared since attachment-only drafts are not sendable through the current compose-bar contract.

## Test plan

- [x] New integration suite `ChatViewModelIngestPendingPayloadTests` covers each requested cell — text-newSession (with and without preset / model loaded), url-appendToActive (empty + populated draft), image-draft, file-newSession, plus persistence-not-configured no-op paths.
- [x] Sabotage check on the `appendToActive` path confirmed both URL append tests fail when the implementation is broken.
- [x] `swift test --filter BaseChatUITests --disable-default-traits` — 488 tests pass.
- [x] `swift test --filter BaseChatCoreTests --disable-default-traits` — 214 tests pass (4 skipped).
- [x] `swift test --filter BaseChatInferenceTests --disable-default-traits` — 1085 tests pass.

Closes #442.

### DIFF
--- a/Sources/BaseChatUI/ViewModels/ChatViewModel+IngestPendingPayload.swift
+++ b/Sources/BaseChatUI/ViewModels/ChatViewModel+IngestPendingPayload.swift
@@ -1,0 +1,335 @@
+import Foundation
+import BaseChatCore
+import BaseChatInference
+
+// MARK: - PendingPayload
+
+/// Content handed to ``ChatViewModel/ingestPendingPayload(_:intent:)``
+/// from a host-app entry point such as a Share Extension, Action
+/// Extension, or App Intent.
+///
+/// Each case maps onto the message-part shape the chat surface already
+/// understands so the ingest path does not need a new wire format. Apps
+/// that already vend ``InboundPayload`` to ``ChatViewModel/ingest(_:)``
+/// continue to work — this API is the public seam for richer entry
+/// points that need to control session lifecycle (new vs. append vs.
+/// draft) without poking at internal state.
+public enum PendingPayload: Sendable {
+
+    /// Plain text — used as the user message body.
+    case text(String)
+
+    /// A URL — serialised into the user message body as its absolute
+    /// string. Hosts that want to attach link metadata should resolve
+    /// the URL themselves and pass the resulting prose as ``text``.
+    case url(URL)
+
+    /// An image, carried verbatim as a ``MessagePart/image`` part.
+    ///
+    /// `mimeType` defaults to `"image/png"` since most extension
+    /// pasteboards expose PNG; pass an explicit value for JPEG, HEIC,
+    /// etc.
+    case image(Data, mimeType: String = "image/png")
+
+    /// A file URL pointing at content the host has already staged
+    /// inside the app group. Today the file's path is rendered into
+    /// the user message body as text — this matches the existing
+    /// ``MessagePart`` surface, which has no file case yet. Image
+    /// files should use ``image(_:mimeType:)`` so they ride as image
+    /// parts.
+    case file(URL)
+}
+
+// MARK: - IngestionPreset
+
+/// Optional configuration applied when ``IngestionIntent/newSession``
+/// creates a fresh session.
+///
+/// Every field is optional — the view model only mutates the matching
+/// state when a value is provided, so partial presets layer cleanly on
+/// top of whatever the user already has selected. Hosts that don't
+/// need any preset can pass `nil` directly to
+/// ``IngestionIntent/newSession(preset:)``.
+///
+/// ## Model selection
+///
+/// ``modelID`` matches against ``ChatViewModel/availableModels`` by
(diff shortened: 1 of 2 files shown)


roryford (commented):
**Verdict:** approve with fixes pushed.

**Pushed commits**
- `c2c634c` fix(ui): persist IngestionPreset sampler params onto session record

**Fix detail**

`applyPreset()` only mutated live `ChatViewModel` state — `temperature`, `topP`, `repeatPenalty`, and `selectedModelID` never reached the `ChatSession` row. The PR description promises the preset "applies (model match-by-UUID-then-name, system prompt, sampler)", so callers reasonably expect a relaunch to see the preset values. Without the fix, only `systemPrompt` survived because it was baked into the `ChatSessionRecord` initializer at insertion time. The fix calls `saveSettingsToSession()` after `applyPreset` and the existing `test_textPayload_newSession_appliesPreset` is now strengthened to assert `activeSession.temperature/topP/repeatPenalty` instead of only the VM state. Sabotage-checked locally — assertions fail without the persistence call.

**Flagged not fixed**

- `.image` × `.draft` doc says "set the input draft to the payload's text" but image payloads carry no text — the draft is cleared. The test comment explains this; the *user-facing* doc on `IngestionIntent.draft` does not. Minor — adjust prose if you want.
- `.image` × `.appendToActive` is a silent no-op (empty body, early return). Likely fine, but undocumented.
- Idempotency: `.newSession` always inserts a fresh row. Calling twice with the same payload creates two sessions. Consistent with `ChatViewModel.ingest(_:)` and the PR's stated semantics, but worth a one-line callout in the doc.

**Service-sharing memo:** complied — `inferenceService` stays internal, the new API composes existing public surface only.

**Concurrency:** clean — fully `@MainActor`, no `Task.detached`, all I/O hops through the existing main-actor persistence path.

**`try?` audit:** no `try?` in `Sources/`. Test setup uses `try?` for fixture creation; SilentCatchAuditTest only scans `Sources/`, so this is fine.

**Tested**
- `swift test --filter BaseChatUITests --disable-default-traits` — 488 pass
- `swift test --filter BaseChatCoreTests --disable-default-traits` — 214 pass (4 skipped)
- New `ChatViewModelIngestPendingPayloadTests` — 9/9 pass

PR #764, opened 2026-04-25
feat(inference): pause generation when ProcessInfo.thermalState reaches .critical

### LINKED ISSUE #745: feat(inference): pause generation when ProcessInfo.thermalState reaches .critical
## Goal

Observe `ProcessInfo.thermalState` and pause active generation when it reaches `.critical`. Resume when state drops back to `.serious` or below. Prevents thermal throttling from manifesting as "the app got really slow" when it should manifest as "we backed off until the device cooled down."

## Scope

- Observer registered on backend init; deregistered on tear-down.
- On `.critical`: pause the current generation token loop (cooperative — check between tokens, no force-kill).
- On state drop: resume.
- Optional UI signal via the inference event stream so clients can show a "device throttling" message.
- Roughly ~30 LOC.

## SwiftLM reference

`Sources/MLXInferenceCore/InferenceEngine.swift:222-235` — observes thermal state via `NotificationCenter.default.publisher(for: ProcessInfo.thermalStateDidChangeNotification)` and a polled check between token emissions.

Pairs with the background-unload work in #114.

_From SwiftLM research pass._

roryford:
## Summary

Closes #745. Part of umbrella #758.

Wires a cooperative thermal gate into `GenerationCoordinator.runToolDispatchLoop` so that under `.critical` thermal pressure the per-token loop parks on a 2-second re-check instead of letting the device thermally throttle generation into a slow crawl. Resumption is automatic when state drops to `.serious` or below, or when the surrounding task is cancelled.

- New `GenerationEvent.diagnosticThrottle(reason:)` case so UI surfaces can render a "device throttling — paused" hint. Emitted **once per pause cycle** (on entry, before the first sleep) — not on every re-check tick.
- New `thermalSleep` DI seam alongside the existing `thermalStateProvider` seam so unit tests can drive the pause loop deterministically without burning real wall-clock time. Production default is `Task.sleep(for:)`.
- One unit test (`test_perTokenLoop_criticalThermal_pausesAndEmitsSingleThrottleEvent`) drives the loop through `.critical → .critical → .nominal`, asserts exactly one throttle event was emitted, and counts both the sleep invocations and provider polls to prove the loop didn't bail on the first re-check. Sabotage check: deleting the `await pauseWhileThermalCritical(...)` call drops every assertion to zero.
- Exhaustive `switch` over `GenerationEvent` in `GenerationStreamConsumer`, `ScenarioRunner`, `EventRecorder`, and `ToolCallContractTests` gains matching cases. Existing `default`-fallthrough switches need no change.

The check fires on every event (one provider read per event when temperature is fine — cheap), so the existing `.background`-priority pre-flight thermal drop in `drainQueue` is unchanged and complementary.

## Release Note

`GenerationCoordinator` now pauses active generation when `ProcessInfo.thermalState` reaches `.critical` and resumes when it drops to `.serious` or below, instead of letting the device thermally throttle the per-token loop. UI surfaces can subscribe to the new `GenerationEvent.diagnosticThrottle(reason:)` event to render a "device throttling — paused" badge while the loop is parked. Pure-additive: existing switches over `GenerationEvent` that use `default:` need no change; the four exhaustive switches in this repo were updated in the same PR.

## Test plan

- [x] `swift test --filter BaseChatInferenceTests --disable-default-traits --skip-update` — 1086 tests, 0 failures
- [x] `swift test --filter BaseChatCoreTests --disable-default-traits --skip-update`
- [x] `swift test --filter BaseChatInferenceSwiftTestingTests --disable-default-traits --skip-update`
- [x] `swift test --filter BaseChatUITests --disable-default-traits --skip-update`
- [x] `swift test --filter BaseChatBackendsTests --disable-default-traits --skip-update`
- [x] `swift test --filter BaseChatTestSupportTests --disable-default-traits --skip-update`
- [x] Sabotage check on the new test — deleting the `await pauseWhileThermalCritical(...)` call fails all five assertions on the new test as expected

### DIFF
--- a/Sources/BaseChatFuzz/EventRecorder.swift
+++ b/Sources/BaseChatFuzz/EventRecorder.swift
@@ -81,6 +81,8 @@ public struct EventRecorder: Sendable {
                     events.append(.init(t: t, kind: "toolLoopLimitReached", v: "\(iterations)"))
                 case .kvCacheReuse(let tokens):
                     events.append(.init(t: t, kind: "kvCacheReuse", v: "\(tokens)"))
+                case .diagnosticThrottle(let reason):
+                    events.append(.init(t: t, kind: "diagnosticThrottle", v: reason))
                 }
                 memoryTick()
             }
--- a/Sources/BaseChatInference/Models/GenerationEvent.swift
+++ b/Sources/BaseChatInference/Models/GenerationEvent.swift
@@ -45,4 +45,13 @@ public enum GenerationEvent: Sendable, Equatable {
     /// from the previous turn. `promptTokensReused` is the number of prompt tokens
     /// whose KV state was preserved, saving their re-decode cost.
     case kvCacheReuse(promptTokensReused: Int)
+
+    /// Emitted by the orchestrator when generation has been paused for a
+    /// runtime-side condition (e.g. `ProcessInfo.thermalState == .critical`).
+    ///
+    /// Fired exactly once per pause cycle — on entry into the wait loop, not
+    /// on every re-check tick. UI surfaces can use this to show a "device
+    /// throttling — paused" hint while the loop blocks between tokens.
+    /// `reason` is a short, human-readable string the UI may display verbatim.
+    case diagnosticThrottle(reason: String)
 }
--- a/Sources/BaseChatInference/Services/GenerationCoordinator.swift
+++ b/Sources/BaseChatInference/Services/GenerationCoordinator.swift
@@ -30,6 +30,20 @@ final class GenerationCoordinator {
     /// non-isolated so it is safe under Swift 6 strict concurrency.
     private let thermalStateProvider: @Sendable () -> ProcessInfo.ThermalState
 
+    /// Sleep hook used by the per-token thermal-pause loop. Defaults to
+    /// `Task.sleep(for:)`. Tests override this to skip the real 2-second
+    /// re-check delay and to count how many times the loop slept.
+    ///
+    /// Throws `CancellationError` when the surrounding task is cancelled —
+    /// the caller propagates that to abort the wait loop alongside a
+    /// regular state transition.
+    private let thermalSleep: @Sendable (Duration) async throws -> Void
+
+    /// Re-check delay between thermal polls when generation is paused.
+    /// Pulled out so the test seam injects only the sleep behaviour, not
+    /// the cadence — keeps the production cadence in production code.
+    private static let thermalRecheckInterval: Duration = .seconds(2)
+
     /// Optional registry used to dispatch model-emitted ``ToolCall`` events.
     ///
     /// Stored here in wave 1 so the coordinator's init surface is stable for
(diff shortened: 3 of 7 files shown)


roryford (commented):
## Verdict: approve with minor flag

The implementation is well-structured and addresses the review checklist correctly. The polling-based design (read-on-each-event + cooperative re-check loop) sidesteps the `thermalStateDidChangeNotification` observer-leak class of bugs entirely, and both `thermalStateProvider` and `thermalSleep` are dependency-injected so the pause path is fully testable without real thermal pressure or wall-clock delays.

### Pushed commits
- `9cef203` test(inference): add cancellation-during-thermal-pause unit test

### Review checklist results

| # | Concern | Result |
|---|---------|--------|
| 1 | Observer lifecycle / leak | N/A — polling, no `NotificationCenter` observer registered. |
| 2 | Resume semantics | `thermalStateProvider() != .critical` resumes; matches docstring "drops to `.serious` or below". No hysteresis but the 2s re-check cadence + entry guard makes thrash risk negligible. |
| 3 | Non-Apple-Silicon Macs | Polling is naturally resilient — `.nominal` from a stub sensor reads as no-pause. |
| 4 | Simulator / mockability | Excellent — both `thermalStateProvider` and `thermalSleep` injectable. |
| 5 | UI feedback | `.diagnosticThrottle(reason:)` event emitted exactly once on entry. Exhaustive switches updated in `GenerationStreamConsumer`, `ScenarioRunner`, `EventRecorder`, `ToolCallContractTests`. |
| 6 | Cancel-vs-pause | Verified: pause loop checks `Task.isCancelled` *and* the catch handles the `CancellationError` thrown by `Task.sleep`. New test `test_perTokenLoop_thermalPause_cancellationUnwindsCleanly` proves `stopGenerationAndWait()` returns promptly when parked mid-pause. Sabotage (replacing `while !Task.isCancelled` with `while true` + swallowing the catch) causes the test to deadlock — confirms the assertion bites. |
| 7 | CLAUDE.md compliance | Clean — async/await, `@Sendable` closures, `@Observable + @MainActor`, no `try?` / `Task.detached` / `assertionFailure` / `UserDefaults.standard`. |
| 8 | Test coverage | Pause-and-resume covered by the original test; cancellation unwind covered by the new test. Both run in <5 ms on the injected sleep hook. |

### Flagged, not fixed (non-blocking)

- **No hysteresis**: if thermal state oscillates around `.critical` at sub-2s intervals the loop will pause/resume each tick. Not a blocker — production thermal transitions don't oscillate that fast — but worth a follow-up if telemetry ever shows thrash.
- **Pause point is post-event-arrival, not back-pressure**: the gate fires after `for try await event in stream.events` yields each token, so the backend keeps decoding into its continuation buffer. This still slows the consumer-side processing but doesn't directly stop MLX/llama.cpp from generating. Acceptable given the cooperative-only contract here.

### Tested suites

- `swift test --filter BaseChatInferenceTests --disable-default-traits` — 1087 tests, 0 failures (1085 baseline + 2 new thermal tests).
- `swift test --filter BaseChatBackendsTests --disable-default-traits` — full pass.

PR #766, opened 2026-04-25
perf(mlx): yield briefly every N tokens to prevent WindowServer starvation

### LINKED ISSUE #747: perf(mlx): yield briefly every N tokens to prevent WindowServer starvation
## Goal

Insert a microsecond-scale yield in the MLX generation loop so heavy generation doesn't starve the GPU command queue WindowServer relies on. Without this, sustained MLX inference can cause UI hitches in other apps.

## Scope

In `MLXBackend`'s token loop, every ~8 tokens:

```swift
if completionTokenCount % 8 == 0 {
    try? await Task.sleep(for: .microseconds(50))
}
```

- Configurable via a `GenerationConfig.yieldEveryNTokens` knob with a sensible default.
- Only applies to MLX (Llama / Foundation / cloud don't share the issue).
- ~3 LOC.

## SwiftLM reference

`Sources/SwiftLM/Server.swift:1745-1747` — same pattern, hardcoded at every-8-tokens / 50μs. SwiftLM's commentary notes this was added after observing UI hitches under sustained generation on Mac.

_From SwiftLM research pass._

roryford:
## Summary

- Insert a 50µs cooperative yield every N tokens inside `MLXBackend`'s generation loop so sustained MLX inference doesn't starve WindowServer's GPU command queue (and hitch other Mac apps). Mirrors the pattern in `SwiftLM/Server.swift`.
- Plumb a `GenerationConfig.yieldEveryNTokens` knob (default 8; `0` disables). Only `MLXBackend` honours it — Llama / Foundation / cloud backends don't share the contention.
- Add `MLXBackend._yieldHookForTesting` test seam so the unit test counts yield invocations deterministically without timing assertions.

Closes #747.

## Test plan

- [x] `swift test --filter BaseChatInferenceTests --disable-default-traits` (no `GenerationConfig` regression; new field round-trips through `Codable`).
- [x] `swift test --filter BaseChatBackendsTests --disable-default-traits` (CI path — 75 tests pass).
- [x] `swift test --filter BaseChatBackendsTests --traits MLX` on Apple Silicon (118 tests pass; new test fires).
- [x] `swift test --filter BaseChatCoreTests / BaseChatUITests / BaseChatTestSupportTests / BaseChatInferenceSwiftTestingTests --disable-default-traits`.
- [x] Sabotage check: changing the modulo to `% (yieldEvery + 1)` fails the new test.

### DIFF
--- a/Sources/BaseChatBackends/MLXBackend.swift
+++ b/Sources/BaseChatBackends/MLXBackend.swift
@@ -103,6 +103,15 @@ public final class MLXBackend: InferenceBackend, @unchecked Sendable {
     /// Defaults to `.auto`, which picks a sensible value based on device RAM.
     public let cachePolicy: MLXCachePolicy
 
+    // MARK: - Test seams
+
+    /// Invoked in place of `Task.sleep(for: .microseconds(50))` at every
+    /// `yieldEveryNTokens`-th token during generation. Tests use this to count
+    /// yield occurrences deterministically without timing assertions.
+    ///
+    /// `nil` in production — the real microsecond sleep runs instead.
+    nonisolated(unsafe) static var _yieldHookForTesting: (@Sendable () async -> Void)?
+
     // MARK: - Init
 
     public init(cachePolicy: MLXCachePolicy = .auto) {
@@ -422,6 +431,13 @@ public final class MLXBackend: InferenceBackend, @unchecked Sendable {
                     messages: messages,
                     parameters: generateConfig
                 )
+                // Inserts a brief cooperative yield every N tokens so sustained MLX
+                // generation doesn't starve WindowServer's GPU command queue and
+                // cause UI hitches in other apps. Mirrors the pattern in SwiftLM's
+                // `Server.swift` (50µs sleep every 8 tokens). Configurable via
+                // `GenerationConfig.yieldEveryNTokens`; `0` disables the yield.
+                let yieldEvery = config.yieldEveryNTokens
+                var completionTokenCount = 0
                 outer: for await generation in mlxStream {
                     if Task.isCancelled { break }
                     if let text = generation.chunk {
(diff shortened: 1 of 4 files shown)


roryford (commented):
## Verdict — approve with one minor follow-up

The yield mechanism is correct: `Task.sleep(for: .microseconds(50))` does briefly release the cooperative thread (and indirectly the GPU command queue), which is exactly what `Task.yield()` would *not* do. 50µs at cadence 8 over a 1000-token generation is ~6.25ms total — well within budget.

### Reviewed against the PR criteria

1. **Correctness of the yield** — right primitive (sleep, not yield). Comment in source explicitly cites SwiftLM's `Server.swift` precedent.
2. **N tuning** — `yieldEveryNTokens` is plumbed through `GenerationConfig` (default 8, `0` disables). Configurable. No measured benchmark in the PR description, but the cost is bounded and the knob is exposed — acceptable.
3. **Cancellation propagation** — `try?` swallows `CancellationError` from the sleep, then the next `for await` iteration's `Task.isCancelled` check breaks the loop. Allowlist entry in `silent_catch_allowlist.txt` documents the rationale. Clean.
4. **Throughput regression test** — none. Flagged but not blocking; cost is theoretically bounded.
5. **Concurrency / Swift 6** — clean. `nonisolated(unsafe)` test seam is appropriate. `try?` is allowlisted.
6. **Trait gating** — test file is gated under `#if MLX`; uses the existing `MockMLXModelContainer` so it runs without Metal. CI default-traits run executes 75 tests with no MLX dependency.
7. **CLAUDE.md compliance** — comments explain *why* (WindowServer starvation, SwiftLM precedent). No defensive guards, no `Task.detached`.

### Pushed commit

- `38c2d23 test(mlx): cover cancellation-during-yield termination path` — adds a deterministic test that cancels generation from inside the yield hook and asserts the stream terminates cleanly without surfacing `CancellationError` and without draining all 32 mock tokens. Closes the gap noted in review item 3.

### Flagged not fixed

- **No `measure { }` baseline.** The 50µs sleep at cadence 8 is back-of-envelope reasonable, but a `BaseChatE2ETests` measure on real hardware comparing throughput with/without the yield would lock in the perf budget and catch future regressions. Not blocking; worth a follow-up issue.

### Tested suites

- `swift test --filter BaseChatBackendsTests --disable-default-traits` — 75 passed, 1 skipped (CI parity).
- `swift test --filter BaseChatBackendsTests --traits MLX` — both new tests pass deterministically (no Metal / model load required).
- `swift test --filter BaseChatInferenceTests --disable-default-traits` — 1085 passed including `SilentCatchAuditTest` with the new allowlist entry.

Issue #485, opened 2026-04-18
feat(ui): update MessageBubbleView accessibility labels for thinking parts


roryford:
## Summary

`MessageBubbleView.accessibilityLabel(for:)` currently returns `"<Role> said: <content>"` using `message.content`, which only includes visible text. Messages with thinking parts need an updated accessibility label so VoiceOver users know reasoning content is present.

## Background

`message.content` correctly excludes `.thinking` parts (it joins only `textContent` results). This means the accessibility label for a message that has both reasoning and visible text will skip the reasoning entirely — VoiceOver users get no indication a thinking block exists.

## Acceptance criteria

- [ ] If a message has thinking parts, the accessibility label announces their presence (e.g. "Assistant said: <visible text>. Includes reasoning.")
- [ ] The reasoning content itself is accessible via the `ThinkingBlockView` disclosure group (not read inline)
- [ ] `ThinkingBlockView` has an appropriate `accessibilityLabel` and `accessibilityHint` (e.g. "Reasoning, collapsed. Double-tap to expand.")
- [ ] Existing `accessibilityLabel` unit tests updated to cover the thinking-present case
- [ ] VoiceOver navigation order: reasoning disclosure group → visible message text (not interleaved)

## Notes

- `MessageBubbleView.accessibilityLabel(for:)` is `public static` and tested — update the test alongside the implementation
- Do not read the full reasoning text inline — it can be thousands of tokens and would be unusable for VoiceOver

Issue #437, opened 2026-04-18
feat(ui): ToolInvocationView + approval surface in BaseChatUI


roryford:
## Motivation

Apps running tool-calling sessions need a consistent way to render in-flight tool calls, prompt for user approval, and display completed tool results. Every BCK consumer building agent UX will otherwise reinvent the same components.

## Depends on

- Generic `ToolCall` / `ToolResult` contract (companion issue)
- Streaming tool-call deltas (companion issue)

## Proposal

Add to `BaseChatUI`:

- `ToolInvocationView` — renders pending / running / completed / failed states
- `ToolApprovalSheet` — optional approval prompt with allow-once / always / deny
- `ToolApprovalPolicy` enum: `.auto`, `.promptEachCall`, `.promptUnknownTools`, `.deny`

Host apps inject a `ToolApprovalResolver` and an optional custom result renderer.

## Acceptance criteria

- Integrates with the ChatView streaming pipeline
- Keyboard-navigable approval sheet on macOS
- Example-app coverage showing a full approve → execute → render flow

PR #767, opened 2026-04-25
feat(ui): live streaming display for thinking tokens

### LINKED ISSUE #481: feat(ui): live streaming display for thinking tokens
## Summary

`ThinkingBlockView` currently shows a static "Thinking…" label while the model reasons. Add live streaming so the reasoning text streams in real time, giving users feedback that the model is making progress rather than appearing stuck.

## Background

`ChatViewModel+Generation` accumulates thinking tokens in `thinkingAccumulator` and only writes them to `MessagePart.thinking` on `.finalizeThinking`. A second `StreamingTokenBatcher` (mirroring the one used for visible tokens) could flush partial thinking content to the message on a separate interval — fast enough to feel live, slow enough not to thrash SwiftUI.

## Acceptance criteria

- [ ] Thinking text streams into `ThinkingBlockView` during generation (not just after `.finalizeThinking`)
- [ ] `ThinkingBlockView` renders the partial text while `isThinkingStreaming: true`
- [ ] The disclosure group is still collapsed by default after streaming completes
- [ ] A separate batcher interval controls thinking update frequency (decoupled from the visible-token batcher)
- [ ] No regression on the visible-token streaming path

## Notes

- Keep the batcher interval configurable on `ChatViewModel` (e.g. `thinkingStreamingUpdateInterval`)
- The final `MessagePart.thinking` value written at `.finalizeThinking` is authoritative — intermediate flushes are display-only
- Consider whether to show a word/token count while streaming ("Reasoning… 142 tokens")

roryford:
## Summary

Replaces the static "Thinking…" placeholder with a live preview of streamed reasoning text, addressing #481.

- Adds a second `StreamingTokenBatcher` dedicated to thinking text — flushes partial content into the message's `.thinking` `MessagePart` on a configurable cadence (`thinkingStreamingUpdateInterval` / `thinkingStreamingBatchCharacterLimit`, both default to the visible-token defaults: 33 ms / 128 chars).
- `.finalizeThinking` still writes the authoritative final text after draining the batcher; intermediate flushes are display-only as the issue prescribes.
- `ThinkingBlockView` now renders a collapsed disclosure group with a `"Thinking… <inline preview>"` label while reasoning is in flight, then switches to the finalized "Reasoning" disclosure (still collapsed by default).
- A transient `Set<UUID>` on `ChatViewModel` (`messageIDsWithStreamingThinking`) tracks which message has an active reasoning stream, populated from a new `onMarkThinkingStreaming` seam on `GenerationCoordinator`.
- Fixes a latent bug uncovered by the new test: `msg.content += batch` clobbered sibling `.thinking` / tool parts (the `content` setter replaces `contentParts` wholesale). Visible-token appends now mutate the trailing `.text` part directly.

## Test plan

- [x] New `Tests/BaseChatUITests/ThinkingStreamingTests.swift` streams 5 `.thinkingToken` events through `MockInferenceBackend` and asserts the message's thinking part is mutated more than once before `.finalizeThinking` (sabotage check confirmed: disabling the partial-flush block drops distinct pre-finalize states from N to 1).
- [x] `swift test --filter BaseChatUITests --disable-default-traits --skip-update` — 481 tests, 0 failures.
- [x] `swift test --filter BaseChatCoreTests --disable-default-traits --skip-update` — 214 tests passed.
- [x] `swift test --filter BaseChatInferenceTests --disable-default-traits --skip-update` — 1085 tests passed.
- [x] `swift test --filter BaseChatInferenceSwiftTestingTests --disable-default-traits --skip-update` — 69 tests passed.
- [x] `swift test --filter BaseChatBackendsTests --disable-default-traits --skip-update` — passed.
- [x] `swift test --filter BaseChatTestSupportTests --disable-default-traits --skip-update` — 13 tests passed.

Closes #481.

### DIFF
--- a/Sources/BaseChatUI/ViewModels/ChatViewModel.swift
+++ b/Sources/BaseChatUI/ViewModels/ChatViewModel.swift
@@ -297,6 +297,27 @@ public final class ChatViewModel {
     /// Maximum characters to buffer before forcing a UI flush during streaming.
     public var streamingBatchCharacterLimit: Int = 128
 
+    /// Minimum interval between batched UI updates for streaming reasoning text.
+    ///
+    /// Decoupled from ``streamingUpdateInterval`` so apps can tune the visible-
+    /// token cadence (which drives the primary message body) independently
+    /// from reasoning preview cadence (which drives the small inline label
+    /// inside the thinking disclosure group). Defaults match the visible-token
+    /// batcher.
+    public var thinkingStreamingUpdateInterval: Duration = .milliseconds(33)
+    /// Maximum characters to buffer before forcing a UI flush for streaming reasoning text.
+    public var thinkingStreamingBatchCharacterLimit: Int = 128
+
+    /// Message IDs whose reasoning block is still being streamed.
+    ///
+    /// Populated when the first ``GenerationStreamConsumer/Action/appendThinkingText``
+    /// fragment lands and cleared on ``GenerationStreamConsumer/Action/finalizeThinking``
+    /// (or stream termination). Views read this to decide whether to render
+    /// a partial-preview affordance versus the finalized disclosure group.
+    /// Internal so views in the same module can read it without widening
+    /// the public surface.
+    var messageIDsWithStreamingThinking: Set<UUID> = []
+
     /// Whether to automatically stop generation when repetitive looping is detected.
     /// Defaults to `true`. Disable for apps that handle loop detection themselves.
     public var loopDetectionEnabled: Bool = true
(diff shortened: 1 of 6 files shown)


roryford (commented):
## Verdict: approve with two follow-up commits pushed

Live-streaming the reasoning preview is implemented well: the new `thinkingBatcher` mirrors the visible-token batcher (same 33 ms / 128 char defaults, configurable independently), the `messageIDsWithStreamingThinking` set is internal-only (no public surface widening), and the latent `msg.content += batch` bug that clobbered sibling `.thinking` parts is correctly fixed via `appendVisibleText` / `writeThinkingPartialText`. Cross-chunk marker boundaries are handled by the pre-existing `ThinkingParser` (not touched here), so the classic "chunk ends mid-`<think`" case is already covered upstream of this PR.

### Pushed commits

- `d6571e6` fix(ui): drop accessibilityValue on streaming thinking disclosure
  The streaming-state DisclosureGroup set `.accessibilityValue(text)` where `text` mutates every ~33 ms — VoiceOver re-announces the value of the focused element on each batched flush, flooding assistive tech with partial reasoning the user did not ask to hear. The static "Reasoning in progress" label conveys state; the inner `Text` (visible only when the user expands the group) still exposes the full accumulated text for assistive reading. PR #485 set the precedent that VoiceOver should see *finalized* reasoning, not every keystroke — this restores that.

- `ba2d653` test(ui): cover appendVisibleText/writeThinkingPartialText helpers
  The PR fixes a real bug (`msg.content += batch` replaces `contentParts` wholesale, dropping sibling `.thinking` / tool parts) but the helpers had no direct unit tests. Adds three tests pinning: leading-`.thinking` preserved, new `.text` part created when none exists, repeated partial-thinking writes mutate in place. Sabotage check confirmed: replacing the helper body with the legacy line made the first two tests fail.

### Flagged, not fixed

- **Performance, low risk:** `writeThinkingPartialText` rewrites the entire `.thinking(displayed)` string on each flush, so a long reasoning stream is O(n²) in total characters processed. Batching at 128 chars / 33 ms keeps the constant small and SwiftUI's `Text` diffing absorbs the rest, but if a model emits a 4 KB+ reasoning chunk this could become noticeable. No evidence of a problem yet — flagging only because the architecture pre-commits to it.
- **Product judgment, not a fix:** there is no user toggle to hide the live preview. PR #437's tool-invocation pattern allows hiding; analogous treatment may be wanted, but that's a product call.

### Tested suites

- `swift test --filter BaseChatUITests --disable-default-traits` — 484 tests, 0 failures (was 481; +3 new).
- `swift test --filter BaseChatInferenceTests --disable-default-traits` — 1085 tests, 0 failures.

Issue #436, opened 2026-04-18
feat(inference): streaming tool-call deltas for partial tool invocations


roryford:
## Motivation

Models emit tool calls progressively during streaming — the tool name may arrive first, then the arguments JSON in chunks. UI needs to render a pending tool-call state and ideally parse partial arguments.

## Depends on

- Generic `ToolCall` / `ToolResult` contract (companion issue)

## Proposal

Add incremental variants to the stream event protocol:

- `toolCallStart(callId:, name:)`
- `toolCallArgumentsDelta(callId:, text:)`
- `toolCallEnd(callId:)`

The final aggregated `ToolCall` is emitted at end. Backends that only produce whole tool calls (some Ollama models) can emit start+end around a single arguments delta.

## Acceptance criteria

- Contract declared with per-backend responsibilities documented
- Partial-JSON parsing utility (best-effort, forgiving) in `BaseChatInference` or a small companion module
- `MockInferenceBackend` supports scripting delta sequences

roryford:
### SwiftLM reference (from cross-repo research)

SwiftLM's `Sources/SwiftLM/Server.swift:2527-2550` pins down the wire format for streaming tool-call deltas, OpenAI-compatible:

```json
{"choices":[{"delta":{"tool_calls":[{"index":0,"id":"call_…","function":{"name":"…","arguments":"…"}}]},"finish_reason":null}]}
```

with a final chunk carrying `finish_reason: "tool_calls"`. Notes:

- `index` is the tool-call slot (multi-tool turns produce alternating-index deltas).
- `function.arguments` is streamed as **incremental JSON string fragments** — clients reassemble; servers do not wait for valid JSON before emitting. This is the most important shape detail.
- `id` is emitted on the first delta for each index; subsequent deltas omit it.

If BCK targets OpenAI-compatible cloud parity, this is the contract to match. Worth referencing alongside the tool-calling work just shipped in #711/#712/#713.

_From SwiftLM research pass — see https://github.com/SharpAI/SwiftLM_

roryford:
Tracked in umbrella #753 (PR group A).

PR #783, opened 2026-04-25
feat(inference): represent parallel tool-call batches in GenerationEvent

### LINKED ISSUE #621: feat(inference): represent parallel tool calls in GenerationEvent and orchestrator
## Problem

`Sources/BaseChatInference/Models/GenerationEvent.swift:19` models tool calls as singular:

```swift
case toolCall(ToolCall)
```

Modern APIs emit multiple tool calls in a single assistant turn and expect parallel execution with ordered result feedback:

- Claude 4.x: `content: [{type: tool_use, id: X}, {type: tool_use, id: Y}]`
- OpenAI gpt-4o: `tool_calls: [{id: X}, {id: Y}]`
- Gemini: `functionCalls: [...]`

Emitting these as N sequential `.toolCall` events loses the "this is a batch, dispatch together" signal. The orchestrator then can't parallelize — it executes them one at a time.

## Proposal

Two options:

**A.** Add a new event: `case toolCallBatch([ToolCall])`. Singular `.toolCall` remains for single-call emissions. Orchestrator handles the batch case via `TaskGroup`.

**B.** Change `.toolCall(ToolCall)` to `.toolCall([ToolCall])` and always carry an array. Cleaner but breaking.

Preference: **A** for compatibility during stabilization; promote to B at the next major version.

Orchestrator changes (builds on #443):

- `TaskGroup`-based parallel dispatch through `ToolRegistry`
- Result ordering in next-turn history preserved via `ToolCall.id` correlation
- Partial-failure semantics: one tool erroring does not abort the others; each surfaces its own `ToolResult(isError: true, ...)`

## Acceptance

- [ ] `GenerationEvent.toolCallBatch([ToolCall])` case
- [ ] Orchestrator uses `TaskGroup` to execute a batch concurrently
- [ ] Result order in next-turn history matches batch order
- [ ] `MockInferenceBackend.scriptedToolCalls` gains a parallel-emission helper so tests can assert batch semantics
- [ ] Cloud adapters (#435) emit `.toolCallBatch` when the provider returns multiple calls in one turn
- [ ] Unit test: 3-tool batch, one throws, other two succeed, assert all three results present with correct kinds

## Related

- Depends on: ToolRegistry/ToolExecutor contract
- Informs #435 (cloud adapters), #434 (Foundation), #414 (Llama)
- Parallel calls exercise more integration seams than serial — prime target for the ToolCallValidityDetector

roryford:
## Summary
- Adds `.toolCallStart` and `.toolCallArgumentsDelta` to `GenerationEvent` for streaming tool-call deltas (#436 prerequisite).
- `ToolCallLoopOrchestrator` dispatches multi-call rounds via `withTaskGroup` when every executor opts in via the new `ToolExecutor.supportsConcurrentDispatch` (default `false` — sequential semantics preserved).
- Result order in the next-turn prompt is preserved by batch-index sort, regardless of completion order, to keep prefix-cache reuse stable.
- Two new `BackendCapabilities` flags: `streamsToolCallArguments` and `supportsParallelToolCalls`. Cloud backends flip these in the follow-up PR.

## Why this shape (post persona review)
- `case toolCall(ToolCall)` retained as the universal call-complete event — proposed `toolCallEnd` / `toolCallBatch` were rejected as redundant by all three reviewers (consumer ordering trap; backends can be wrong; orchestrator can derive the batch internally).
- Streaming deltas are opt-in per backend (`streamsToolCallArguments`) — MLX inline parser and non-streaming Ollama can't emit honest character-granularity deltas.
- Parallel dispatch is opt-in per executor (`supportsConcurrentDispatch`) — preserves `ToolRegistry` reentrancy contract for tools with shared state.

## Test plan
- Unit: parallel dispatch concurrency, batch-order result determinism, sequential fallback when an executor opts out, cancellation drops late results, streaming delta forwarding.
- Sabotage check ran (and removed) on the batch-order sort path — confirmed `test_parallelDispatch_preservesBatchOrderInResults` fails when sort goes descending.
- All six CI suites green locally with `--disable-default-traits`.

Closes #621.

### DIFF
--- a/Sources/BaseChatFuzz/EventRecorder.swift
+++ b/Sources/BaseChatFuzz/EventRecorder.swift
@@ -89,6 +89,10 @@ public struct EventRecorder: Sendable {
                     // fuzz scenarios can pin its presence/absence without
                     // affecting reasoning text accumulation.
                     events.append(.init(t: t, kind: "thinkingSignature", v: signature))
+                case .toolCallStart(let callId, let name):
+                    events.append(.init(t: t, kind: "toolCallStart", v: "\(callId):\(name)"))
+                case .toolCallArgumentsDelta(let callId, let textDelta):
+                    events.append(.init(t: t, kind: "toolCallArgumentsDelta", v: "\(callId):\(textDelta)"))
                 }
                 memoryTick()
             }
--- a/Sources/BaseChatInference/Models/GenerationEvent.swift
+++ b/Sources/BaseChatInference/Models/GenerationEvent.swift
@@ -18,6 +18,25 @@ public enum GenerationEvent: Sendable, Equatable {
     /// call and feeding a ``ToolResult`` back into the conversation.
     case toolCall(ToolCall)
 
+    /// Streaming start of a tool call. `callId` matches ``ToolCall/id`` of the
+    /// corresponding ``toolCall(_:)`` event later in this round; `name` is
+    /// final (no `.toolCallNameDelta` exists — providers emit name up front).
+    ///
+    /// Backends only emit this when
+    /// ``BackendCapabilities/streamsToolCallArguments`` is `true`. Backends
+    /// that produce whole calls (MLX inline parser, Ollama non-streaming)
+    /// skip start/delta and emit only ``toolCall(_:)``.
+    ///
+    /// Contract: `callId` is non-empty, unique within a turn, and matches
+    /// the id of the ``toolCall(_:)`` event that closes this stream.
+    case toolCallStart(callId: String, name: String)
+
+    /// JSON-arguments fragment for an in-flight call, emitted in
+    /// concatenation order. Consumers may attempt forgiving partial-JSON
+    /// parsing for progressive UI; the authoritative arguments string lands
+    /// on the final ``toolCall(_:)`` event.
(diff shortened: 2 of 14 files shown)


roryford:
Addressed Copilot review in 256610e:

- **MockInferenceBackend per-turn queue desync**: ADDRESSED. When a turn's delta sequence is non-empty it now skips popping `scriptedToolCallsPerTurn` for that turn, matching the documented precedence and preventing future-turn desync if a test ever populates both queues.
- **Streaming-deltas test assertion message**: ADDRESSED. Updated to "2 .toolCallStart + 3 .toolCallArgumentsDelta + 2 .toolCall".
- **dispatchParallel cancellation re-check inside yield loop**: DISMISSED. After `withTaskGroup` returns, all results are already collected and the yield loop has no suspension points (`continuation.yield` is sync), so `Task.isCancelled` cannot change between the existing single check and the loop's end. The cancellation test passes against this design — the in-flight check is sufficient.

`BaseChatInferenceTests` (1163), `BaseChatBackendsTests` (80), `BaseChatTestSupportTests` (13), `BaseChatInferenceSwiftTestingTests` (69) all green locally with `--disable-default-traits`.

Issue #681, opened 2026-04-24
test(ci): guardrail against libmalloc SIGABRT re-emerging in merged Core/UI/Backends filter


roryford:
## Problem

#671 merged \`BaseChatCoreTests\`, \`BaseChatUITests\`, and \`BaseChatBackendsTests\` into a single \`swift test\` invocation to cut CI wall-clock. \`BaseChatInference\` stays split into two invocations specifically because mixing its XCTest + Swift Testing harnesses in one process triggers a libmalloc double-free SIGABRT (documented in CLAUDE.md).

That caveat is inline-commented in \`.github/workflows/ci.yml\` today, but there's no programmatic guardrail. If someone later adds heavy Swift Testing (\`@Suite\` / \`@Test\`) to \`BaseChatCoreTests\`, \`BaseChatUITests\`, or \`BaseChatBackendsTests\`, the merged filter invocation could start tripping the same SIGABRT that Inference was split to avoid.

(\`BaseChatBackendsTests\` already has a small amount of Swift Testing: \`SSEExtractEventsTests\`, \`CloudErrorSanitizerTests\`. Today that hasn't tripped the issue — likely because the volume is low compared to \`BaseChatInferenceTests\` — but the risk scales with growth.)

## Fix direction

Options, roughly by effort:

1. **Audit lint (cheap).** A test that greps \`Tests/BaseChatCoreTests\`, \`Tests/BaseChatUITests\`, \`Tests/BaseChatBackendsTests\` for \`@Suite\` / \`@Test\`, counts hits, and \`XCTFail\`s if the count grows above a committed threshold. Mirrors \`SilentCatchAuditTest\`'s allowlist pattern.
2. **Watchdog in \`scripts/test.sh\` (medium).** Detect the merged-filter invocation and warn if \`grep -c \"@Suite\"\` across the included targets exceeds a threshold.
3. **Split the merge (high-cost / anti-goal).** Revert #671's merge for the target that grew. Only if (1)/(2) keep firing.

Option (1) is the lowest-effort highest-signal starting point — the audit doesn't block anyone, just nudges.

## Acceptance

- Committing a new \`@Suite\` or \`@Test\` to \`BaseChatCoreTests\`, \`BaseChatUITests\`, or \`BaseChatBackendsTests\` either trips a CI lint with a clear message pointing at this issue, OR the contributor has updated a numeric threshold / allowlist with reviewer sign-off.
[...]

PR #787, opened 2026-04-26
refactor(backends): extract ThinkingBlockManager and centralize tokenizer heuristic

roryford:
## Summary
- Adds `ThinkingBlockManager` value type (`open()` + `flushIfOpen()`); replaces the duplicated thinking state machine in `OpenAIBackend`, `ClaudeBackend`, and `OpenAIResponsesBackend`. Each parser keeps its inline event-extraction logic — the manager only owns the open-state + `.thinkingComplete` emission so the close-rule is identical across backends.
- Wraps each parse loop in `do/catch` to flush any open thinking block **before** rethrowing on parser error, so UI consumers don't hang in a thinking-only state on upstream failure.
- Routes `LlamaBackend`'s two `max(1, text.count / 4)` fallbacks through `HeuristicTokenizer.tokenCount(_:)` (promoted to `package static`) so the heuristic lives in one place across `BaseChatInference` and `BaseChatBackends`.
- No behavior change on the success path; existing thinking-block tests continue to pass.

## Test plan
- [x] `BaseChatBackendsTests` (no traits) pass — 85 XCTest + 3 swift-testing
- [x] `BaseChatBackendsTests --traits MLX,Llama` pass — 221 (incl. new `LlamaBackendTests/test_tokenCount_noVocab_matchesHeuristicTokenizer`)
- [x] `BaseChatBackendsTests --traits CloudSaaS` pass — 72 (incl. new error-path tests for Claude and Responses)
- [x] `BaseChatInferenceSwiftTestingTests` pass — 69 (HeuristicTokenizer suite)
- [x] `BaseChatCoreTests`, `BaseChatInferenceTests`, `BaseChatUITests`, `BaseChatTestSupportTests` pass
- [x] Sabotage 1: `flushIfOpen` no-op → 5 `ThinkingBlockManagerTests` and 5 `OpenAIResponsesBackendTests` (incl. error-path) fail. Reverted.
- [x] Sabotage 2: `HeuristicTokenizer.tokenCount` set to `chars/8` → `HeuristicTokenizerTests` (4 tests) fail. Reverted.
- [x] Sabotage 3: `LlamaBackend` fallback set to local `chars/8` (bypass shared helper) → new Llama parity test fails, confirming the call site is genuinely routed. Reverted.

### Notes
- The Claude error-path test lives in a new `ClaudeThinkingErrorPathTests.swift` (XCTest) rather than as a `@Test` in `CloudThinkingTokenTests.swift`, because the `SwiftTestingAuditTest` allowlist caps that file at 8 `@Suite`/`@Test` annotations (issue #681). Adding it to the existing file would have failed CI.
- `OpenAIBackend`'s `extractStreamError` always returns `nil`, so its parser cannot throw mid-stream from the SSE payload alone; no error-path test is added for that backend (the manager is still exercised by the existing reasoning tests).

Plan: `/Users/roryford/.claude/plans/pull-main-keep-looking-starry-panda.md` (Worker C)

### DIFF
--- a/Sources/BaseChatBackends/ClaudeBackend.swift
+++ b/Sources/BaseChatBackends/ClaudeBackend.swift
@@ -278,79 +278,80 @@ public final class ClaudeBackend: SSECloudBackend, TokenUsageProvider, CloudBack
     ) async throws {
         let tokenStream = SSEStreamParser.parse(bytes: bytes, limits: effectiveSSEStreamLimits)
 
-        // `thinkingOpen` flips to true the first time we see a thinking_delta.
-        // It flips back to false — and fires .thinkingComplete exactly once —
-        // the first time we see anything that clearly isn't thinking anymore.
-        var thinkingOpen = false
+        // `thinking` flips to open the first time we see a thinking_delta.
+        // It flushes — and fires .thinkingComplete exactly once — the first
+        // time we see anything that clearly isn't thinking anymore.
+        // Signature events bypass open/close entirely.
+        var thinking = ThinkingBlockManager()
 
-        func flushThinkingCompleteIfNeeded() {
-            if thinkingOpen {
-                continuation.yield(.thinkingComplete)
-                thinkingOpen = false
-            }
-        }
+        do {
+            for try await payload in tokenStream {
+                if Task.isCancelled { break }
 
-        for try await payload in tokenStream {
-            if Task.isCancelled { break }
+                let eventType = Self.parseEventType(from: payload)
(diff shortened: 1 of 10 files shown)


roryford:
Addressed Copilot review (1 inline comment): dropped the unused `Foundation` import in `ThinkingBlockManager.swift` (7f65c3e). Manager only uses `BaseChatInference` symbols. No behavior change. Backends tests pass locally (`--disable-default-traits`).

Issue #744, opened 2026-04-25
feat(server): optional BaseChatServer target exposing OpenAI-compatible HTTP endpoints


roryford:
## Goal

Add an optional executable target `BaseChatServer` that exposes any `BaseChatInference.InferenceBackend` over an OpenAI-compatible HTTP API. Lets users point Cursor, ChatGPT-clone apps, and any OpenAI-SDK client at a local Mac running BCK.

## Scope

- New target: `BaseChatServer` executable. Depends on `BaseChatInference` + `BaseChatBackends` + Hummingbird + ArgumentParser.
- Endpoints (v1):
  - `POST /v1/chat/completions` — streaming + non-streaming. Reuse the SSE chunk shape from #436.
  - `GET /v1/models` — list configured/available backends + loaded models.
  - `GET /health` — liveness.
  - `GET /metrics` — Prometheus-style. Optional, behind a flag.
- Middleware:
  - `ApiKeyMiddleware` — bearer token auth. `/health` and `/metrics` exempt.
  - `CORSMiddleware` — configurable origin, default `*` only on `--unsafe-cors`.
- Concurrency: `AsyncSemaphore(limit: parallelSlots)` gating generation. No rate limiting in v1.
- CLI: `--port`, `--api-key`, `--parallel`, `--backend mlx|llama|foundation|ollama|cloud`, `--model <id>`.

## Non-goals (v1)

- `/v1/embeddings` — defer; SwiftLM also doesn't ship this and a closed BCK issue (#684) tracked Llama embeddings.
- Rate limiting beyond the concurrency semaphore.
- `/v1/completions` (legacy non-chat).
- Multi-tenant / per-key quotas.

## SwiftLM reference

`Sources/SwiftLM/Server.swift` is the shape — but **not** the file. SwiftLM's 3,068-line `Server.swift` mixes router, middleware, prompt cache, SSE builders, and MLX glue with documented type-checker timeouts. Borrow the shape; keep BCK's target separation discipline:

- Router setup: `Server.swift:802-916`
- Streaming handler: `Server.swift:1264-1900`
- `AsyncSemaphore` actor (~30 LOC, no deps): `Server.swift:2305-2335`
- `ApiKeyMiddleware` + `CORSMiddleware`: `Server.swift:2337-2402`
- ArgumentParser CLI surface: `Server.swift:200-285`
- SSE headers — use `X-Accel-Buffering: no` to defeat reverse-proxy buffering.
[...]

Issue #753, opened 2026-04-25
track: tool calling — coordinated rollout


roryford:
Umbrella tracking issue coordinating the tool-calling rollout. **All work complete as of 2026-05-03 — closing.**

PR-to-issue mapping audited and verified: each shipping PR's title/body references the source issue, and each headline symbol exists in `Sources/`.

## Shipping order

- [x] PR-A "Cloud tool calling" — #435, #436 — Shipped: #789 (depends on #783).
- [x] PR-B "Parallel + batch event shape" — #621 — Shipped: #783.
- [x] PR-C "Orchestrator + error paths" — #443, #622, #629, #631, #628 — All shipped.
- [x] PR-D "Capability routing" — #439, #438 — Shipped: #810. (`RouterBackend.swift`, `GenerationCapabilityRequirement.swift` present.)
- [x] PR-E "Llama tool calling" — #414, #609 — Shipped: #836. (`LlamaToolCallParser.swift`, `GBNFSchemaPreValidator.swift` present.)
- [x] PR-F "Foundation tool calling" — #434 — Shipped: #812. (`FoundationBackend.supportsToolCalling = true`.) Native tool API still tracked at #713.
- [x] PR-G "Agent-loop compression" — #444 — Shipped: #811. (`TurnHistoryCompressor.swift` present.)

## Independent modules

- [x] #620 AppIntents bridge — shipped: #798.
- [x] #625 MCP source — shipped: #790. Follow-up tracked at #792.
- [x] #627 fuzz detector — shipped: #813. (`Detectors/ToolCallValidityDetector.swift` present.)

Closing.

roryford:
All child issues complete and PR mappings verified by codebase audit (symbols present in Sources/). See updated body for the full PR map.

Issue #444, opened 2026-04-18
feat(inference): turn-history compression for agent-loop transcripts


roryford:
## Motivation

Chat compression (already shipped) summarizes user-visible dialogue. Agent tool-call transcripts have a different shape — dozens of tool calls and results that matter structurally but do not need to be preserved verbatim. A distinct compression path keeps context usable through long agent sessions.

## Proposal

Add a `TurnHistoryCompressor` distinct from the existing dialogue `Compressor`:

- Triggered when approaching a context budget mid-loop
- Collapses blocks of tool-call transcripts into a compact record ("N tools called; key results: …")
- Preserves the most recent K turns verbatim
- Emits a structured record so future turns can still reference prior tool results by id

## Depends on

- Generic `ToolCall` / `ToolResult` contract (companion issue)
- Multi-turn tool-call orchestrator (companion issue) — invokes this when budget approached

## Acceptance criteria

- Compression path distinct from user-visible dialogue compression
- Round-trip tests — the resulting context still enables the model to answer reference questions about earlier tool results
- Configurable thresholds

roryford:
Tracked in umbrella #753 (PR group G).

PR #806, opened 2026-04-26
docs(plans): #744 BaseChatServer implementation plan

roryford:
## Summary

- Plan-only PR for #744 (`BaseChatServer` — OpenAI-compatible HTTP target).
- No source changes; ships `docs/plans/744-basechatserver.md`.
- Records 17 design decisions, the module split (`BaseChatServerCore` library + thin executable), wire mapping (`GenerationEvent` → OpenAI SSE chunks), security posture, test strategy, and a 6-PR rollout to `0.13.0`.
- v2 deferrals (multi-model swap, `/metrics`, `/v1/embeddings`, cloud pass-through, structured output, etc.) tracked on #807, not folded in here.

## Why a plan PR

`BaseChatServer` is greenfield (no prior server code in the repo) and the SwiftLM reference at `Server.swift` (3068 LOC, documented type-checker timeouts) is a cautionary tale. Locking the architecture before code lands lets reviewers argue with the design rather than the diff.

Three persona reviewers (architect / DevOps / QA) will respond inline on this PR.

## Open questions for review (excerpt)

1. `0.13.0` window vs slipping to `0.14.0` to close out tool calling first.
2. `BaseChatServerCore` public vs `package` visibility (embed in another exe?).
3. Drop `--parallel` from v1 CLI surface, or keep it documented?
4. `/v1/models` lists *loaded* model only, or *available* backends/models on disk?

## Test plan

- [ ] No code changes — verify the doc renders cleanly on GitHub.
- [ ] Three persona reviews land in PR comments before any implementation PR opens.
- [ ] Plan revisions amend this PR (force-push allowed) until merged.

Closes nothing yet — #744 stays open and is closed by the final implementation PR (PR 6 in the rollout).

### DIFF
--- a/docs/plans/744-basechatserver.md
+++ b/docs/plans/744-basechatserver.md
@@ -1,0 +1,211 @@
+# Plan — `BaseChatServer` (#744)
+
+OpenAI-compatible HTTP server target that exposes any `BaseChatInference.InferenceBackend` over `/v1/chat/completions`. Lets users point Cursor, Continue, and any OpenAI-SDK client at a local Mac running BaseChatKit.
+
+This is a **plan document**, not implementation. It records the decisions taken before code lands so reviewers can argue with the design, not the diff.
+
+## Status
+
+- Tracking issue: #744
+- Branch: `feat/744-server-plan`
+- Target release: `0.13.0` (headline feature)
+- Greenfield work — no prior server code in the repo (`grep -r "Hummingbird\|Vapor\|NIOHTTP" Sources/` returns nothing).
+
+## What we are building
+
+A new opt-in SwiftPM target `BaseChatServer` (executable) plus `BaseChatServerCore` (library, where the routing lives). Together they implement:
+
+- `POST /v1/chat/completions` — streaming (SSE) and non-streaming JSON. Tools, tool_choice, temperature, top_p, max_tokens, seed, stream supported.
+- `GET /v1/models` — lists the single loaded model.
+- `GET /health/live` and `GET /health/ready` — process liveness vs model-readiness.
+- Bearer-token auth (`ApiKeyMiddleware`), conservative CORS, request-cancellation on client disconnect.
+- CLI entry point with `--port`, `--bind`, `--api-key-file`, `--parallel`, `--backend mlx|llama|foundation`, `--model <id>`.
+
+Everything inference-shaped (protocols, `GenerationEvent`, cancellation, `BackendRegistrar`) already exists and is reused as-is.
+
+## Non-goals (v1)
+
+Tracked separately on the v2 umbrella issue:
+
+- `/v1/embeddings` — depends on a real `EmbeddingBackend` (#684 was closed without one shipping).
+- `/v1/completions` (legacy non-chat).
+- `/metrics` Prometheus exporter — structured `Log.*` only in v1.
+- Multi-model swap on the request `model:` field — single backend / single model per process in v1.
+- Cloud / Ollama backend pass-through (would proxy OpenAI to OpenAI; ship-ready but pointless as a v1 headline).
+- `response_format: json_schema` cross-backend — depends on #108 `StructuredOutputStrategy`.
+- `n > 1`, `logprobs`, `logit_bias`, `frequency_penalty`, `presence_penalty` — return `400 invalid_request_error` rather than silently ignoring.
+- Pre-built binary distribution / Homebrew formula.
+- Multi-tenant / per-key quotas / rate limiting beyond the parallel semaphore.
+
+## Decisions
+
+| # | Decision | Rationale |
+|---|----------|-----------|
+| 1 | **Module shape: library `BaseChatServerCore` + thin executable `BaseChatServer`.** | SwiftLM's 3,068-line `Server.swift` (with documented type-checker timeouts) is a warning. Library form lets routes be unit-tested via Hummingbird's `Application.test()` against `MockInferenceBackend` from `BaseChatTestSupport` — no socket, no flake. |
+| 2 | **HTTP framework: Hummingbird 2.x.** | Apache 2.0, NIO-native, Swift-async-first, no Vapor ORM/templating bulk. No realistic alternative — using NIOHTTP1 directly would reinvent routing. |
+| 3 | **CLI: `swift-argument-parser`.** | Apple-stack standard. Keeps CLI surface readable next to `bck-tools` / `fuzz-chat`. |
+| 4 | **New SPM trait `Server` (opt-in, not default).** | Default dependency graph stays Hummingbird-free. Mirrors the `Fuzz` / `MCPBuiltinCatalog` precedent — heavy/optional deps live behind traits. |
+| 5 | **v1 backends: `mlx`, `llama`, `foundation` only.** | `cloud` and `ollama` would proxy OpenAI shape to OpenAI shape — the value is local backends. Cloud pass-through is on the v2 umbrella. |
+| 6 | **Single backend, single loaded model per process.** | Local backends serialize through one Metal context anyway. `/v1/models` returns the one. Multi-model swap is a v2 feature with real complexity (LRU eviction, memory budget). |
+| 7 | **Tool calling is pass-through; orchestrator disabled.** | OpenAI clients dispatch tools themselves. Server emits `tool_calls` and stops. `ToolCallLoopOrchestrator` is for in-app agent loops, not the wire protocol. |
+| 8 | **Finish reason derived in the server, not added to `GenerationEvent`.** | Avoids churning the inference protocol for one consumer. Mapping: clean exit → `"stop"`; any tool-call event → `"tool_calls"`; `.toolLoopLimitReached` or max-tokens → `"length"`; error → `"error"` (custom; OpenAI-SDK clients tolerate). |
+| 9 | **Usage estimation for local backends via `TokenizerProvider`.** | Cloud backends populate `.usage(prompt:completion:)` from the wire; local backends don't. Estimating from the tokenizer is honest enough — better than omitting the field, which breaks OpenAI client billing/observability paths. Document the estimation source in the response. |
+| 10 | **Drop thinking blocks from server output for v1.** | OpenAI shape has no home for them; clients won't render them. Add `?include=thinking` extension only if a real consumer asks. |
+| 11 | **Default bind `127.0.0.1`.** | Binding `0.0.0.0` by default is a footgun on machines that move between networks. `--bind 0.0.0.0` must be explicit. |
+| 12 | **API key sourced from `--api-key-file <path>`, not `--api-key <value>`.** | Process-table leaks via `ps` are real. `--api-key` is accepted with a deprecation note (dev-only). |
+| 13 | **Default `--parallel 1`.** | Local backends serialize through one Metal context. >1 would queue inside `MLXBackend` anyway. The `AsyncSemaphore` is structural for future cloud/Ollama use. |
+| 14 | **Unsupported OpenAI fields: 400 `invalid_request_error` with the field name.** | Silent ignore is the worst possible behavior — clients spend hours tracing "why doesn't `seed` work" when the server quietly dropped it. Loud failure beats quiet drift. |
+| 15 | **Two health endpoints: `/health/live` and `/health/ready`.** | Single `/health` is ambiguous for Kubernetes / launchd / `pgrep` style supervisors. Liveness vs readiness is a 5-LOC distinction with real value. |
+| 16 | **Hand-roll narrow Codable models for `ChatCompletionRequest` / `Chunk` / `Response`.** | No shippable Swift OpenAI SDK exists that would fit. The surface we use is small (~150 LOC). `OpenAIToolEncoding` (outbound) uses `[String: Any]` and isn't reusable verbatim — extracting a shared module is a refactor we'd regret in v1. Re-converge in v2 if duplication hurts. |
+| 17 | **Use `SSEStreamLimits` from `BaseChatInference` for response-side budget caps.** | Same defensive limits the parser enforces inbound apply outbound. One bad request shouldn't stream 50MB. |
+
+## Module structure
+
+```
+Sources/
+  BaseChatServerCore/          # library — routes, models, middleware, no socket
+    Models/
+      ChatCompletion+Request.swift
+      ChatCompletion+Response.swift
+      ChatCompletion+Chunk.swift
+      OpenAIError.swift           # invalid_request_error / authentication_error / etc.
+    Routes/
+      ChatCompletionsRoute.swift  # POST /v1/chat/completions (streaming + non-streaming)
+      ModelsRoute.swift           # GET  /v1/models
+      HealthRoute.swift           # GET  /health/live, /health/ready
+    Middleware/
+      ApiKeyMiddleware.swift
+      CORSMiddleware.swift
+    Streaming/
+      SSEEncoder.swift            # data: …\n\n framer + [DONE] sentinel
+      EventMapper.swift           # GenerationEvent → ChatCompletionChunk
+      FinishReasonDeriver.swift   # stream-end → "stop" | "tool_calls" | "length"
+    Concurrency/
+      AsyncSemaphore.swift        # ~30 LOC actor (lifted shape from SwiftLM)
+    Server.swift                  # buildApplication(config:) -> Hummingbird.Application
+
+  BaseChatServer/                # executable
+    main.swift                   # @main, parses CLI, calls buildApplication, .run()
+    CLI.swift                    # ArgumentParser surface
+
+Tests/
+  BaseChatServerTests/
+    Routes/
+      ChatCompletionsRouteTests.swift
+      ChatCompletionsStreamingTests.swift
+      ChatCompletionsToolsTests.swift
+      ModelsRouteTests.swift
+      HealthRouteTests.swift
+    Middleware/
+      ApiKeyMiddlewareTests.swift
+      CORSMiddlewareTests.swift
+    Streaming/
+      SSEEncoderTests.swift
+      EventMapperTests.swift
+      FinishReasonDeriverTests.swift
+    Concurrency/
+      AsyncSemaphoreTests.swift
+    Integration/
+      ClientDisconnectCancellationTests.swift   # disconnect mid-stream → stopGeneration() fires
+      EndToEndOpenAIClientTests.swift            # real openai-python via subprocess (gated)
+```
+
+`BaseChatServerCore` depends on: `BaseChatInference`, `BaseChatBackends`, `BaseChatTestSupport` (test only), `Hummingbird`.
+`BaseChatServer` depends on: `BaseChatServerCore`, `swift-argument-parser`.
+
+The split keeps `BaseChatServerCore` testable without sockets, and the executable target stays a thin wrapper — same pattern as the `bck-tools`/`fuzz-chat` precedents but with a real library underneath because the surface justifies it.
+
+## Wire mapping
+
+`GenerationEvent` → OpenAI SSE chunks:
+
+| `GenerationEvent` case | OpenAI chunk |
+|---|---|
+| `.token(s)` | `choices[0].delta.content = s` |
+| `.toolCallStart(id, name)` | `choices[0].delta.tool_calls[idx] = {id, type: "function", function: {name}}` |
+| `.toolCallArgumentsDelta(id, frag)` | `choices[0].delta.tool_calls[idx].function.arguments = frag` |
+| `.toolCall(call)` | (assemble fallback if no streamed deltas seen for id) |
+| `.usage(p, c)` | `usage = {prompt_tokens: p, completion_tokens: c, total_tokens: p+c}` on final chunk if `stream_options.include_usage = true` |
+| `.thinkingToken(_)` / `.thinkingComplete` / `.thinkingSignature(_)` | dropped (v1) |
+| `.toolResult(_)` | dropped (server is pass-through; client owns dispatch) |
+| `.kvCacheReuse(_)` / `.diagnosticThrottle(_)` | dropped (internal signals) |
+| `.toolLoopLimitReached(_)` | derive `finish_reason: "length"` |
+| stream end clean | derive `finish_reason: "stop"` (or `"tool_calls"` if any tool-call event seen) |
+| stream error | emit `data: {"error": {…}}\n\n` then close; OpenAI clients handle this |
+
+## Cancellation
+
+The framework already has the contract: `stopGeneration()` cancels the in-flight stream and resets backend state, and structured concurrency propagates from Hummingbird's per-request task. When the client disconnects mid-stream:
+
+1. Hummingbird cancels the response task.
+2. The route handler's `for try await event in stream.events` throws `CancellationError`.
+3. `defer { backend.stopGeneration() }` fires.
+4. `MLXBackend.stopGeneration()` (or whichever) tears down state and is ready for the next request.
+
+This needs an explicit acceptance test (`ClientDisconnectCancellationTests`) — it's an easy regression and not visible until the server is heavily used.
+
+## Test strategy
+
+- **Unit (route-level):** Hummingbird `Application.test()` + `MockInferenceBackend`. Cover all OpenAI shape edge cases — empty messages, malformed tools, unsupported fields → 400, auth header missing → 401, etc. No sockets.
+- **Streaming:** `EventMapperTests` drive a synthetic `AsyncStream<GenerationEvent>` through the encoder and assert SSE bytes. Includes the parallel-tool-call case (multiple `index` lanes interleaved) since that's the part most likely to break under real load.
+- **Cancellation:** `ClientDisconnectCancellationTests` cancels the test client mid-response and asserts `stopGeneration()` was invoked on the mock.
+- **Sabotage:** Per CLAUDE.md, every assertion gets a temporary code break verifying the test fails before commit.
+- **End-to-end (gated):** `EndToEndOpenAIClientTests` shells out to `python -c "from openai import OpenAI; …"` against a real bound port, gated on `RUN_SERVER_E2E=1` and macOS. Confirms wire compatibility against the canonical client. Not in default CI.
+- **CI gate:** `BaseChatServerTests --disable-default-traits` runs in CI alongside the existing suite. No hardware required (mock backend only).
+
+## Security posture
+
+- Default bind `127.0.0.1`.
+- `--bind 0.0.0.0` requires either `--api-key-file` to be set OR `--unsafe-no-auth` to be passed explicitly. Refuse to start otherwise.
+- API key compared with `constantTimeEquals` to avoid timing oracles.
+- CORS disabled by default. `--cors-origin <origin>` accepts a single origin; `--unsafe-cors-any` is required to allow `*`.
+- Request body size capped at 1 MiB (configurable via `--max-request-bytes`). Defends against accidental large pastes.
+- Response stream capped using `SSEStreamLimits` defaults (50 MiB total, 5000 events/sec).
+- Errors funneled through `OpenAIError` enum with explicit code mapping — never leak Swift type names or stack traces.
+- No file paths, model names, or backend identifiers logged at default level — only at `--log-level debug`.
+- Dependency review: Hummingbird and ArgumentParser are both Apache 2.0, MIT-compatible. No new transitive deps that haven't been vetted.
+
+## CI
+
+Add to the pre-push checklist and CI matrix:
+
+```bash
+swift test --filter BaseChatServerCoreTests --disable-default-traits --traits Server
+```
+
+The `Server` trait is required because the test target depends on Hummingbird. If the trait is off, the target compiles to zero files and the filter no-ops — same pattern as `Ollama` / `Fuzz` today.
+
+CI line in `.github/workflows/*.yml` mirrors the existing `swift test --filter` invocations. ~30 seconds of runtime added (mostly link).
+
+## Rollout
+
+1. **PR 1 — this plan doc.** No code. Three persona reviews land here.
+2. **PR 2 — skeleton.** `Package.swift` trait + targets, empty `BaseChatServerCore`, empty executable, `BaseChatServerCoreTests` placeholder. Confirms the build graph is clean before any feature work. ~150 LOC.
+3. **PR 3 — non-streaming `/v1/chat/completions` + `/v1/models` + `/health/*` + auth/CORS middleware.** Mock backend tests only. ~600 LOC.
+4. **PR 4 — streaming SSE + `EventMapper` + cancellation.** Adds `ClientDisconnectCancellationTests`. ~400 LOC.
+5. **PR 5 — tool calling pass-through.** OpenAI client `tools=[…]` flow end-to-end against the mock. ~200 LOC.
+6. **PR 6 — CLI surface + executable target + README section.** Manual smoke against `mlx-community/Llama-3.2-3B-Instruct-4bit` and a real `openai` Python client. ~150 LOC. Lands the user-visible surface.
+7. **Release 0.13.0.** Highlight section in CHANGELOG with a 6–8 line code snippet showing `swift run BaseChatServer --backend mlx --model …` → `from openai import OpenAI; client.chat.completions.create(...)`.
+
+Each PR is independently reviewable, CI-green at every step, and revertable. PR 6 is the one that ships the feature; PRs 2–5 are infrastructure that's safe to land incrementally because nothing references it from outside the new module.
+
+## Risks and mitigations
+
+| Risk | Mitigation |
+|---|---|
+| Hummingbird API churn between minor versions | Pin to `2.x.y` exact in `Package.swift`; bump deliberately. |
+| Local-backend usage estimates diverge from cloud actuals enough to break observability tools | Document the source explicitly (`x-basechat-usage-source: tokenizer-estimate` header in non-streaming responses); revisit in v2 with a real counting hook. |
+| Client disconnect doesn't propagate cancellation in some Hummingbird configuration | Explicit acceptance test on day one. Easy to detect, painful in production. |
+| Parallel tool-call events arrive out of order across `index` lanes and break OpenAI's strict-index expectation | `EventMapperTests` covers interleaved lanes; `ToolCallLoopOrchestrator` already sorts by batch index for KV-prefix stability (#783), so the upstream ordering is correct. |
+| OpenAI SDK clients reject responses with our custom `finish_reason: "error"` | Test against the real `openai` Python client in `EndToEndOpenAIClientTests`. If the SDK rejects, fall back to `"stop"` + an `error` field on the chunk. |
+| Server target pulls Hummingbird into the default graph by accident | Trait gate enforced by `swift package resolve` failing if the default-trait build pulls server deps. CI `--disable-default-traits` invocation confirms. |
+| Type-checker timeouts (the SwiftLM cautionary tale) | Module split (`BaseChatServerCore` vs `BaseChatServer`) keeps any one file small. Hard rule: no file over 400 LOC. |
+
+## Open questions for review
+
+1. Is `0.13.0` the right release window, or should this slip to `0.14.0` to keep `0.13.0` focused on closing out the tool-calling track (#753 PR-C + #444)?
+2. Should `BaseChatServerCore` ship public-API for embedding it in another executable (e.g., a Mac menubar app that wants the same routes), or stay `package`-visible?
+3. The `--parallel` flag is structural in v1 (default 1, semaphore mostly idle). Drop it from the v1 CLI to reduce surface, or keep it documented as future-facing?
+4. Should `/v1/models` reflect *available* backends/models on disk (richer for clients) or only the *loaded* one (simpler, honest)?
+
+These are the four genuine forks where I'd like reviewer disagreement, in addition to anything else they spot.
--- a/docs/plans/744-basechatserver.md
+++ b/docs/plans/744-basechatserver.md
@@ -7,63 +7,79 @@ This is a **plan document**, not implementation. It records the decisions taken 
 ## Status
 
 - Tracking issue: #744
+- v2 deferrals umbrella: #807
 - Branch: `feat/744-server-plan`
-- Target release: `0.13.0` (headline feature)
+- Target release: `0.14.0` (`0.13.0` ships the tool-calling track close-out per #753 PR-C; the server gets its own headline release because the network surface is the more dangerous surface and deserves concentrated reviewer attention)
 - Greenfield work — no prior server code in the repo (`grep -r "Hummingbird\|Vapor\|NIOHTTP" Sources/` returns nothing).
 
 ## What we are building
 
-A new opt-in SwiftPM target `BaseChatServer` (executable) plus `BaseChatServerCore` (library, where the routing lives). Together they implement:
+A new opt-in SwiftPM target `BaseChatServer` (executable) plus `BaseChatServerCore` (library, `package`-visible, where the routing lives). Together they implement:
 
 - `POST /v1/chat/completions` — streaming (SSE) and non-streaming JSON. Tools, tool_choice, temperature, top_p, max_tokens, seed, stream supported.
 - `GET /v1/models` — lists the single loaded model.
 - `GET /health/live` and `GET /health/ready` — process liveness vs model-readiness.
 - Bearer-token auth (`ApiKeyMiddleware`), conservative CORS, request-cancellation on client disconnect.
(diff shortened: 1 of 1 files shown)


roryford:
## DevOps review

Plan is solid in shape — module split, trait gating posture, and security defaults are all right. My pushback is concentrated on **rollout cardinality, CI cost, and operability gaps** that will bite once this is running on a user's iMac at 3 AM. Net: ship it, but compress the PR rollout and fill in the operability blanks before PR 6.

### CI cost & gating

- `.github/workflows/ci.yml:164-170` already runs two `swift test` invocations (XCTest batch, then isolated Swift Testing). Adding a third for `BaseChatServerCoreTests --traits Server` is **not "~30s of mostly link"** — it's a fresh `swift test` process bring-up plus Hummingbird's NIO dep graph compiling for the first time on this runner. Realistic add: 60–90s on a cold cache, plus the SwiftPM cache key in `.github/workflows/ci.yml:72` will churn whenever `Package.resolved` changes (which this PR will). Budget honestly.
- **Fold the Server suite into the existing XCTest batch step**, not a new step. The batch at `ci.yml:166` already takes `--disable-default-traits`; pass `--traits Server` there and add `--filter BaseChatServerCoreTests`. One `swift test` invocation, one link.
- The `paths:` filter at `ci.yml:6-13` already covers `Sources/**` and `Tests/**`, so server-only PRs will trigger the full suite anyway. **Don't** add a path-scoped sub-job — too much ceremony for a target this size, and skipping the rest of the suite on server-only PRs is risky given how `BaseChatInference` is shared.
- Add a CI lint analogous to the BaseChatUI→BaseChatUIModelManagement guard at `ci.yml:105-117`: assert `Sources/BaseChatBackends/`, `Sources/BaseChatUI/`, etc. don't `import Hummingbird`. The Server trait is only effective if no default-graph target ever pulls it.

### Trait-gating proof

The plan asserts Hummingbird stays out of the default graph. **Make this falsifiable in CI**: add a one-shot step `swift package show-dependencies --disable-default-traits --format flatlist | grep -qv hummingbird || exit 1`. Cheap, catches the regression instantly. Same drum I beat for `Fuzz`/`Ollama`.

### Hummingbird dep hygiene

- Latest Hummingbird is **2.22.0 (2026-04-10)** and that release explicitly **drops Swift 6.0 support**. Current `swift-tools-version: 6.1` (`Package.swift:1`) clears it, but pin the dep with `.upToNextMinor(from: "2.22.0")` not `from: "2.0.0"` — Hummingbird has had churn within 2.x and minor-version pinning matches the conservative posture taken with `swift-syntax` at `Package.swift:62`.
- Hummingbird ships `ExistentialAny` + `InternalImportsByDefault` strict settings as of 2.22.0. Confirm `BaseChatServerCore` builds clean under those before committing to the dep — if it forces strict-imports churn across our targets, that's its own scope.
- Plan rationale says "no realistic alternative" — agree. Don't relitigate.

### Rollout cardinality (push back here)

**6 PRs is too many for ~1500 LOC.** Specifically:

- **PR 2 ("skeleton, no feature work, ~150 LOC") burns a CI run for nothing.** Per CLAUDE.md "Issue & PR hygiene" — single-file/empty-target PRs are a smell. Merge PR 2 into PR 3.
- **PR 5 ("tool-calling pass-through, ~200 LOC") and PR 6 ("CLI + executable, ~150 LOC") are too thin to justify standalone CI runs** at 10× billing. Fold PR 5 into PR 4 (streaming includes tool-call deltas anyway — that's the same `EventMapper`) and PR 6 into PR 4 or PR 5.
- Realistic shape: **3 PRs** (skeleton+non-streaming, streaming+tools+cancellation, CLI+executable+README). Saves ~3 CI cycles × ~5 min × 10× = real money and reviewer round-trips.

### Release-artifact story / Gatekeeper

`swift run BaseChatServer ...` is acceptable for v1 **only if the README is honest** about:

- First-run Gatekeeper prompt when binding to a port (the dyld loader is unsigned in a `.build/debug` binary; macOS 14+ shows a "downloaded from internet" dialog if the user `tar`s the binary across hosts).
- App Translocation: if a user moves the binary out of `.build/`, macOS may quarantine + translocate it. Document the `xattr -dr com.apple.quarantine` workaround.
- No keychain integration in v1 — `--api-key-file` reads plaintext. Say so.

If any of this is missing from the README in PR 6, it'll generate issue noise. Pre-built signed/notarized binary is correctly deferred (#807) but the v1 docs can't pretend the rough edges aren't there.

### Operability gaps to close before PR 6

The plan mentions structured `Log.*` but not:

- **Log format.** JSON-line output via `--log-format json` for prod, text for dev. This is 20 LOC and indispensable for `tail -f | jq` debugging.
- **`SIGTERM` graceful shutdown.** Hummingbird supports it natively. The plan must say in-flight streams get N seconds (default 30) to drain before forced close. Without this, `launchctl kickstart -k` will eat mid-stream completions.
- **`SIGINT` for foreground use** (Ctrl-C should drain, not abort).
- **Crash/panic destination.** No `OSLog` mention — at minimum log uncaught errors via `os_log` so users can `log show --predicate ...` after a crash.
- **Bind-port-in-use error.** Hummingbird surfaces it as a generic NIO error; map to a clear "port 8080 in use; pass --port" message.

Add a "Operability" section to the plan covering these. They're not large but they're the difference between "works on my machine" and "deployed".

### Bind defaults & TLS

- `127.0.0.1` default + `--unsafe-no-auth` gate on `0.0.0.0` is correct.
- **IPv6:** `--bind ::1` should be supported; on dual-stack macs some clients (Cursor's Electron WebView) prefer IPv6 loopback. Document. `[::]` for "all interfaces v6" similarly behind the auth gate.
- **TLS:** HTTP-only is fine for v1 *because* the default is loopback. The moment a user hits `--bind 0.0.0.0` they likely want TLS. Don't ship TLS in v1, but **document the recommended path** (Caddy or `cloudflared` reverse-proxy in front). One paragraph in the README.
- Cursor's WebView CSP rejects mixed-content over HTTPS pages — relevant only for hosted-Cursor users. Note in v2 deferrals (#807) that an `--unsafe-cors-any` user is probably going to hit this.

### Open questions — direct answers

1. **`0.13.0` window**: ship it in **`0.13.0`**. Tool-calling track (#753) and BaseChatServer are independent in source; coupling release windows means one slip drags the other. Each gets its own Highlight in the CHANGELOG.
2. **`BaseChatServerCore` visibility**: keep it **`package`-visible for v1**. Public is a one-way door — hosts that want to embed (#807 menubar wrapper) can wait until v2 when the API has shaken out. Premature `public` always costs more than promotion later.
3. **Drop `--parallel` from v1 CLI**: **drop it**. Default-1 with no observable effect is cargo-cult surface. Add it in v2 alongside cloud pass-through where it actually does work. CLAUDE.md "fewer, larger units of work" applies to flags too.
4. **`/v1/models` shape**: return **only the loaded model** in v1. Cursor and the `openai` Python SDK don't care about non-loaded backends; richer enumeration is a v2 feature behind real demand. Honest > clever.

### Minor

- Risk table mentions Hummingbird API churn — pin minor-version (above), not just major.
- "Hard rule: no file over 400 LOC" is good. Enforce with a CI check, not just plan prose.

roryford:
## Architect review

Solid plan overall — the module split, trait gating, and pass-through tool stance are the right calls and the rollout is honestly staged. My main argument is with two of the deferrals (`OpenAIToolEncoding` reuse, `finishReason` derivation) where I think v1 is the cheapest moment to do the right thing, not the most expensive.

### Module boundaries & dependency direction

- `BaseChatServerCore` depending on `BaseChatBackends` (plan line 113) is the wrong edge. The library shouldn't *need* the concrete backend types — it needs `InferenceBackend` (a protocol from `BaseChatInference`) plus `BackendRegistrar` from `BaseChatBackends`. Make the executable wire the backend in, and let `BaseChatServerCore` depend only on `BaseChatInference`. That preserves the same one-way edge `CLAUDE.md` enforces for UI ("UI never imports backends") and keeps the door open for an embedder that registers its own backend (open question 2 below). Test target can pull `BaseChatTestSupport` for `MockInferenceBackend` regardless.
- The plan does not call out a CI lint analogous to the existing UI-import lint. Worth adding one entry to the dep-graph guard so that nothing in `Sources/BaseChatUI` or `Sources/BaseChatCore` ever imports `BaseChatServerCore` (it would be tempting in a menubar host app, and it would yank Hummingbird into the default UI graph).

### Reuse vs duplication — `OpenAIToolEncoding`

I read `Sources/BaseChatBackends/OpenAIToolEncoding.swift`. Decision 16's claim that extraction is a refactor "we'd regret in v1" is overstated. Look at lines 23-108: `encodeToolDefinition`, `applyToolChoice`, `encodeToolCall`, `encodeChatCompletionsEntry` are all pure functions over BCK types, gated behind `#if CloudSaaS`, that produce `[String: Any]` payloads. The only thing that makes them un-shareable today is the `CloudSaaS` trait gate and the `[String: Any]` shape. Outbound (server → client) we want the *same* envelope — tools list, tool_calls on assistant turn, tool role with `tool_call_id`. Hand-rolling parallel Codable structs in `BaseChatServerCore` to emit the same JSON is exactly the duplication that bites later when we add a field (e.g. `function.strict`) and forget the second implementation. I'd push for a tiny `BaseChatOpenAIWire` module (or moving `OpenAIToolEncoding.swift` up to `BaseChatInference` un-trait-gated — it has no Cloud deps at all, the `#if` is just inherited from the file's neighbours) before PR 5 lands, not after. Cost: a few hours; benefit: one source of truth for the OpenAI shape.

### `GenerationEvent` mapping integrity

The wire mapping table (plan lines 122-134) treats `.toolCallStart` + `.toolCallArgumentsDelta` as the primary path and `.toolCall` as a fallback. That inverts reality for v1's three target backends:

- `MLXBackend.swift` line 62 sets `supportsToolCalling: true` but does **not** set `streamsToolCallArguments` — it emits whole `.toolCall(call)` events only.
- `LlamaBackend.swift` line 53 sets `supportsToolCalling: false`.
- `FoundationBackend.swift` line 80 sets `supportsToolCalling: false`.

So the *only* v1 backend that emits tools (MLX) takes the "fallback" path. The mapping needs to make whole-call the primary path and synthesize a `tool_calls` chunk + final chunk pair from `.toolCall(call)`, and the streamed-deltas path becomes the (currently dead) Ollama/cloud-on-v2 expansion. Worth a sentence in the plan to invert the framing — otherwise the first MLX smoke test will surface this.

### Finish-reason — extend the protocol now, not later

Decision 8 punts `.finished(reason:)` because "one consumer." That's not actually true today — `GenerationCoordinator`, `GenerationStreamConsumer`, `ToolCallLoopOrchestrator`, `ScenarioRunner`, and `EventRecorder` all already infer end-of-stream and tool-call-completion from event-shape heuristics (see grep for `toolLoopLimitReached` across `Sources/`). The server is the *fifth* consumer doing the same derivation. Adding `case finished(reason: FinishReason)` to `GenerationEvent` (with `.stop / .toolCalls / .length / .error`) is a one-PR change that simplifies four existing call sites and the new server. Doing it before PR 4 lands keeps the wire mapping clean; doing it after means the server invents a private mapping that will drift. This is exactly the moment.

### Future-proofing v1 → multi-model

Single-model is right, but two things in the v1 design subtly lock in single-model:

1. `/v1/models` returning the loaded model only is fine, but the route handler needs to be written against an `InferenceService` (or the `BackendRegistrar` set), not a captured `InferenceBackend`. Wiring `let backend: any InferenceBackend` straight into the route closure will be the migration pain.
2. Decision 13's `--parallel 1` semaphore is per-process. v2 multi-model wants per-model semaphores. Make the `AsyncSemaphore` a protocol or pass it in, don't make it a singleton in `BaseChatServerCore`.

### Risks

The "Hummingbird API churn" mitigation says "pin to `2.x.y` exact." Pinning exact transitive ranges in a *library* is hostile to consumers — pin in the executable's `Package.resolved` and use `from: "2.x.0"` in `Package.swift`. Otherwise downstream packages that also use Hummingbird will hit unsatisfiable graphs.

`SSEStreamLimits` (decision 17): worth flagging that this type today is an *inbound parser* concept (`Sources/BaseChatInference/Services/SSEStreamParser.swift`). Reusing the values is fine; reusing the same enforcement code path isn't possible — outbound caps need their own enforcement. The plan's framing is slightly misleading here.

### Open questions

1. **0.13.0 vs 0.14.0** — slip to 0.14.0. The tool-calling track (#753 PR-C, #444) is the more compelling 0.13.0 headline and shipping both in one minor risks underbaking either. The server is also the more dangerous of the two (network surface, auth, CORS) and benefits from a release where it's *the* feature reviewers focus on.
2. **`BaseChatServerCore` public vs package** — public, but only after the dep direction is fixed (above). With `BaseChatServerCore` depending on `BaseChatInference` only, it's a clean library a menubar host can embed. With it depending on `BaseChatBackends`, public ships Hummingbird+all-backends as a transitive surface and we'll regret it.
3. **`--parallel` flag** — keep, but document it as `(currently advisory; see #807)`. Removing and re-adding a CLI flag is a breaking change; leaving it documented is free.
4. **`/v1/models`: available vs loaded** — loaded only, for v1. "Available" requires the server to know about disk state and download status, which pulls in `BaseChatCore` (SwiftData) or a parallel scanner. Both are scope creep. v2 with multi-model is the right time — extend the schema then.

roryford:
## QA review

Plan is structurally sound and the test list covers the happy paths, but several adversarial edges are missing and a few decisions don't yet have an assertion that would fail if we silently regressed them. If we ship v1 exactly as written, the gaps below are the ones I'd expect to bite us first in production.

### Coverage gaps in the named test suites

- **Backpressure / slow consumers.** `SSEStreamLimits` (see `Sources/BaseChatInference/Services/SSEStreamParser.swift`) caps total bytes and event rate but does nothing about a slow reader. We need a dedicated test that holds the response body iterator without draining and asserts that the route either applies Hummingbird flow control or surfaces a write timeout — and that `stopGeneration()` still fires when the buffer fills. None of the listed suites covers this.
- **Request parsing edges.** No tests listed for: malformed JSON body, body exceeding `--max-request-bytes`, ragged UTF-8 in `messages[].content`, duplicate header keys, `Content-Type: text/plain`, gzip body. These are 400/415 paths the plan implies but doesn't pin down.
- **Degenerate inputs.** `messages: []`, system-only message with no user turn, empty string content, `tool_choice: "none"` with `tools: []`, `stream: true` with `max_tokens: 0`. Each should have a deterministic 400 or empty-stream contract.
- **`stream_options.include_usage`.** Decision 9 says local backends emit estimated usage; the wire-mapping table includes it; no test in the plan verifies it appears (or is absent when `include_usage` is unset/false). This is exactly the decision that drifts silently — `MockInferenceBackend` doesn't emit `.usage` today (`Sources/BaseChatTestSupport/MockInferenceBackend.swift`), so we'll need to extend it or use `TokenTrackingMockBackend`.
- **Mid-stream backend error.** `MidStreamErrorBackend` already exists at `Sources/BaseChatTestSupport/MidStreamErrorBackend.swift` — perfect fit for a test that asserts the route emits the `data: {"error": …}\n\n` chunk, closes cleanly, and that `stopGeneration()` was invoked. Plan doesn't list this even though the wire-mapping table promises the behavior.
- **Tool-call interleaving sub-cases.** Plan says `EventMapperTests` covers parallel lanes but doesn't enumerate: `index=1` start before `index=0` start, deltas arriving for an `index` that never had a start (malformed upstream), `.toolCall(_)` arriving without preceding deltas (MLX inline path), and the closing `.toolCall(_)` carrying different `arguments` than the concatenated deltas (which wins?). Each is a real upstream pattern.
- **Auth edges.** `ApiKeyMiddlewareTests` is named but unscoped. Cases needed: missing header (401), wrong scheme (`Basic …` instead of `Bearer …`), correct prefix with empty token, header injection (`Bearer foo\r\nX-Inject: …`), constant-time compare verified by length-mismatch and equal-length-different-byte cases. The constant-time path needs a sabotage check that swaps in `==` and confirms the timing test fails — except plan has no timing test, so the security promise of decision is unverified.
- **CORS edges.** Preflight `OPTIONS` with `Access-Control-Request-Method`, multi-origin allowlist behavior, `--unsafe-cors-any` actually emitting `*`, and `Vary: Origin` on responses when origin is echoed. None enumerated.
- **Cancellation rigor.** `ClientDisconnectCancellationTests` is one bullet for a behavior with at least five distinct race windows: disconnect between tokens, during a single token write, during tool-call argument streaming, after `[DONE]` but before drain, and the natural-completion-vs-disconnect tie. Plan should split this into a parameterized suite. Also need a test where `stopGeneration()` itself throws/hangs and assert the route doesn't deadlock.
- **`SlowMockBackend` / `ChaosBackend`** in `BaseChatTestSupport/` already give us throttle-then-resume and partial-failure shapes for free — plan should call them out so PR 4 doesn't reinvent them.

### Decisions without a failing test if regressed

| # | Decision | Test that would fail if broken? |
|---|----------|---|
| 8 | `finish_reason` derivation | `FinishReasonDeriverTests` listed — good. Add explicit cases for `.toolLoopLimitReached` → `"length"` and error → `"error"`. |
| 9 | Usage estimation via tokenizer | **No test listed.** Add to `EventMapperTests` or a new `UsageEstimationTests`. |
| 11 | Default bind `127.0.0.1` | **No test listed.** Needs a CLI parser test that asserts default and rejects `0.0.0.0` without `--api-key-file` or `--unsafe-no-auth`. |
| 12 | `--api-key-file` vs `--api-key` | **No test listed.** Need a test that the value is read from disk, file-not-found returns a clear error, and `--api-key` emits the deprecation warning to stderr. |
| 14 | 400 on unsupported fields | Mentioned in unit suite; needs explicit cases for each of `n`, `logprobs`, `logit_bias`, `frequency_penalty`, `presence_penalty`, plus `response_format: json_schema`. The error body must name the field. |
| 15 | `/health/live` vs `/health/ready` | `HealthRouteTests` listed — verify `/ready` returns 503 before `loadModel` completes, 200 after. The pre-load case is the one that regresses. |

### `EndToEndOpenAIClientTests` design

Shelling to `python -c "from openai import …"` is the wrong knob for the friction it adds: contributors need a Python install, a pinned `openai` SDK version, and network access to PyPI on first run. The SDK also evolves — a minor bump can change wire expectations under us with no signal. Two better options: (a) commit a vendored, pinned `openai-python` wheel + `requirements.txt` and run via `python -m venv` deterministically; (b) skip Python entirely and write a Swift wire-compat fixture suite that replays the exact byte sequences the OpenAI SDK emits and consumes, captured once from a real client. Option (b) is what `Tests/BaseChatBackendsTests/CloudBackendSSETests.swift` already does for the inverse direction — same pattern, no subprocess. Keep one Python smoke test gated on `RUN_SERVER_E2E=1` for sanity, but don't make it the only wire-compat signal.

Cursor / Continue wire compat isn't CI-testable but should be a documented manual smoke checklist in PR 6's description (model dropdown loads, streaming renders, tool calls round-trip, cancel button works, reconnect after server restart).

### Sabotage check applicability

CLAUDE.md mandates sabotage; plan says "every assertion gets a temporary code break." For byte-sequence SSE assertions this is mostly straightforward (flip a string, confirm fail), but two paths are awkward: (1) the constant-time compare — sabotaging to `==` won't fail any functional test, only a timing test that the plan doesn't include; (2) cancellation tests where the assertion is "stopGeneration was called" — sabotaging the cancel path is easy, but ensure the test actually throws on absence rather than silently passing if the mock's `stopCallCount` wasn't observed at all.

### Determinism / hardware gating

No `XCTSkipIf` patterns in the plan. Server tests against `MockInferenceBackend` shouldn't need any, but the gated E2E suite needs guards for: `RUN_SERVER_E2E` env var, port-availability (don't hard-code 8080 — bind to `:0` and read back), and Python presence. Without these the suite is a flake source.

### Performance regression baseline

No `XCTMeasure` baselines in the plan. For a server, time-to-first-token and tokens/sec floor matter and they're easy to regress invisibly via middleware ordering or buffer sizing. Recommend adding one `XCTMeasure` suite in PR 4 covering: TTFT for a 10-event stream, full-stream wall time for 1000 events, and request parse + auth overhead for a non-streaming 1KB body. Treat as v1, not v2 — once the server is in users' hands, retroactive baselines are noisy.

### Test-support enhancements should land in PR 2 or 3, not later

`MockInferenceBackend` doesn't currently support: emitting parallel tool-call lanes with controlled interleaving, throttle-then-resume, or scripted `.usage` events. These need to land before PR 4 (streaming) opens, or PR 4 will grow to include test-support churn and become hard to review. Plan should explicitly list test-support deltas as part of PR 2 scope.

### Tests the plan should add (consolidated)

1. `BackpressureTests` — slow consumer, write-timeout, cancellation under buffer pressure.
2. `RequestParsingEdgeTests` — malformed JSON, oversize body, ragged UTF-8, wrong content-type.
3. `DegenerateInputTests` — empty messages, system-only, empty content, `max_tokens: 0`.
4. `UsageEstimationTests` — `include_usage` on/off, tokenizer fallback path, `x-basechat-usage-source` header.
5. `MidStreamErrorRouteTests` — using `MidStreamErrorBackend`, asserts SSE error chunk and `stopGeneration()`.
6. Expand `ChatCompletionsToolsTests` to cover lane interleaving sub-cases enumerated above.
7. `ApiKeyMiddlewareTests` parameterized over the auth-edge matrix; add a constant-time timing test (statistical, not a single sample).
8. `CORSPreflightTests` covering `OPTIONS`, `--unsafe-cors-any`, `Vary: Origin`.
9. `ClientDisconnectCancellationTests` parameterized over the five race windows.
10. `CLIArgumentTests` covering decisions 11, 12, default-bind enforcement.
11. `UnsupportedFieldsTests` enumerated against decision 14.
12. `HealthReadinessTests` covering pre-load 503 → post-load 200.
13. `SSEPerformanceBaselineTests` (`XCTMeasure`) — TTFT, full-stream, parse+auth.

### Answers to the four open questions, QA lens

1. **0.13.0 vs 0.14.0.** No QA opinion on release timing per se, but if 0.13.0 is also closing #753, the test-support churn on `MockInferenceBackend` will collide. Sequence the test-support work first regardless of which release.
2. **Public vs `package` `BaseChatServerCore`.** `package` is more testable in the short run — public APIs accrete contract tests we can't easily change later. Stay `package` until a real consumer asks.
3. **`--parallel` flag.** Drop from v1 CLI. A flag with no observable effect can't be tested honestly — any `--parallel 4` test would assert nothing the default-1 path doesn't already assert. Keeping it documented invites users to file bugs against behavior we never claimed to deliver. Reintroduce with cloud/Ollama pass-through when there's something to assert.
4. **Richer `/v1/models`.** Loaded-only is the testable choice. "Available on disk" pulls in filesystem state, model-cache layout, and HuggingFace cache assumptions — every one of those becomes a flake source on contributor machines and a snapshot to keep in sync. If clients want discovery, that's a v2 feature with its own test surface, not a v1 nice-to-have.

Net: ship the structure as-is, but expand the test list before PR 4 lands and pull `MockInferenceBackend` enhancements into PR 2's scope.

roryford:
**Review of #806 (BaseChatServer plan)**

Overall: direction is solid and the scope discipline (one library + thin exec, trait-gated, mock-backend-testable) is exactly the house style. Four observations before code lands.

1. **`prefill_progress` forwarding is missing from the wire-mapping table.** PR #804 explicitly names `BaseChatServer` as the cooperating proxy and ships the `X-BaseChat-Prefill-Progress: true` header + `event: prefill_progress` SSE shape. The plan's `GenerationEvent → chunk` table should include `.prefillProgress(nPast:nTotal:tokensPerSecond:)` emitting an `event: prefill_progress` frame when the request set the opt-in header — otherwise the v0.13.0 server can't actually feed the consumer that #804 just built for it.

2. **`TrafficBoundaryAuditTest` will block this on day one.** Hummingbird/NIOHTTP1 is the repo's first inbound networking surface; the audit currently only knows about `URLSessionProvider`-style outbound I/O. Plan needs an explicit item: extend the audit (new rule for inbound listeners) or add `BaseChatServerCore` files to a server-scoped allowlist with the reviewer sign-off the audit demands. Don't discover this in PR 2.

3. **Decision #8 invents `finish_reason: \"error\"`.** The OpenAI spec enumerates `stop | length | tool_calls | content_filter | function_call`. Even if the Python SDK tolerates a custom value today, Cursor/Continue/strict third-party clients may reject it. Risk-table mitigation says \"fall back to `stop` + error field\" — make that the v1 default, not the fallback.

4. **Open question 2 (Core public vs package):** ship `BaseChatServerCore` as `package`-visible. Going public locks routing/middleware shapes you'll want to refactor between v1 and v2; menubar-app embedding is a v2 ask, not a v1 commitment.

Skipping for now: SwiftLM line-reference fidelity, exact Hummingbird 2.x version pin, and the `--parallel` CLI question (open question 3) — those are fine to settle in PR 2.

— review by Claude Code worker (background agent)

roryford:
## Maintainer synthesis

**Verdict: accept with revisions, single force-push to plan v2.** All four reviews are substantive and most of the critique lands. The biggest things I'm taking: the dep-direction inversion (`BaseChatServerCore` → `BaseChatInference` only, executable wires backends), inverting the wire-mapping primary path so MLX's whole-call `.toolCall(_)` is the default and streamed deltas are the expansion, the `TrafficBoundaryAudit` blocker, and the OpenAI-spec `finish_reason` correction. Rollout compresses from 6 PRs to 3. The Python E2E shape changes. A handful of the bigger asks (lifting `OpenAIToolEncoding`, adding `.finished(reason:)` to `GenerationEvent`) belong on #807, not v1 — reasoning below.

### Accepted (will land in plan v2)

- **Dep direction.** Architect: `BaseChatServerCore` depends on `BaseChatInference` only; executable wires `BackendRegistrar` + concrete backends in. Preserves the same one-way edge CLAUDE.md enforces for UI and unblocks #807 menubar embedding without dragging Hummingbird through the default graph.
- **Wire-mapping primary path.** Architect: only MLX emits tool calls in v1 and it emits `.toolCall(_)` whole, not deltas (verified in `Sources/BaseChatBackends/MLXBackend.swift:62`). Mapping table will list `.toolCall(_)` as primary and `.toolCallStart`/`.toolCallArgumentsDelta` as the (currently dead) Ollama/cloud expansion path.
- **`finish_reason: \"error\"` is non-spec.** Worker review is right — OpenAI enumerates `stop|length|tool_calls|content_filter|function_call`. Plan v2 will make `finish_reason: \"stop\"` + an `error` object on the chunk the v1 default, not the fallback.
- **`TrafficBoundaryAuditTest` extension is a v1 blocker.** Worker review caught this. `Tests/BaseChatInferenceTests/TrafficBoundaryAuditTest.swift:67` only knows outbound `URLSessionProvider`; Hummingbird is the repo's first inbound surface. Plan v2 adds an explicit PR-2 item: extend the audit with an inbound-listener rule (or scoped allowlist with sign-off), discovered now not at PR-2 review time.
- **`prefill_progress` forwarding.** Worker review: PR #804 already names BaseChatServer as the cooperating proxy. Wire-mapping table will add `.prefillProgress(...)` → `event: prefill_progress` SSE frame, gated on the `X-BaseChat-Prefill-Progress: true` request header.
- **Rollout compression to 3 PRs.** DevOps: PR-2 skeleton burns CI for nothing; PR-5 (tools) and PR-6 (CLI) are too thin. New shape: PR-A skeleton+non-streaming+health+auth+CORS, PR-B streaming+tools+cancellation+`EventMapper`, PR-C CLI+executable+README+manual smoke. Saves ~3 CI cycles at 10× billing.
- **CI: fold into existing batch step, add dep-graph lint.** DevOps: pass `--traits Server --filter BaseChatServerCoreTests` into the existing batch invocation at `.github/workflows/ci.yml:166`, not a new step. Add a one-shot `swift package show-dependencies --disable-default-traits --format flatlist | grep -qv hummingbird || exit 1` plus an import-grep lint asserting no default-graph target imports `Hummingbird` (mirrors the UI→ModelManagement guard at `ci.yml:105-117`).
- **Hummingbird pinning.** Architect + DevOps both pushed. Plan v2 uses `.upToNextMinor(from: \"2.22.0\")` in `Package.swift`, exact pinning lives in the executable's `Package.resolved`. Library-exact-pin is hostile to downstream consumers.
- **Operability section.** DevOps: SIGTERM graceful drain (default 30s), SIGINT for foreground, JSON-line `--log-format` option, `OSLog` for uncaught errors, friendly bind-port-in-use message, README honesty about Gatekeeper / App Translocation / plaintext `--api-key-file`. New \"Operability\" section in plan.
- **Test-support deltas land in PR-A.** QA: `MockInferenceBackend` doesn't emit `.usage` or controlled tool-call lane interleaving today. Pulling these into PR-A scope so PR-B doesn't grow to include test-support churn. `SlowMockBackend`, `ChaosBackend`, `MidStreamErrorBackend` exist already — plan will reference them by path.
- **Test list expansion.** All 13 of QA's enumerated suites get added to the plan's test strategy section, plus the explicit decision-vs-test mapping table (decisions 9, 11, 12, 14, 15 currently have no failing test if regressed).
- **`XCTMeasure` baselines in PR-B.** QA: TTFT, full-stream wall time, parse+auth overhead. Cheaper to set baselines now than retroactively.
- **Cancellation suite split.** QA: `ClientDisconnectCancellationTests` parameterized over the five race windows, plus a test where `stopGeneration()` itself throws/hangs to assert no deadlock.
- **`--parallel` dropped from v1 CLI.** DevOps + QA aligned. Cargo-cult surface; reintroduce with cloud/Ollama pass-through in v2 when there's something to assert. (Worker review said \"keep documented\" — overruled, see Pushed back.)
- **`/v1/models` returns loaded model only.** All three reviewers aligned. Plan already implied this; v2 will state it as a hard decision and route handlers will be written against `InferenceService` (not a captured `InferenceBackend`) so v2 multi-model has a clean migration.
- **`AsyncSemaphore` not a singleton.** Architect: pass it in or make it a protocol so v2 per-model semaphores aren't a rewrite.
- **IPv6 + TLS-via-reverse-proxy docs.** DevOps: `--bind ::1` supported, `[::]` behind auth gate, README points users at Caddy/cloudflared for TLS.
- **400-LOC file rule enforced in CI**, not just plan prose. DevOps.

### Pushed back

- **Architect: lift `OpenAIToolEncoding` to a shared `BaseChatOpenAIWire` module before PR-B.** Real concern but wrong moment. The duplication is genuine, but extracting a new module to dissolve `[String: Any]` payloads into Codable structs is a refactor that affects `CloudBackend` and the new server simultaneously. CLAUDE.md \"one feature = one PR\" cuts both ways: doing the lift inside the server PR couples two independently-revertible changes. **Going on #807** as a v2 cleanup; plan v2 will note the duplication explicitly so we don't lose it.
- **Architect: add `.finished(reason: FinishReason)` to `GenerationEvent` before PR-B.** Same reasoning. The protocol churn touches `GenerationCoordinator`, `GenerationStreamConsumer`, `ToolCallLoopOrchestrator`, `ScenarioRunner`, `EventRecorder` plus every backend's stream end. That's a cross-cutting refactor with its own risk profile and deserves its own PR. The server's private `FinishReasonDeriver` is small and isolated; if v2 extracts the protocol case, the deriver collapses to a passthrough. **Going on #807.**
- **Worker: keep `--parallel` documented as future-facing.** Disagree with QA's reasoning carrying more weight: a flag with no observable effect can't be tested honestly, and the flag→behavior surface is cheap to re-add. Removing it is technically a CLI breaking change but no v1 user has shipped against it yet. Drop now, re-add cleanly in v2.
- **DevOps: skip the path-scoped CI sub-job.** Already agreed not to add one; calling it out so this doesn't resurface.

### Genuine splits left for the human maintainer

- **0.13.0 vs 0.14.0 release window.** DevOps says 0.13.0 (decoupled tracks). Architect says 0.14.0 (server is the more dangerous surface and deserves its own headline, tool-calling #753 is the more compelling 0.13.0 story). I lean 0.14.0 on the safety argument but this is a product call — one review wants concentrated reviewer attention on the network surface, the other wants release independence. **Calling for human verdict.**
- **`BaseChatServerCore` public vs `package`.** Three of four reviewers say `package` (Architect's \"public\" is conditional on the dep direction being fixed first, which it now is, so it's effectively a 2-2). Premature `public` is a one-way door; `package` is testable today and promotable later. I lean `package` for v1, but Architect's menubar-host argument has weight if #807 work starts soon. **Calling for human verdict.**

### Plan v2 commit list

1. Rewrite Decision 8: `finish_reason` uses spec-only values; error path emits `\"stop\"` + `error` object on chunk (not custom `\"error\"`).
2. Rewrite Module structure: `BaseChatServerCore` deps shrink to `BaseChatInference` + `Hummingbird` (drop `BaseChatBackends`); executable target gets `BaseChatBackends` + `BackendRegistrar` wiring.
3. Invert the wire-mapping table: `.toolCall(_)` is primary tool path, deltas are the expansion path; add `.prefillProgress(...)` → `event: prefill_progress` row gated on `X-BaseChat-Prefill-Progress: true` request header.
4. Add new \"Operability\" section: SIGTERM/SIGINT semantics, `--log-format json|text`, OSLog destination, port-in-use message, README honesty checklist.
5. Add new \"Trait & dep-graph guards\" subsection under CI: `swift package show-dependencies` lint, import-grep lint, file-LOC lint.
6. Rewrite Decision 13 + Open Question 3: `--parallel` dropped from v1 CLI.
7. Rewrite Decision 16: keep hand-rolled Codable for v1; add explicit pointer to #807 for the `BaseChatOpenAIWire` extraction with `OpenAIToolEncoding` reuse rationale.
8. Rewrite Rollout: 3 PRs (PR-A skeleton+non-streaming+middleware, PR-B streaming+tools+cancellation, PR-C CLI+executable+README); list test-support deltas (`MockInferenceBackend` `.usage` + parallel tool-call interleaving) as PR-A scope.
9. Rewrite Test strategy: enumerate the 13 QA suites; add the decision-vs-test mapping table; reference `SlowMockBackend`/`ChaosBackend`/`MidStreamErrorBackend` by path; add `XCTMeasure` baselines (TTFT, full-stream, parse+auth) to PR-B; replace `python -c` shell-out with vendored-wheel + `python -m venv` deterministic fixture path, and add a Swift wire-compat byte-replay suite (mirrors `Tests/BaseChatBackendsTests/CloudBackendSSETests.swift`); shell-out kept as one `RUN_SERVER_E2E=1`-gated sanity check.
10. Add new \"Pre-PR-A blockers\" subsection: extend `TrafficBoundaryAuditTest` with inbound-listener rule before any server source files exist on disk.
11. Update Hummingbird pin: `.upToNextMinor(from: \"2.22.0\")` in `Package.swift`; note `ExistentialAny` + `InternalImportsByDefault` strict-settings smoke must pass before committing the dep.
12. Update Risks table: minor-version pin (not exact); add `AsyncSemaphore` injectability for v2 per-model.
13. Update Open questions: Q3 (`--parallel`) marked resolved → drop; Q4 (`/v1/models`) marked resolved → loaded only; Q1 (release) and Q2 (visibility) flagged as awaiting human verdict.
14. Add explicit IPv6 + reverse-proxy-TLS paragraph to Security posture section.

### Open follow-ups for #807

- Lift `OpenAIToolEncoding` to a shared `BaseChatOpenAIWire` module; converge server outbound + cloud inbound on one Codable surface.
- Add `case finished(reason: FinishReason)` to `GenerationEvent`; collapse `FinishReasonDeriver` and the four existing in-tree consumers that derive end-of-stream from event-shape heuristics.
- `--parallel` reintroduction alongside cloud/Ollama backend pass-through.
- Constant-time-compare statistical timing test for `ApiKeyMiddleware` (single-sample timing tests are noise; needs a proper microbenchmark harness).
- Per-model `AsyncSemaphore` for multi-model swap.
- Pre-built signed/notarized binary distribution + Homebrew formula.
- `/v1/models` enrichment with on-disk availability once multi-model swap lands.
- `?include=thinking` extension for clients that grow a way to render reasoning blocks.
- Menubar-app embedding host — gates the `package` → `public` promotion of `BaseChatServerCore`.

Force-push to `feat/744-server-plan` lands once the two open splits are called.

roryford:
## Plan v2 pushed — ready for re-review

Force-push avoided in favor of a clean second commit (`0b54e8f`) so reviewers can see the diff. All 14 items from the maintainer synthesis are applied; the two open splits resolved on the conservative side (0.14.0 + \`package\`-visible).

### What changed

- **Decisions renumbered** — was 17, now 21 (added 6 new explicit decisions: dep direction, Core visibility, finish-reason spec values, prefill_progress contract, AsyncSemaphore injectability, no-built-in-TLS posture).
- **Wire-mapping table inverted** — \`.toolCall(_)\` is the primary v1 path (verified at \`MLXBackend.swift:62\`), streamed deltas demoted to expansion path. \`prefill_progress\` row added per #804.
- **Pre-PR-A blockers** section is new — \`TrafficBoundaryAuditTest\` extension and \`MockInferenceBackend\` enhancements must land before any \`Sources/BaseChatServer*\` files exist on disk.
- **Operability** section is new — SIGTERM drain, \`--log-format json|text\`, OSLog, friendly bind-port-in-use, README honesty checklist.
- **CI** section gains three falsifiable lints — dep-graph cleanliness, import-grep, 400-LOC ceiling. Test invocation folds into the existing batch step at \`ci.yml:166\`, not a third \`swift test\` call.
- **Rollout** compresses 6 → 3 PRs (PR-A/B/C). PR-A absorbs test-support deltas so PR-B doesn't bloat.
- **Test strategy** expanded — 13 enumerated suites, decision-vs-test mapping table for the 6 decisions that previously had no failing-on-regression test, \`XCTMeasure\` baselines, \`WireCompatReplayTests\` replaces Python shell-out as primary (one gated SDK sanity remains).
- **Cancellation** is parameterized over 5 race windows.
- **Security** gets explicit IPv6 + reverse-proxy-TLS paragraph.
- **Open questions** all marked resolved at the bottom.

### Deferred to #807

Two architect asks I pushed back on are now explicit non-goals with #807 pointers:

- Lifting \`OpenAIToolEncoding\` to \`BaseChatOpenAIWire\` Codable module
- Adding \`case finished(reason: FinishReason)\` to \`GenerationEvent\`

Both are real refactors that cross-cut \`CloudBackend\` + the new server simultaneously; doing them inside the server PR couples independently-revertible changes. Plan notes this so we don't lose the convergence work.

### Two splits called

- **Release window: 0.14.0** (was open). 0.13.0 ships #753 PR-C close-out; the server gets concentrated reviewer attention on its own headline release. Reversible — easy to pull forward if PR-C lands faster than expected.
- **\`BaseChatServerCore\` visibility: \`package\`** (was open). One-way door; promote on the menubar embedding ask.

If either verdict is wrong, holler — both are single-line plan edits.

Ready for re-review or merge.

roryford:
Senior reviewer pass — pushed two factual tightenings (289991f); plan otherwise lands well.

**Verified against the codebase:**
- Module layering matches CLAUDE.md (`BaseChatServerCore` Inference-only, `BaseChatServer` exe wires `BaseChatBackends`). The same one-way edge pattern as the UI/UIModelManagement peel.
- `Server` trait gating mirrors existing `Fuzz` / `MCPBuiltinCatalog` precedent in `Package.swift`.
- CI line refs `.github/workflows/ci.yml:166` (batch test step) and `:105-117` (UI→ModelManagement import-grep guard) check out.
- `Tests/BaseChatInferenceTests/TrafficBoundaryAuditTest.swift` exists and is outbound-only — Pre-PR-A blocker (1) is the right framing.
- Test-support inventory checks out: `MidStreamErrorBackend.swift`, `SlowMockBackend.swift`, `ChaosBackend.swift` all present in `Sources/BaseChatTestSupport/`.
- `Tests/BaseChatBackendsTests/CloudBackendSSETests.swift` exists — the byte-replay precedent the wire-compat suite cites.

**Fixes pushed:**
1. **Wire-mapping section** — the cited `MLXBackend.swift:62` ref pointed at `supportsToolCalling: true`, not the absence of `streamsToolCallArguments`. Repointed to line 57 (the `BackendCapabilities` literal) and added the corroborating `supportsToolCalling: false` lines on `LlamaBackend.swift:53` and `FoundationBackend.swift:80`. Also named the streamed-delta backends with line refs (`OllamaBackend:123`, `OpenAIBackend:81`, `ClaudeBackend:93`) so the "currently dead path" claim is checkable.
2. **Module-structure deps** — production library targets can't test-only-depend on `BaseChatTestSupport`. Split into three lines: `BaseChatServerCore` library deps, `BaseChatServerCoreTests` deps (where `BaseChatTestSupport` actually belongs), and the executable's deps.

**Not fixed (intentional):**
- Decision 18 (`/health/live` + `/health/ready`) upgrades issue #744's `GET /health` surface. That's a deliberate, well-justified expansion.
- Decision 7 / decision 16 drop `cloud`/`ollama` backends and `--parallel` from v1; both are tracked on #807 with sound rationale.

**Open question for the maintainer (not in scope to fix):** decision 21 mentions `SSEStreamLimits` for response-side caps but caveats that flow control is Hummingbird's writer backpressure. Worth confirming Hummingbird 2.x's writer actually exposes a backpressure-aware async sink before PR-B baselines that test goes in — easy to discover painfully at PR-B time otherwise.

Plan is ready for merge once persona reviewers weigh in.

PR #810, opened 2026-04-26
feat(inference): capability routing — per-request capability requirements + RouterBackend

### LINKED ISSUE #438: feat(inference): RouterBackend — dispatch per-request across backends by capability
## Motivation

Apps often want "small local for classification, larger local or remote for reasoning" routing. Today this requires host-app glue. A framework-provided router would compose existing backends without any backend changes.

## Proposal

Add `RouterBackend: InferenceBackend` to `BaseChatInference`:

- Constructed with an ordered list of `(InferenceBackend, RoutingPredicate)`
- On each `generate`, picks the first backend whose predicate matches the request
- Default predicates: by model tier, by context size, by presence of tools / images, by explicit override in a request hint

Composition-only — no changes to existing backends.

## Depends on

- Per-request capability requirements on `GenerationConfig` (companion issue)

## Acceptance criteria

- Unit tests against `MockInferenceBackend` instances with various predicates
- DocC page explaining the routing model and when to use it
- No breaking change to the `InferenceBackend` protocol

### LINKED ISSUE #439: feat(inference): per-request capability requirements on GenerationConfig
## Motivation

A request can declare what it needs (tools, vision, minimum context, minimum model tier) independently of which backend happens to be wired up. Useful for validation (fail fast if a request cannot be served), for RouterBackend dispatch, and for UI gating.

## Proposal

Extend `GenerationConfig` with an optional `requirements: GenerationRequirements`:

- `requiresTools: Bool`
- `requiresVision: Bool`
- `minContextTokens: Int?`
- `minModelTier: ModelCapabilityTier?`

Backends expose a `capabilities: BackendCapabilities` surface. Framework offers `backend.canSatisfy(requirements:)`.

## Acceptance criteria

- No breaking change — requirements are optional
- `RouterBackend` consumes this for dispatch
- Existing backends populate `capabilities` accurately, including `FoundationBackend` gating on OS version

roryford:
## Summary

Bundled rollout of two cooperating issues from the tool-calling umbrella (#753).
One alone would have no consumer or input — they ship as a single PR.

- **#439**: per-request capability requirements on `GenerationConfig`
- **#438**: `RouterBackend` that dispatches each request across children by capability

### `GenerationConfig.requiredCapabilities`

```swift
let config = GenerationConfig(
    requiredCapabilities: [.toolCalling, .minContextTokens(8_000)]
)
```

Each `GenerationCapabilityRequirement` case maps onto an existing
`BackendCapabilities` flag — no parallel system. The new field defaults
to an empty set, so existing call sites are unaffected. Codable
round-trips, and older payloads decode to an empty set.

### `RouterBackend`

```swift
let router = RouterBackend(children: [smallLocal, capableCloud])
let stream = try router.generate(
    prompt: prompt,
    systemPrompt: nil,
    config: GenerationConfig(requiredCapabilities: [.toolCalling])
)
```

The router walks `children` in priority order and picks the first whose
`BackendCapabilities.satisfies(requirements)` is `true`. When no child
qualifies it throws `InferenceError.noBackendSatisfiesRequirements`
listing the missing capabilities. Single child per request — the router
does not fan out, retry across children, or load-balance.

Lifecycle ops (`stopGeneration`, `unloadModel`, `resetConversation`)
fan out to every child. `loadModel` is rejected — picking a model is the
host's job; reach for `RouterBackend` only when a single conceptual
session multiplexes across already-loaded children.

`capabilities` returns the union of children's capabilities so a UI can
ask "can the runtime as a whole do X?". Per-request the right answer
comes from `requiredCapabilities` and the dispatch in `generate(...)`.

#### Per-request capability routing

Requests can now declare what they need from a backend (tool calling,
thinking, grammar-constrained sampling, a minimum context window) on
the `GenerationConfig` itself. The new `RouterBackend` composes a list
of children and dispatches each request to the first one that can serve
it — useful for "small local for classification, larger local or remote
for reasoning" wiring without per-app glue.

```swift
let router = RouterBackend(children: [smallLocal, capableCloud])
let config = GenerationConfig(
    requiredCapabilities: [.toolCalling, .minContextTokens(8_000)]
)
let stream = try router.generate(prompt: prompt, systemPrompt: nil, config: config)
```

`requiredCapabilities` defaults to an empty set, so existing call sites
keep working unchanged. When no wired backend satisfies a request the
router throws `InferenceError.noBackendSatisfiesRequirements` listing
the missing capabilities. See [#438], [#439], [#753].

### Features

- **inference:** per-request `requiredCapabilities` on `GenerationConfig` ([#439])
- **inference:** `RouterBackend` composes children and dispatches by capability ([#438])

## Test plan

- [x] `swift test --filter BaseChatCoreTests --disable-default-traits`
- [x] `swift test --filter BaseChatInferenceTests --disable-default-traits` (includes new `RouterBackendTests`)
- [x] `swift test --filter BaseChatInferenceSwiftTestingTests --disable-default-traits`
- [x] `swift test --filter BaseChatUITests --disable-default-traits`
- [x] `swift test --filter BaseChatUIModelManagementTests --disable-default-traits`
- [x] `swift test --filter BaseChatMCPTests --disable-default-traits`
- [x] `swift test --filter BaseChatBackendsTests --disable-default-traits`
- [x] `swift test --filter BaseChatTestSupportTests --disable-default-traits`
- [x] `swift test --filter BaseChatAppIntentsTests --disable-default-traits`
- [x] Sabotage check: stripping the requirement filter in `selectBackend` causes 5 router tests to fail; reverted.

Closes #438
Closes #439
Tracking: #753

---

Reviewed end to end against the umbrella's PR-D scope and the product-engineer lens (API ergonomics, failure-mode quality, discoverability, naming, test depth, release-note clarity).

**Findings — no code changes.** The implementation is honest:

- `GenerationCapabilityRequirement` cases map 1:1 onto existing `BackendCapabilities` flags; no parallel system. The parameterised `.minContextTokens(Int)` case is the only one that doesn't reduce to a flag, and that's the right shape — context size is a quantitative requirement, not a flag.
- `RouterBackend.selectBackend(for:)` is exposed publicly so a host can do a dry-run check before calling `generate`, not just discover via thrown error mid-stream.
- `loadModel` deliberately rejects with a clear `inferenceFailure` message — matches the doc-comment's stance that model selection is the host's job.
- `BackendCapabilities` union semantics are documented at the type level (per-flag OR; numeric maxima taken; `requiresPromptTemplate` deliberately false-beats-true).
- `noBackendSatisfiesRequirements` payload computes the *unsatisfied* set across all children, not just a copy of the request — gives the host enough to render a fix-it message. Falls back to the full requirement list only when the failure is partial-coverage (every child fails some, no single requirement fails everywhere) — guards against an empty diagnostic.
- Codable backwards-compat is real: the `older payload decodes to empty set` test reconstructs a legacy payload by stripping the field from a serialised default, not just asserting the default.
- Tests cover the satisfiable-across-children-but-no-single-child guard — the right product contract for a router that does *not* split a single request.

- `noBackendSatisfiesRequirements`'s error description prints requirements via `String(describing:)` which renders enum cases like `minContextTokens(1000)`. Readable for a developer log, ugly for an end-user surface. If a future host wants to render this in UI, a `humanDescription` helper on `GenerationCapabilityRequirement` would help — left out of scope here.
- Demo coverage: `Example/` does not yet show a router wiring. Could be a follow-up if the maintainer wants a "small local + remote fallback" scenario in the demo picker.

### DIFF
--- a/Sources/BaseChatInference/Models/GenerationCapabilityRequirement.swift
+++ b/Sources/BaseChatInference/Models/GenerationCapabilityRequirement.swift
@@ -1,0 +1,77 @@
+import Foundation
+
+/// A capability a request requires of whichever backend serves it.
+///
+/// Used by ``GenerationConfig/requiredCapabilities`` and ``RouterBackend``
+/// to fail fast when no wired backend can satisfy a request, and to dispatch
+/// the request to a child backend that can.
+///
+/// New requirements should map onto an existing ``BackendCapabilities`` flag
+/// rather than introducing a parallel system. The mapping lives in
+/// ``BackendCapabilities/satisfies(_:)``.
+public enum GenerationCapabilityRequirement: Sendable, Hashable, Codable {
+    /// The backend must support tool/function calling
+    /// (``BackendCapabilities/supportsToolCalling``).
+    case toolCalling
+
+    /// The backend must support structured (JSON-schema) output
+    /// (``BackendCapabilities/supportsStructuredOutput``).
+    case structuredOutput
+
+    /// The backend must support a native JSON-object generation mode
+    /// (``BackendCapabilities/supportsNativeJSONMode``).
+    case jsonMode
+
+    /// The backend must emit thinking events
+    /// (``BackendCapabilities/supportsThinking``).
+    case thinking
+
+    /// The backend must support grammar-constrained sampling
+    /// (``BackendCapabilities/supportsGrammarConstrainedSampling``).
+    case grammarConstrainedSampling
+
+    /// The backend must be capable of emitting multiple tool calls per turn
+    /// (``BackendCapabilities/supportsParallelToolCalls``).
+    case parallelToolCalls
+
+    /// The backend must stream tool-call arguments
+    /// (``BackendCapabilities/streamsToolCallArguments``).
+    case streamingToolCalls
+
+    /// The backend must persist KV cache state across consecutive
+    /// `generate()` calls (``BackendCapabilities/supportsKVCachePersistence``).
+    case kvCachePersistence
+
+    /// The backend's effective context window must be at least `tokens` wide.
+    case minContextTokens(Int)
+}
+
+extension BackendCapabilities {
+    /// Whether this capability set satisfies a single requirement.
+    public func satisfies(_ requirement: GenerationCapabilityRequirement) -> Bool {
+        switch requirement {
+        case .toolCalling:                return supportsToolCalling
+        case .structuredOutput:           return supportsStructuredOutput
+        case .jsonMode:                   return supportsNativeJSONMode
+        case .thinking:                   return supportsThinking
+        case .grammarConstrainedSampling: return supportsGrammarConstrainedSampling
+        case .parallelToolCalls:          return supportsParallelToolCalls
+        case .streamingToolCalls:         return streamsToolCallArguments
+        case .kvCachePersistence:         return supportsKVCachePersistence
+        case .minContextTokens(let n):    return contextWindowSize >= n
+        }
+    }
+
+    /// Whether this capability set satisfies every requirement in the set.
+    public func satisfies(_ requirements: Set<GenerationCapabilityRequirement>) -> Bool {
+        requirements.allSatisfy { satisfies($0) }
+    }
+
+    /// Subset of `requirements` that this capability set fails to satisfy.
+    /// Used for diagnostic error messages when routing fails.
+    public func unsatisfied(
+        from requirements: Set<GenerationCapabilityRequirement>
+    ) -> Set<GenerationCapabilityRequirement> {
+        Set(requirements.filter { !satisfies($0) })
+    }
+}
--- a/Sources/BaseChatInference/Models/GenerationCapabilityRequirement.swift
+++ b/Sources/BaseChatInference/Models/GenerationCapabilityRequirement.swift
@@ -44,6 +44,25 @@ public enum GenerationCapabilityRequirement: Sendable, Hashable, Codable {
 
     /// The backend's effective context window must be at least `tokens` wide.
     case minContextTokens(Int)
+
+    /// Stable sort key for diagnostic ordering. Logged error payloads need a
+    /// deterministic order so test assertions and log diffs stay stable across
+    /// runs — the underlying `Set` storage iterates in insertion-dependent
+    /// order. Not part of the public API contract; the case names happen to
+    /// match `String(describing:)` today but callers should not rely on that.
+    var sortKey: String {
+        switch self {
+        case .toolCalling:                return "1.toolCalling"
+        case .structuredOutput:           return "2.structuredOutput"
+        case .jsonMode:                   return "3.jsonMode"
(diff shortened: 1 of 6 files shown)


roryford:
## Review pass — rebased + Copilot fixes

**Rebase:** clean after one trivial conflict in `Sources/BaseChatInference/Protocols/InferenceBackend.swift` — `streamPrefillProgress` (PR #804, on main) and `requiredCapabilities` (this PR) both landed in the same field block. Resolved by keeping both, sorted into the existing legacy/modern initializer pair plus updated Codable.

**Copilot inline review** — all six items addressed in `8b95e77`:

| # | Concern | Fix |
|---|---|---|
| 1 | Empty-children init produces unhelpful `.noBackendSatisfiesRequirements([])` | `precondition(!children.isEmpty, …)` at init |
| 2 | Nondeterministic ordering in error payload | added `GenerationCapabilityRequirement.sortKey`; sorted at throw site |
| 3 | Manual `BackendCapabilities` union easy to miss when fields added | left inline (single source of truth, not refactored) — see follow-up note |
| 4 | Doc-comment said router "does not advertise a loaded model" but `isModelLoaded` is `true` if any child is | rewrote comment |
| 5 | Empty `unmet` → dangling colon in `errorDescription` | branched on `isEmpty` to a generic message |
| 6 | `selectBackend` ignored `isModelLoaded` | now prefers loaded children; falls back to unloaded only if no loaded satisfies |

Re #3, kept the inline-merge as-is for now: a `BackendCapabilities.merged(with:)` helper has wider design implications (memoryStrategy/cancellationStyle merge semantics belong on the type, not on the router) and the current test coverage would not have caught the manual-merge drift Copilot warns about. Worth a separate follow-up if it bites.

**Sibling overlap matrix (Batch 4)**
- #811 (`feat/444-turn-history-compression`): touches `Services/ToolCallLoopOrchestrator.swift` + `Services/TurnHistoryCompressor.swift` — **no overlap**
- #812 (`feat/434-foundation-tool-calling`): backends-only — **no overlap**
- #813 (`feat/627-toolcall-validity-detector`): fuzz harness only — **no overlap**
- #814 (`feat/mlx-kv-cache-reuse-749`): MLX backend only — **no overlap**

The conflicting field on main was #804 (`streamPrefillProgress`), which is already merged. None of the in-flight Batch-4 PRs touch `Protocols/InferenceBackend.swift` or `Services/InferenceError.swift`, so this PR can land independently.

**Layering:** `BaseChatInference` only — `import Foundation` in both new files; no `BaseChatBackends` reference.

**Breaking changes:** none. `requiredCapabilities` defaults to empty; existing callers unaffected. Older Codable payloads decode to `[]`. The init signature gains a defaulted parameter.

**Tests:** local pre-push pass — `BaseChatInferenceTests` (1209), `BaseChatInferenceSwiftTestingTests` (83), `BaseChatBackendsTests` (4 cloud), `BaseChatTestSupportTests` (16), `BaseChatCoreTests` (271), `BaseChatMCPTests` (94). Two new tests pin loaded-preference behaviour; one new test pins the empty-unmet error-description shape.

CI run will follow.

PR #811, opened 2026-04-26
feat(inference): turn-history compression for agent-loop transcripts

### LINKED ISSUE #444: feat(inference): turn-history compression for agent-loop transcripts

roryford:
## Summary

Adds an opt-in `TurnHistoryCompressor` protocol invoked by `ToolCallLoopOrchestrator` before each new generate round. Older tool calls and tool results can be folded into a compact summary while the most recent N rounds stay verbatim, keeping long agent loops within a model's context window without losing referenceable older facts.

The default is a `NoOpTurnHistoryCompressor` — existing callers see no behaviour change. Apps opt in by passing a `BudgetTurnHistoryCompressor` (or a custom implementation) to the orchestrator initializer.

## Design

- `TurnHistoryRecord` — captures one round (intermediate tokens, tool calls, tool results) with a 1-indexed `step` so summary text can reference "step 2: weather(...) → 18C".
- `CompressedTranscript` — pure-data output of a compressor: `summary`, `foldedRecords`, `preservedRecords`. Order is preserved; `preservedRecords` is always a suffix of the input.
- `TurnHistoryCompressor` — the protocol. `Sendable`, deterministic, idempotent on already-compressed input.
- `BudgetTurnHistoryCompressor` — default in-tree implementation. Trips at a configurable character budget and emits a "Earlier turns summarised: N rounds, M tool calls. Notable results: …" block.

The orchestrator now tracks per-round records, calls `compress(records:)` between rounds, accumulates a carry-over summary string for facts that were folded in earlier iterations, and rebuilds the next-turn prompt as `initialPrompt + summary + verbatim appendices`. KV-cache prefix reuse stays well-defined: when the no-op compressor is used the prompt is byte-for-byte identical to the legacy path.

## Test plan

- [x] Pure compressor: empty/short input is unchanged; oversize input folds older records and preserves recent N; summary cites step numbers and excerpts results.
- [x] `NoOpTurnHistoryCompressor` always returns input unchanged.
- [x] Idempotent on already-compressed suffix.
- [x] Default-off integration: omitting the compressor produces a final prompt with verbatim appendices and *no* summary block (existing behaviour).
- [x] Round-trip integration: a 5-turn loop with a tight budget folds older results into a summary; the final-round prompt contains the latest result verbatim, the summary references older steps, and the earliest verbatim result is gone.
- [x] Sabotage check: bypassing the compressor in the orchestrator (substituting `NoOpTurnHistoryCompressor()` at the call site) makes the integration test fail on three assertions.
- [x] Pre-push checklist: all 9 CI-safe suites pass locally.

#### Turn-history compression for agent-loop transcripts

Long agent loops historically blew past the context window because every tool call, tool result, and intermediate model turn stayed in the transcript fed to the next generate round. `ToolCallLoopOrchestrator` now accepts an optional `TurnHistoryCompressor` that folds older rounds into a compact summary while preserving the most recent N verbatim. The default is a no-op so existing callers see no behaviour change.

```swift
let compressor = BudgetTurnHistoryCompressor(
    characterBudget: 8_000,
    preserveRecentTurns: 2
)
let orchestrator = ToolCallLoopOrchestrator(
    backend: backend,
    registry: registry,
    policy: ToolCallLoopPolicy(maxSteps: 12),
    compressor: compressor
)
```

The bundled `BudgetTurnHistoryCompressor` peels oldest records once a character budget is exceeded and emits a `step N: tool(args) → result` summary so the model can still answer reference questions about earlier turns. Custom compressors conform to the `TurnHistoryCompressor` protocol.

See [#444].

Closes #444

---

Reviewed end to end against the product-engineer lens (default-off ergonomics, what-gets-lost during compression, recent-N preservation, idempotency, test depth, release-note clarity).

**Findings — no code changes.** The implementation is honest and well-bounded:

- Default is `NoOpTurnHistoryCompressor` so existing callers see byte-for-byte identical wire format. The `test_defaultOff_existingPromptShapeUnchanged` test pins this — silently regressing the legacy shape would fail it.
- `BudgetTurnHistoryCompressor` defaults are conservative (`characterBudget: 8_000`, `preserveRecentTurns: 2`, `maxResultExcerpts: 5`). Apps that want compression turned on can construct it without arguments and get sensible behavior.
- The protocol contract (preserve-order, idempotent, deterministic) is documented and enforced — `preservedRecords` is a suffix, `foldedRecords` is a prefix, and the `test_idempotent_onAlreadyCompressedSuffix` test pins the second-pass-is-no-op invariant. KV-cache prefix reuse depends on this and the test guards it.
- Carry-over summary across iterations works — earlier folds' summary text accumulates onto `carryOverSummary` so older facts survive even after their records leave the running list. The `test_compressionEnabled_foldsOldResultsAfterBudget` test verifies an old fact is reachable via the summary on the final-turn prompt.
- Excerpt format is structured and step-numbered (`step N: tool(args) → result`) so the model can reference earlier turns by step number in subsequent turns.
- The `BudgetTurnHistoryCompressor` clamps inputs (`max(1, characterBudget)`, etc.) so degenerate constructions don't fold-everything.

- **`TurnHistoryRecord.intermediateTokens` is captured but never read.** The orchestrator's `consumeOnce` was modified to record visible tokens per round, but neither `rebuildPrompt` nor `BudgetTurnHistoryCompressor.renderSummary` references the field. It's also not in the legacy "appendix" shape that pre-#811 callers received. So today it's dead data — collected, persisted into `Equatable`, and never used. Two reasonable resolutions: (a) wire it into the rebuilt prompt so model-said-thinking carries forward, or (b) drop the field for now and reintroduce when there's a consumer. Worth deciding before this lands so we don't ship an unused public field.
- **Token budget vs character budget**: the public knob is character count. The default's docstring explains the rough proxy ("8000 chars conservative for 4k token model after system + initial"). If a future tokenizer dependency makes its way in, a token-budget knob would be more honest. Not a blocker.
- **Recent-N is `min(preserveRecentTurns, records.count)`** — never folds every record. This is documented in the implementation but not in the protocol's public docstring; a host implementing a custom compressor might not know that's the contract. Worth mentioning in a later doc pass.

### DIFF
--- a/Sources/BaseChatInference/Services/ToolCallLoopOrchestrator.swift
+++ b/Sources/BaseChatInference/Services/ToolCallLoopOrchestrator.swift
@@ -229,6 +229,7 @@ public struct ToolCallLoopOrchestrator: Sendable {
     private let backend: any InferenceBackend
     private let dispatcher: Dispatcher
     private let policy: ToolCallLoopPolicy
+    private let compressor: any TurnHistoryCompressor
 
     /// Creates an orchestrator driven by a single ``ToolExecutor``.
     ///
@@ -241,14 +242,19 @@ public struct ToolCallLoopOrchestrator: Sendable {
     ///   - backend: The backend that produces tokens and tool calls.
     ///   - executor: Receives every dispatched call.
     ///   - policy: Loop bounds. Defaults to ``ToolCallLoopPolicy/init(maxSteps:perStepTimeout:loopDetectionWindow:)``.
+    ///   - compressor: Optional ``TurnHistoryCompressor`` invoked before each
+    ///     new generate round. Defaults to ``NoOpTurnHistoryCompressor`` so
+    ///     existing callers see no behaviour change.
     public init(
         backend: any InferenceBackend,
         executor: any ToolExecutor,
-        policy: ToolCallLoopPolicy = .init()
+        policy: ToolCallLoopPolicy = .init(),
+        compressor: any TurnHistoryCompressor = NoOpTurnHistoryCompressor()
     ) {
         self.backend = backend
         self.dispatcher = .singleExecutor(executor)
         self.policy = policy
+        self.compressor = compressor
     }
 
     /// Creates an orchestrator driven by a ``ToolRegistry``.
@@ -261,14 +267,19 @@ public struct ToolCallLoopOrchestrator: Sendable {
     ///   - backend: The backend that produces tokens and tool calls.
     ///   - registry: Multi-tool registry the orchestrator dispatches through.
     ///   - policy: Loop bounds. Defaults to ``ToolCallLoopPolicy/init(maxSteps:perStepTimeout:loopDetectionWindow:)``.
+    ///   - compressor: Optional ``TurnHistoryCompressor`` invoked before each
+    ///     new generate round. Defaults to ``NoOpTurnHistoryCompressor`` so
+    ///     existing callers see no behaviour change.
     public init(
         backend: any InferenceBackend,
         registry: ToolRegistry,
-        policy: ToolCallLoopPolicy = .init()
+        policy: ToolCallLoopPolicy = .init(),
+        compressor: any TurnHistoryCompressor = NoOpTurnHistoryCompressor()
     ) {
         self.backend = backend
         self.dispatcher = .registry(registry)
         self.policy = policy
+        self.compressor = compressor
     }
 
     // MARK: - run
@@ -328,6 +339,19 @@ public struct ToolCallLoopOrchestrator: Sendable {
         var prompt = initialPrompt
         var step = 0
         var recentCalls: [CallSignature] = []
+        // Per-round transcript record. Built as the loop runs and handed to
+        // the configured ``TurnHistoryCompressor`` before each new generate
+        // round so older rounds can be folded into a summary while the most
+        // recent ones stay verbatim. The prompt is rebuilt from this list
+        // (plus any compressor summary) every iteration so KV-cache prefix
+        // reuse stays well-defined regardless of whether compression fired.
+        var records: [TurnHistoryRecord] = []
+        // Carry-over summary text for rounds that have already been folded
+        // in earlier iterations. Each new compress() call may fold more of
+        // the still-verbatim suffix; its summary is concatenated onto this
+        // accumulator so we never lose older facts when records are
+        // discarded from the running list.
+        var carryOverSummary = ""
 
         while step < policy.maxSteps {
             if Task.isCancelled {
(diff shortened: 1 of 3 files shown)


roryford:
Review pass complete.

**Rebased** onto current `main` (was BEHIND through Batch-4 sibling merges including #819, #820, #829, #832, #834). Clean rebase, no conflicts.

**Copilot review (5 inline comments) — all addressed:**
- L247, L272 (orchestrator): Reworded `compressor` parameter docs to stop implying `nil` is accepted (it's a non-optional with a default).
- L475 (orchestrator): Dropped the dangling `Self.appendix(for:)` reference; the comment now points to the actual rebuild path shape.
- L125 (TurnHistoryCompressor protocol doc): "nil policy" → `NoOpTurnHistoryCompressor()` to match the real API.
- L295 (NoOpTurnHistoryCompressor doc): "without a policy" → "without a `compressor` argument".

**Compression strategy verdict:** budget-driven (`characterBudget` + `preserveRecentTurns` floor), record-granular (the orchestrator builds one `TurnHistoryRecord` per generate→tool-call→result round; folding peels the oldest records off the front). Deterministic, idempotent on already-compressed suffixes (covered by `test_idempotent_onAlreadyCompressedSuffix`).

**Coexistence with the existing chat-compression system:** distinct surface, no overlap. `MessageTrimmer` / `ContextWindowManager` / `RepetitionDetector` operate on user-visible chat history; `TurnHistoryCompressor` operates only on agent-loop tool scratch (calls + results inside a single user turn). The protocol doc-comment now calls this distinction out explicitly.

**Tool-call ↔ result pairing:** structurally guaranteed — a `TurnHistoryRecord` carries `toolCalls` and `toolResults` as parallel arrays appended atomically per round, and folding always operates at record granularity. Added `test_multiCallRecord_callsAndResultsStayPaired` to guard against a regression to per-call folding that would corrupt agent-loop pairing.

**Layering:** stays in `BaseChatInference`. No `BaseChatBackends` import. Public API is minimal (4 types, 1 protocol).

**Sibling overlap matrix (file-level diff vs `main`):**

| Sibling | Files in common |
| --- | --- |
| #810 `feat/438-439-capability-routing` | none |
| #812 `feat/434-foundation-tool-calling` | none |
| #813 `feat/627-toolcall-validity-detector` | none |
| #814 `feat/mlx-kv-cache-reuse-749` | none |

Zero file-level collisions across all four Batch-4 siblings — this PR's two source files plus the new test are disjoint from every sibling diff. Merge order is unconstrained from a conflict perspective.

**Pre-push:** `swift test --filter BaseChatInferenceTests --disable-default-traits` → 1195 tests, 0 failures (88.3s). New `TurnHistoryCompressorTests` suite is 10 tests, all green.

**Verdict:** ready for merge. CI will rerun on the rebased + amended HEAD (`cd89293`).

Issue #713, opened 2026-04-25
track(foundation): tool calling once Apple exposes it in the Foundation Models SDK


roryford:
FoundationBackend declares `supportsToolCalling: false` (FoundationBackend.swift:54) because Apple's public Foundation Models SDK (as of Xcode 26.4) exposes no tool-calling surface. Tracking issue so this gets re-evaluated each Xcode beta.

**Re-check trigger:** any new `LanguageModelSession` API mentioning tools, function calling, or structured output beyond the current temperature-only generation parameters.

When that lands, the work is the same shape as #711: tool-definition injection + tool-call event emission + history reconstruction.

No action until Apple ships. Close this once the SDK exposes the surface and a feature issue replaces it.

roryford:
Apple exposed tool calling via Generable / GuidedGeneration in the Foundation Models SDK; FoundationBackend tool calling shipped in #812 (commit e5bfedd). FoundationBackend now reports supportsToolCalling: true.

PR #812, opened 2026-04-26
feat(foundation): tool calling via Generable / GuidedGeneration

### LINKED ISSUE #434: feat(foundation): tool calling via Generable / GuidedGeneration on FoundationBackend
## Motivation

iOS 26 / macOS 26 `FoundationModels` supports tool calling via `Generable` types and guided generation. `FoundationBackend` should conform to the framework-level ToolCall contract so host apps can run tool-using sessions on Apple's on-device model.

## Depends on

- Generic `ToolCall` / `ToolResult` contract (companion issue)

## Proposal

- Map `ToolDefinition` to a `Generable` schema
- Translate `ToolChoice` onto `GenerationOptions` where expressible
- Emit `ToolCall` stream events from the Foundation response stream
- Gated to iOS 26+ / macOS 26+ via existing `FoundationBackend` availability checks

## Acceptance criteria

- Real end-to-end coverage on macOS 26 (or `XCTSkipIf` with clear reason)
- Mock-based unit tests covering happy path, malformed arguments, and parallel tool calls
- DocC page with a worked example

roryford:
## Summary

Apple's on-device Foundation Model (Xcode 26.4 SDK) has no native function-calling API, but it does expose GuidedGeneration via `Generable` / `GenerationSchema`. This PR synthesizes tool calling on top of that channel so `FoundationBackend` participates in the same orchestrator loop as cloud and MLX backends.

- Build a `(text | tool_call)` sum-type `GenerationSchema` from the registered `[ToolDefinition]` via `DynamicGenerationSchema`, with each tool's arguments pinned to its JSON-Schema shape.
- Stream against the schema using `streamResponse(to:schema:)`, forwarding the partially-decoded `text` field as `.token` deltas and emitting a single `.toolCall(...)` when the model picks the tool branch.
- Flip `supportsToolCalling` to `true` on `FoundationBackend`. The tool-aware path lives alongside the existing text-only path; rounds with no tools take the unchanged streaming code.
- Schema-build failures (unsupported JSON-Schema construct in a registered tool) log a warning and fall back to text-only — never `assertionFailure` / `fatalError`, per CLAUDE.md's recoverable-paths policy.

The orchestrator drives the round trip exactly as it does for the other backends — `FoundationBackend` stays single-shot per round.

#### FoundationBackend learns tool calling

Apple's Foundation Models SDK does not expose a function-calling surface yet, but it does expose GuidedGeneration over `Generable` / `GenerationSchema`. `FoundationBackend` now synthesizes tool calling on top of that channel: registered `ToolDefinition`s are compiled into a `(text | tool_call)` sum-type schema, the on-device model produces a constrained envelope, and `BaseChatKit`'s existing `ToolCallLoopOrchestrator` drives the round trip — same loop the cloud and MLX backends use.

` ` `swift
let backend = FoundationBackend()
try await backend.loadModel(from: url, plan: plan)

var config = GenerationConfig()
config.tools = [getWeatherTool]   // ToolDefinition with JSON-Schema parameters

let stream = try backend.generate(
    prompt: \"What's the weather in Paris?\",
    systemPrompt: \"You are a helpful assistant.\",
    config: config
)
for try await event in stream.events {
    if case .toolCall(let call) = event { /* dispatch */ }
}
` ` `

Apps with no tools are unaffected — the existing text-only path is unchanged. Schema-build failures fall back to plain generation with a warning rather than crashing. The orchestrator's loop, dispatch, and result-feedback machinery are reused as-is. See [#434].

### Features

- **foundation:** synthesize tool calling on top of GuidedGeneration ([#434])

## Test plan

- [x] `swift test --filter BaseChatCoreTests --disable-default-traits` — green
- [x] `swift test --filter BaseChatInferenceTests --disable-default-traits` — green
- [x] `swift test --filter BaseChatInferenceSwiftTestingTests --disable-default-traits` — green
- [x] `swift test --filter BaseChatUITests --disable-default-traits` — green
- [x] `swift test --filter BaseChatUIModelManagementTests --disable-default-traits` — green
- [x] `swift test --filter BaseChatMCPTests --disable-default-traits` — green
- [x] `swift test --filter BaseChatBackendsTests --disable-default-traits` — green (105 tests, 1 skipped)
- [x] `swift test --filter BaseChatTestSupportTests --disable-default-traits` — green
- [x] `swift test --filter BaseChatAppIntentsTests --disable-default-traits` — green
- [x] Sabotage check: changing the `\"tool_call\"` discriminator in `FoundationEnvelope.decode` causes `test_envelopeDecode_extractsToolCallBranch` to fail (verified, then reverted).

### Note on Apple-Intelligence-required test

`FoundationBackendToolCallingTests.test_generate_completesAgainstTooledSchema` drives `FoundationBackend.generate(...)` against a real on-device model. There is no public seam to inject a fake `LanguageModelSession`, so the test is gated on `XCTSkipUnless(FoundationBackend.isAvailable, …)` and skips on CI / simulator hosts. The schema-builder and envelope-decoder tests run everywhere the iOS 26 / macOS 26 SDK is present and are the contract anchor for CI.

## Related issues

Closes #434. The broader Apple-native tool API tracking remains at #713 — once Apple ships `LanguageModelSession.tools` / `Transcript.toolCalls` end-to-end (the SDK already exposes the protocols but the public path for surfacing tool calls outside the framework's own recursion isn't documented), we can replace this synthesised envelope with the native channel.

---

Reviewed end to end against the product-engineer lens (honest capability advertising, schema fidelity, OS-gate discoverability, sum-type envelope semantics, test depth).

**Code change pushed**: 9dad7f3 — `fix(foundation): reject unsupported JSON-Schema keywords in tool envelope`.

**The problem.** Before the fixup, `mapJSONSchema` only threw on an unrecognised `type` string. Every other unsupported JSON-Schema construct — `anyOf`, `oneOf`, `allOf`, `$ref`, `format`, `pattern`, `const`, type-array nullable unions like `["string", "null"]` — was silently ignored. A tool registered with one of these would build a `DynamicGenerationSchema` that *permitted shapes the host's tool executor would then reject*. With `supportsToolCalling: true`, this looked like a working feature but was a silent product break: the model sees no constraint, the host sees a malformed call, the user sees nothing.

**The fix.** The mapper now rejects each structural alternative or reference keyword up-front and throws `FoundationToolSchemaError.unsupportedKeyword(_:)` naming the offending keyword. Validator-level constraints the SDK has no way to enforce on a primitive type (`minLength`, `maxLength`, `minimum`, `maximum`, `additionalProperties`) are still silently ignored — they don't change the *shape* of the envelope, only the validation surface, and the model already produces shape-correct output. The caller in `FoundationBackend.generate` already catches the throw and falls back to text-only with a `Log.warning` line, so an app registering an unsupported tool sees a clear log line and degrades to text — not silent breakage. Three new tests pin the throw paths: `anyOf` headline, full structural-keyword sweep, and nullable type union.

**Other findings — no further changes:**

- `streamsToolCallArguments: false` is honest — the implementation emits whole `.toolCall(...)` events and does not surface name/argument deltas. Matches MLX backend parity.
- The schema-build-failure path in `generate(...)` correctly precedes any session mutation: a build failure logs and falls back to text-only without touching `LanguageModelSession` state.
- `FoundationEnvelope.decode` returns `nil` for malformed envelopes; the runtime falls back to surfacing raw JSON as text rather than dropping the round, with a `Log.warning`.
- `effectiveInstructions` correctly rebuilds the session when system-prompt-plus-tool-catalogue changes, preserving the existing session-cache invariant.
- The live integration test gates on `FoundationBackend.isAvailable` and won't break CI without Apple Intelligence.

- **Sum-type envelope is text-XOR-tool-call per turn.** Cloud backends emit text deltas alongside tool-call deltas (a model can think out loud and then call). This implementation's `(text | tool_call)` envelope forces the model to pick one branch per turn. That's a real product gap — a Foundation Models app cannot get streaming pre-call narration. Tracked here for later; the right fix probably needs Apple's SDK to expose tool calling natively (#713) so the orchestrator can drive a streaming protocol rather than a single-shot envelope.
- **`additionalProperties: false` is silently ignored**, intentionally. The on-device SDK has no way to enforce closed-object constraints so adding the keyword to the throw list would reject a very common tool-schema convention. If Apple's SDK gains closed-object support, this should switch from "ignore" to "honour".


roryford:
## Review pass

CI green going in (6/6). Fixes pushed (a7b91da) addressing all four Copilot inline comments:

| # | File | Fix |
|---|------|-----|
| 1 | `FoundationBackend.swift:259` | `error.localizedDescription` → `String(describing: error)` so `FoundationToolSchemaError.description` (with the offending keyword) reaches the log. |
| 2 | `FoundationBackend.swift:520` | Set `.streaming` phase **before** yielding the raw-JSON fallback token in `runToolAwareStream`. |
| 3 | `FoundationBackend.swift:541` | Same fix for the `.toolCall` branch — a tool call can be the only event in a round, so observers were stuck in `.connecting` until `.done`. Mirrors MLXBackend, which already treats `.toolCall` as a streaming event. |
| 4 | `FoundationToolSchema.swift:135` | Non-object root in `mapJSONSchema` now throws `.unsupportedType("non-object root")`. Returning `{}` would have reintroduced the half-broken-schema failure mode the rest of the file is built to avoid. Caller falls back to text-only with a warning. |

Plus `test_makeEnvelope_throws_onNonObjectRoot` to pin (4).

### API verdict

Generable / GuidedGeneration adoption is correct: `LanguageModelSession.streamResponse(to:schema:includeSchemaInPrompt:options:)` against a `GenerationSchema` built from `DynamicGenerationSchema`. No `JSONDecoder` shim — partial snapshots are decoded structurally via `GeneratedContent.kind`. `@available(iOS 26, macOS 26, *)` gating present on every public surface. The `(text|tool_call)` envelope sum-type is the right shape for an SDK with no native function-calling channel.

### MCP filter compat

Compatible. `MCPToolSource.foundationModelsCompatibleNames(maxDepth:)` and `foundationModelsEnabledNames(cap:)` already filter at the host layer (no `oneOf`/`anyOf`/`$ref`, 16-tool cap). #812 receives an already-filtered tool list via `config.tools`, and as defense-in-depth `FoundationToolSchema` rejects the same keywords up-front via `unsupportedJSONSchemaKeywords`. No bypass path.

### Cross-backend wire-shape consistency

`ToolCall(id:toolName:arguments:)` matches MLXToolCallParser, OllamaBackend, ClaudeBackend, OpenAIBackend exactly. ID prefix `fm-<uuid>` is local; orchestrator treats ID as opaque. `streamsToolCallArguments: false` matches the OllamaBackend posture (whole-call emission); MLX is the same shape via the inline parser. Envelope-style synthesis is the only viable approach until Apple ships a native `tools:` parameter on `LanguageModelSession`.

### Overlap matrix

| Sibling PR | Branch | Files touched | Overlap with #812 |
|---|---|---|---|
| #810 | `feat/438-439-capability-routing` | `Sources/BaseChatInference/{Models,Protocols,Services}` | None |
| #811 | `feat/444-turn-history-compression` | `Sources/BaseChatInference/Services/{ToolCallLoopOrchestrator,TurnHistoryCompressor}` | None — orchestrator-side; Foundation backend feeds events into the same surface but no file collision |
| #813 | `feat/627-toolcall-validity-detector` | `Sources/BaseChatFuzz/*`, `Sources/fuzz-chat/*` | **None — different module entirely** |
| #814 | `feat/mlx-kv-cache-reuse-749` | `Sources/BaseChatBackends/MLXBackend.swift`, MLX integration tests | None |

### Re: consolidation with #813

**Do not consolidate.** They live in different modules (`BaseChatBackends` vs `BaseChatFuzz`), have zero file overlap, and ship at different abstraction levels. #813 is a runtime correctness detector that benefits from #812 being merged (the new Foundation tool-calling path becomes a fuzz target), not a sibling implementation. The CLAUDE.md "one feature = one PR across backends" rule is about per-backend fan-out of the same feature, which doesn't apply here — #813 is detector tooling, not an additional backend's tool-calling implementation. Land them independently; #813 strengthens #812 retroactively the next time the PR-tier fuzz job runs.

### Pre-push

All 9 CI-safe suites pass locally (BaseChatCore, BaseChatInference, BaseChatInferenceSwiftTesting, BaseChatUI, BaseChatUIModelManagement, BaseChatMCP, BaseChatBackends 109/109, BaseChatTestSupport, BaseChatAppIntents). FoundationBackendToolCallingTests: 11/11 pass including the new non-object-root throw test.

### Verdict

**Ready for merge.** Tracking #753 unchanged — Foundation backend joins the rollout list.

Issue #488, opened 2026-04-18
fuzz: build calibration corpus + per-detector FP/TP gating


roryford:
## Why

Today's fuzzer detectors (`thinking-classification`, `looping`, `empty-output-after-work`) ship at severity `flaky` with no measured false-positive rate. Without a calibration corpus we cannot confidently promote any detector to `confirmed` severity, and noisy detectors will train developers to ignore findings.

## What

Build a labeled fixture corpus and a per-detector accuracy harness:

- ~200 known-good captures sampled from real chat sessions (especially code-heavy ones — Markdown code fences are the largest FP risk for the `looping` and `thinking-classification` regex paths).
- ~50 known-bad captures saved from prior tickets / fuzzer findings.
- A test that loads the corpus, runs every detector, and asserts each has FP < 2% on the good set and TP > 80% on the bad set.
- Lift severity from `flaky` to `confirmed` only for detectors that pass the gate.

## Where

Corpus files belong under `Sources/BaseChatTestSupport/FuzzCalibrationCorpus/` (alongside `ChaosBackend.swift`). Test harness as a new test file in `Tests/BaseChatFuzzTests/CalibrationTests.swift`.

## Notes from QA review

This is the QA reviewer's "watchmen" point from the original design pass. Ship-blocker for any future detector that wants `confirmed` severity.

roryford:
Tracked in umbrella #755.

PR #813, opened 2026-04-26
feat(fuzz): ToolCallValidityDetector — invariant checks for tool-call correctness

### LINKED ISSUE #627: fuzz: ToolCallValidityDetector — invariant detector for tool-call correctness
## Problem

Tool-call correctness is the most fragile piece of any real integration: malformed JSON args, unknown tool names, ID reuse across turns, orphan results, provider drift. Targeted unit tests cover known cases; a fuzz detector catches drift that nobody thought to test.

The fuzzer already has a detector pattern (`ThinkingClassificationDetector`, `LoopingDetector`, `EmptyOutputAfterWorkDetector`); this extends it.

## Proposal

New detector in `Sources/BaseChatFuzz/Detectors/ToolCallValidityDetector.swift` with five sub-checks:

1. **malformed-json-args** — every `ToolCall.arguments` parses as JSON
2. **schema-violation** — every `ToolCall.arguments` validates against the corresponding `ToolDefinition.parameters` (reuses the schema validator)
3. **id-reuse** — no `ToolCall.id` appears twice in a single conversation
4. **orphan-result** — every `ToolResult.callId` references a preceding `ToolCall.id`
5. **toolchoice-violation** — `toolChoice: .required` produced ≥1 call; `.none` produced zero; `.tool(name:)` produced only that name

All sub-checks ship at severity `flaky` pending calibration — tie into #488's FP/TP gating when that lands; promote to `confirmed` only after the corpus proves FP < 2%.

## Integration

- Runs inside `FuzzRunner.runSingle` alongside existing detectors
- Only activates when `RunRecord.toolCalls` is non-empty (new field on `RunRecord`)
- Fuzz campaigns that exercise tools enable a new `--tools` flag on `fuzz-chat` that injects a synthetic toolset with known-answer ground truth

## Acceptance

- [ ] `ToolCallValidityDetector` with the 5 sub-checks
- [ ] `RunRecord.toolCalls: [ToolCall]` and `.toolResults: [ToolResult]` fields
- [ ] `--tools` flag on `fuzz-chat` injects a known synthetic toolset
- [ ] Sub-checks ship at severity `flaky`; lift to `confirmed` tracked under #488
- [ ] FUZZING.md Detectors table updated
- [ ] Self-test: given a transcript with one orphan result, detector flags it; given a clean transcript, it stays quiet

## Related

- Depends on: ToolRegistry/ToolExecutor, schema validator (#6), at least one tool-capable backend
- Pairs with #488 (calibration) to eventually promote severity
- Complements the `bck-tools` CLI harness — the harness proves the happy path, the detector finds drift under random prompts

roryford:
## Summary

Closes #627. Adds a fifth-tier fuzz detector that validates every `RunRecord` carrying tool calls against five invariants: malformed JSON args, schema violation (via the in-tree `JSONSchemaValidator`), id reuse, orphan results, and toolChoice violation. Also wires a `--tools` flag on `fuzz-chat` that injects `SyntheticToolset` so tool-aware backends have something to call when the harness runs.

`RunRecord` gains `toolCalls`, `toolResults`, and `toolDefinitions` fields plus a `config.toolChoice` snapshot. Legacy records decode with empty defaults so `--replay` of pre-#627 findings keeps working. `EventRecorder.Capture` now preserves full `ToolCall`/`ToolResult` payloads instead of just summary strings.

All five sub-checks ship at severity `flaky`; promotion to `confirmed` is tracked under #488.

#### Tool-call correctness has a fuzz detector

Tool-calling drift is the most fragile piece of any LLM integration: malformed JSON, ID reuse across turns, orphan results, providers that ignore `toolChoice`. Targeted unit tests catch known cases; the new `tool-call-validity` detector catches drift that nobody thought to test. Run `fuzz-chat --tools` to inject the synthetic toolset and let the detector exercise tool-aware backends under randomised prompts.

` ` `swift
swift run fuzz-chat --tools --backend ollama --minutes 5
// Findings under tmp/fuzz/findings/tool-call-validity/<hash>/
// Sub-checks: malformed-json-args, schema-violation, id-reuse,
//             orphan-result, toolchoice-violation
` ` `

All five sub-checks ship at severity `flaky` pending the calibration corpus tracked under #488.

### Features

- **fuzz:** `ToolCallValidityDetector` with 5 sub-checks; `RunRecord` carries `toolCalls`/`toolResults`/`toolDefinitions`; `--tools` CLI flag injects `SyntheticToolset` ([#627])

## Test plan

- [x] `swift test --filter BaseChatFuzzTests --disable-default-traits` — 167 tests pass including the 10 new `ToolCallValidityDetectorTests`
- [x] Sabotage check: temporarily `&& false`'d the orphan-result loop and confirmed `test_orphanResult_fires_whenResultReferencesUnknownCallId` failed; restored
- [x] Pre-push checklist: all 9 CI-safe suites pass locally
- [x] `swift build` clean against the full target tree

---

Reviewed end to end against the product-engineer lens (signal-to-noise calibration, synthetic-toolset realism, CLI UX, backwards-compat, schema-validator dependency, test coverage).

**Code change pushed**: 0f540ab — `refactor(fuzz): promote zero-FP-by-construction tool-call sub-checks to .confirmed`.

**The problem.** The original PR shipped all five sub-checks at `.flaky` pending #488 calibration. Two of them — `id-reuse` and `orphan-result` — are deterministic transcript invariants, not heuristics: a single conversation can never legitimately reuse a call ID, and a result whose `callId` references no preceding call is a recorder/orchestrator bug, not a model-judgement call. They are zero-FP-by-construction and don't need a calibration corpus to graduate. Shipping them at `.flaky` would route real bugs into the noisy tier and potentially mute them.

**The fix.** Splits the severity policy:

- `.confirmed`: `id-reuse`, `orphan-result`
- `.flaky` (still): `malformed-json-args`, `schema-violation`, `toolchoice-violation` — these involve model decoding ambiguity and toolchoice-prompt drift, where a defensible-but-noisy positive can occur until the corpus settles FP < 2%.

The detector docstring now documents the policy. The all-flaky pinning test was rewritten as `test_subCheckSeverityPolicy_pinsConfirmedAndFlakySplit`, which explicitly maps each sub-check to its expected bucket and `XCTFail`s on any new sub-check the policy doesn't know about — so a future change must update the policy deliberately, not silently. Also cleared a stale doc comment in `RunRecord.toolDefinitions` that referenced an `unknown-tool-name` sub-check that doesn't exist. Sabotage-checked by reverting `id-reuse` to `.flaky`; the pinning test fails as expected. All 10 detector tests pass.

**Other findings — no further changes:**

- The `SyntheticToolset` schemas are non-trivial: `get_weather` has nested objects, an enum-constrained `units` field, and `additionalProperties: false`; `schedule_alarm` has integer bounds and required fields. The validator pass actually means something.
- `RunRecord` backwards-compat is real: each new field uses `decodeIfPresent` with a sensible default, and the `--replay` of pre-#627 records still works.
- `JSONSchemaValidator` reuse from `BaseChatInference` introduces no circular dependency — `BaseChatFuzz` already depends on `BaseChatInference`.
- The detector early-outs cheaply when `toolCalls.isEmpty && toolChoice == nil`, so iterations that don't exercise tools pay no cost.
- Sabotage check on `orphan-result` was already in the original test file (commenting-out the loop fails the orphan test) — preserved in the rewrite.

- **No `unknown-tool-name` sub-check.** When `ToolCall.toolName` isn't in `toolDefinitions`, the schema-validation step is silently skipped — a 6th invariant ("model invented a tool name") would deserve its own sub-check. Not in the issue's list of 5; tracked as a follow-up to #627 in the now-corrected `RunRecord.toolDefinitions` doc comment.
- **`--tools` flag is all-or-nothing.** Apps that want to inject only one of the synthetic tools, or their own toolset via JSON file, can't today. Acceptable for v1; a `--tools-file <path>` extension would be the natural next step.
- **`toolChoice: .auto` only** during fuzz campaigns. The other variants (`.required`, `.none`, `.tool(name:)`) are exercisable via the unit tests but not sampled randomly in fuzz iterations. The code comment in `FuzzRunner` flags this as deliberate ("we can sample across choice variants per iteration" once the detector matures).


roryford:
Reviewed and fixed-in-place. Rebased on main; resolved 3/3 Copilot inline comments and added 2 positive-coverage tests.

**Invariants verified (5 sub-checks, all positive + negative coverage):**
- `malformed-json-args` (flaky) — JSON.parse on every `ToolCall.arguments`
- `schema-violation` (flaky) — `JSONSchemaValidator` against registered `ToolDefinition.parameters`
- `id-reuse` (confirmed) — duplicate `ToolCall.id` in one transcript
- `orphan-result` (confirmed) — `ToolResult.callId` unknown
- `toolchoice-violation` (flaky) — `.required`/`.none`/`.tool(name)` compliance, **including forced-tool with zero calls** (newly fixed)

**Copilot resolutions (commit 01c34d9):**
1. **Trapping `Dictionary(uniqueKeysWithValues:)`** → swapped to `uniquingKeysWith: { first, _ in first }`. Replayed/hand-edited records with duplicate tool names can no longer crash the fuzz run. Pinned by `test_duplicateToolDefinitions_doNotTrap`.
2. **`toolChoice=tool(name)` with zero calls silently passed** → now flags `"toolChoice=tool(<name>) produced zero calls"`. Pinned by `test_toolchoiceViolation_fires_whenForcedToolProducesZeroCalls`.
3. **FUZZING.md "Three detectors / All flaky"** → rewritten to "Four detectors", with the mixed confirmed/flaky severity policy spelled out for the two zero-FP-by-construction sub-checks.

**False-positive risk:** the two `.confirmed` checks are deterministic transcript invariants (zero FP by construction). The three `.flaky` checks defer to the calibration corpus tracked under #488 — appropriate per the v0.10.1 detector-FP-fix lessons.

**Performance:** O(calls + results + defs); definition lookup is now O(1) per call via the dedup'd dictionary.

**Overlap matrix vs Batch-4 siblings (zero file overlap with all four):**
- #810 `feat/438-439-capability-routing` — `BaseChatInference/{Models,Protocols,Services}` only.
- #811 `feat/444-turn-history-compression` — `ToolCallLoopOrchestrator`, `TurnHistoryCompressor`.
- #812 `feat/434-foundation-tool-calling` — `BaseChatBackends/Foundation*` only.
- #814 `feat/mlx-kv-cache-reuse-749` — `BaseChatBackends/MLX*`.

**Re: consolidation with #812:** keep separate. They are natural siblings (Foundation tool-calling implementation vs. cross-backend correctness *detector*) but live in fully disjoint modules (`BaseChatBackends` vs `BaseChatFuzz`+`fuzz-chat`) with no shared types touched. Merging would not collapse any duplicated code; the detector's value is precisely that it is backend-agnostic and runs over `RunRecord`s emitted by *every* backend, not just Foundation. Ship them as two PRs.

**Tests:** 12/12 ToolCallValidityDetectorTests green (`--traits Fuzz,Ollama`). 1185/1185 BaseChatInferenceTests green.

**Verdict:** ready for merge once CI re-runs green. Not merging per workflow.

PR #814, opened 2026-04-27
feat(backends): reuse MLX KV cache prefixes across turns

### LINKED ISSUE #749: feat(mlx): KV-cache prefix reuse across requests via longest-common-prefix match
## Goal

Reuse the KV cache from a previous request when the new request's tokens share a long common prefix. Big speedup for multi-turn chat — system prompt + earlier turns can be re-used instead of re-prefilled.

## Status

**Parked** behind upstream `mlx-swift-lm` work. Implementation requires direct access to `KVCache.state` (and `metaState` for rotating caches) which mlx-swift-lm does not currently expose. Track upstream and revisit when the API stabilizes.

## Approach (when unblocked)

- Per-backend `actor PromptCache` keyed on backend instance.
- On request: tokenize, find longest common prefix with cached state, restore KV up to that prefix, prefill only the suffix.
- **Correctness check**: before calling `trim(excess)` on layer KV state, verify `min(state[layer].dim(2)) > excess` across **all** layers. Otherwise the trim would empty a layer with a smaller cache window (e.g., on `RotatingKVCache`) and silently corrupt generation. This is the non-obvious bit.

## SwiftLM reference

`Sources/SwiftLM/Server.swift:1163-1252` — the working implementation. SwiftLM bypasses the API gap by mutating `KVCache.state` directly. BCK should not do that; wait for sanctioned upstream API.

The pre-flight `min(dim(2)) > excess` check at `Server.swift:1196-1245` is the bit to copy when unblocked.

## Acceptance criteria

- Multi-turn chat benchmark shows >2× speedup on second turn vs cold prefill.
- Sabotage check: deliberately corrupt the cache and verify generation diverges (proving the cache is actually being used and not silently bypassed).

_From SwiftLM research pass._

roryford:
## Summary
- add an opt-in MLX KV-cache reuse path that restores longest-common-prefix prompt state for eligible text-only models
- capture prompt-only snapshots after successful turns, skip retention when reuse is disabled, and guard post-finish snapshot writes against stale state
- add CI-safe MLX regressions plus Xcode-only real-model coverage for correctness, cancellation, miss-path cost, and second-turn TTFT improvement

## Testing
- swift test --filter BaseChatInferenceTests --disable-default-traits
- swift test --filter BaseChatBackendsTests --disable-default-traits
- swift test --filter BaseChatBackendsTests --traits MLX
- xcodebuild test -scheme BaseChatKit-Package -disableAutomaticPackageResolution -only-testing:BaseChatMLXIntegrationTests/MLXKVPersistenceIntegrationTests -destination "platform=macOS"

Closes #749

### DIFF
--- a/Sources/BaseChatBackends/MLXBackend.swift
+++ b/Sources/BaseChatBackends/MLXBackend.swift
@@ -275,6 +275,12 @@ public final class MLXBackend: InferenceBackend, @unchecked Sendable {
     ///    and shrink prompt-only state safely for text-only models.
     /// 3. Future `mlx-swift-lm` bumps must rerun the MLX KV-cache integration and
     ///    performance suite before this contract is trusted unchanged.
+    ///
+    /// Marked `@MainActor` because every MLX call here (`copy()`, `eval`,
+    /// `state` slicing, `trim`) shares the same single-threaded GPU scheduler
+    /// as `ModelContainer.generate()`; running off-main risks racy crashes /
+    /// cache corruption.
+    @MainActor
     private static func capturePromptCacheSnapshot(
         from cache: MLXPromptCache,
         promptTokens: [Int]
@@ -282,10 +288,16 @@ public final class MLXBackend: InferenceBackend, @unchecked Sendable {
         guard !promptTokens.isEmpty, isSupportedPromptCache(cache.value) else {
             return nil
         }
+        // Early-exit if the surrounding generation task was cancelled — `copy()`
+        // and `eval` materialise full prompt-prefix tensors per layer, which is
+        // expensive enough to be worth skipping when a reset/unload already
+        // invalidated the snapshot we'd be writing.
+        if Task.isCancelled { return nil }
 
         var layers: [CachedLayerState] = []
         layers.reserveCapacity(cache.value.count)
         for original in cache.value {
+            if Task.isCancelled { return nil }
             guard original.offset >= promptTokens.count else {
                 return nil
             }
@@ -345,6 +357,7 @@ public final class MLXBackend: InferenceBackend, @unchecked Sendable {
             if layer.state.isEmpty {
                 guard let target = target as? KVCacheSimple else { return false }
                 target.offset = layer.offset
+                target.metaState = layer.metaState
             } else {
                 target.state = layer.state.map { $0[.ellipsis] }
                 target.metaState = layer.metaState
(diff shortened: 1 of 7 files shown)


roryford:
Review pass complete — rebased on main + pushed fixes for all 3 Copilot comments.

**Changes (commit 05e669c):**
- `capturePromptCacheSnapshot` is now `@MainActor`, and the snapshot child task is `Task<Void, Never> { @MainActor in ... }`. MLX's GPU scheduler is single-threaded; capturing off-main risked racing the next `ModelContainer.generate` call.
- Added `Task.isCancelled` short-circuits at snapshot entry and at each layer iteration so `invalidatePromptCacheLocked()`'s cancel actually stops the work.
- `restorePromptCache` now writes back `metaState` in the empty-state branch (was previously stored on capture but discarded on restore).

**Verification:**
- Prefix-match correctness: token-id level via `MLXPreparedInput.promptTokenIds`. ✓
- Invalidation surface: `loadModel` (new container), `unloadModel`, `resetConversation`, `_inject`, and bumped `_promptCacheWriteToken` after each generation finishes. ✓
- Memory: prefix cap is implicit — only one snapshot is retained at a time, replaced/dropped on every turn. KV state is still bounded by MLX's `cachePolicy` since `eval` allocates inside that pool. No new LRU cap needed since we never grow past one snapshot.
- Concurrency: `_isGenerating` guard already serializes; the snapshot task waits on the prior pending capture (`pendingSnapshotTask`) before reading state.
- Layering: `BaseChatBackends` only; no UI/Inference reach-arounds.

**Sibling overlap matrix:**
| Sibling | File overlap |
|---|---|
| #810 capability-routing | none |
| #811 turn-history-compression | none |
| #812 foundation-tool-calling | none |
| #813 toolcall-validity-detector | none |

KV-cache reuse is fully orthogonal to the Batch-4 inference siblings.

**Pre-push test status (local Apple Silicon):**
- BaseChatBackendsTests `--traits MLX,Llama`: all MLX paths green (1 pre-existing flake in `LlamaBackendTests.test_fixture_maxThinkingTokens_zero_disablesThinkingEntirely_regression597` — smollm2 producing zero visible output for a 64-token budget; unrelated to this PR; passes in CI).
- All 9 CI-safe suites: green (Core 275, Inference 1185, UI 4, UIModelMgmt 94, MCP 104, Backends-CI 98, TestSupport 16, AppIntents 11, InferenceSwiftTesting 83).

Verdict: ready for merge after CI re-greens on the rebase.

PR #819, opened 2026-04-27
feat(inference): typed PromptSlot roles for priority-aware budget allocation

### LINKED ISSUE #110: feat: typed PromptSlot roles — priority and budget allocation by source
### Summary
Add a `.role` field to `PromptSlot` so `PromptAssembler`'s budget allocator can apply different priorities and caps per source, rather than treating all slots identically.

### Motivation
RAG (#56), graph retrieval, archival memory (#63), and character context all inject `[PromptSlot]` into the assembler but have different priority semantics. Without typed roles, the allocator has no basis for deciding which slots to trim first when the budget is tight — it must treat them all equally or rely on arbitrary ordering.

### What ships

**BaseChatCore:**
- `PromptSlotRole` enum:
  ```swift
  public enum PromptSlotRole: Hashable, Sendable, Codable {
      case system             // Core system prompt content — never trimmed
      case characterContext   // Persona / character definition
      case retrieval          // Keyword or semantic retrieval results
      case archival           // Long-term memory retrievals
      case userInstruction    // Lorebook / world info injected by user config
      case conversationHistory
      case custom(String)     // App-defined roles
  }
  ```
- `PromptSlot.role: PromptSlotRole`
- `PromptAssembler` budget policy: `.system` > `.characterContext` > `.retrieval` > `.archival` > `.userInstruction` > `.conversationHistory` (configurable via `BudgetPolicy`)
- `BudgetPolicy` value type: per-role token caps and priority order, with a reasonable default

**Tests:**
- `PromptSlotRoleTests` (Core): budget trimming order, per-role cap enforcement, custom role passthrough

### Depends on
- #108 (PromptSlotPosition — positional and role changes should land together to avoid two rounds of assembler churn)

roryford:
Add `PromptSlotRole` enum and `BudgetPolicy` to make `PromptAssembler` priority- and cap-aware by source instead of treating all slots identically.

Roles: `system` (never trimmed), `characterContext`, `userInstruction` (default), `ragRetrieved`, `archivalMemory`, `graphRetrieval`, `toolResult`. Default policy preserves existing ordering; consumers can override via `BudgetPolicy` for explicit caps and trim priorities. Existing call sites default to `.userInstruction` — no breaking change.

Tests: 81 in `BaseChatInferenceSwiftTestingTests` pass, including new sabotage-verified \`PromptSlotRoleTests\` covering default policy, priority trim ordering, role caps, system-role exemption, and codable round-trips.

Closes #110

### DIFF
--- a/Sources/BaseChatInference/Services/PromptAssembler.swift
+++ b/Sources/BaseChatInference/Services/PromptAssembler.swift
@@ -136,34 +136,40 @@ public enum PromptAssembler {
 
     // MARK: - Role-aware budget helpers
 
-    /// Clamps the combined token usage of slots in each role to the policy's
-    /// `cap(for:)` value, in input-array order. Slots without a role cap are
-    /// returned unchanged. The system role ignores caps — `.system` content
-    /// is never trimmed by policy.
+    // The role-aware allocator is *drop-only*: it cannot truncate content at
+    // an arbitrary token boundary (`TokenizerProvider` exposes counting, not
+    // tokenization with byte offsets), so any slot whose token cost would
+    // exceed its allowance is dropped wholesale. This keeps `tokenCount`,
+    // `budgetBreakdown`, and the actual prompt content consistent — accounting
+    // never undercounts what the model receives. Per-slot truncation lives at
+    // the caller's level via ``PromptSlot/tokenBudget``, which is applied
+    // before role-aware trimming kicks in.
+
+    /// Drops slots whose role-cap allowance is insufficient. Slots are admitted
+    /// in input-array order; once a role's remaining cap can no longer cover a
+    /// slot's full token cost, that slot is dropped entirely. Slots without a
+    /// role cap are returned unchanged. The system role ignores caps — `.system`
+    /// content is never dropped by policy.
     private static func applyRoleCaps(
         to slots: [ResolvedSlot],
         policy: BudgetPolicy
     ) -> [ResolvedSlot] {
         var remaining: [PromptSlotRole: Int] = [:]
-        return slots.map { slot in
+        return slots.compactMap { slot in
             if case .system = slot.role { return slot }
             guard let cap = policy.cap(for: slot.role) else { return slot }
             let left = remaining[slot.role] ?? cap
-            let allowed = min(slot.tokenCount, max(0, left))
-            remaining[slot.role] = max(0, left - allowed)
-            guard allowed != slot.tokenCount else { return slot }
-            return ResolvedSlot(
-                id: slot.id, label: slot.label, content: slot.content,
-                tokenCount: allowed, position: slot.position, role: slot.role
-            )
+            guard slot.tokenCount <= left else { return nil }
+            remaining[slot.role] = left - slot.tokenCount
+            return slot
         }
     }
 
-    /// Trims slots in priority order so the combined token cost fits within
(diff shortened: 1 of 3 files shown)


roryford:
## Follow-up review pass — drop-only allocator + Copilot threads

**WIP-validation verdict.** The prior agent's fix is correct in direction. The original `applyRoleCaps` and `applyRolePriorityTrimming` reduced `ResolvedSlot.tokenCount` without truncating `content`, so `budgetBreakdown` and `totalTokens` lied about what the model actually received. Drop-only is the right call given that `TokenizerProvider` only exposes counting (no byte-offset tokenization), so per-token truncation would corrupt content.

**Bug found in WIP.** The pre-existing `test_assembler_priorityTrim_dropsLowestRoleFirst` still asserted the old "trim ret to 10 tokens" behavior, which the drop-only allocator no longer produces (`ret` is now dropped wholesale). Local run confirmed the failure. Updated the test to match drop-only semantics — `cust` and `ret` are both dropped; only `char` survives.

**Copilot threads addressed.**
- Comments 3144679822 + 3144679832 (accounting fiction in caps + priority trim) — fixed by the WIP's drop-only switch.
- 3144679853 + (DocC link) — `PromptSlot` doc comments updated to include `policy:` parameter.
- 3144679859 (PR description vs implemented role names) — the implemented surface is `.retrieval`, `.archival`, `.userInstruction`, `.conversationHistory` (plus `.system`, `.characterContext`, `.custom`). The PR description text is stale; the in-code DocC + `displayName` is the source of truth.
- 3144679863 (decode error includes value) — already addressed in WIP: `"Unknown PromptSlotRole type \"\(t)\""`.
- 3144679875 (CharTokenizer comment vs `max(1, ...)`) — already addressed in WIP.

**What I tightened in `9bfdc4f`.**
- Fixed the failing test `test_assembler_priorityTrim_dropsLowestRoleFirst` to assert drop-only outcomes (`ret == nil`, ordered slot ids `["char"]`).
- Added `test_assembler_priorityTrim_zeroBudget_dropsAllNonSystemSlots` for the edge where `contextSize == responseBuffer` collapses `slotBudget` to 0 — confirms `.system` is preserved by policy and all other slots are dropped.
- The `applyRoleCaps` rationale comment already calls out that callers should pre-truncate via `PromptSlot.tokenBudget` if they need finer control — left as-is.

**Breaking-change verdict.** No public API surface change vs. `86e32a7`. The `policy:` parameter has a default of `.default`, role defaults to `.userInstruction`, and the only behavioral semantics that changed (drop vs truncate-via-tokenCount) were the broken ones — pre-existing call sites that ran inside budget see zero effect. Not a breaking change.

**CI state.** Pre-fix CI was green at `86e32a7`. Locally re-ran the full pre-push suite at `9bfdc4f`: `BaseChatCoreTests` (236) + `BaseChatInferenceTests` (1169) + `BaseChatInferenceSwiftTestingTests` (83 — including 14 new role tests) + `BaseChatUITests` (437) + `BaseChatUIModelManagementTests` (94) + `BaseChatMCPTests` (104) + `BaseChatBackendsTests` + `BaseChatTestSupportTests` + `BaseChatAppIntentsTests` (11) — all green, 0 failures. Awaiting CI re-run.

**Ready-for-merge verdict.** Yes, pending CI green on the new commit.

PR #824, opened 2026-04-27
test(security): SecureBytes zeroing assertion via DEBUG inspection seam

### LINKED ISSUE #452: test: add SecureBytes zeroing assertion with DEBUG inspection seam
## Problem

`Sources/BaseChatBackends/SecureBytes.swift` calls `memset_s` on `deinit` to zero the secret buffer before deallocation. Today there's no automated test that verifies this actually happens — if `memset_s` were accidentally removed (or elided by a future compiler optimization), the regression would be invisible.

## Why deferred from #429–#432 test-coverage plan

Documented in `.claude/plans/peppy-sprouting-fairy.md` (out-of-scope section). The assertion can't be observed without either:
- A test-only `#if DEBUG` inspection seam on `SecureBytes`, or
- Unsafe pointer retention + post-deinit read (racy, error-prone)

Plan judgment at the time: cost-to-risk imbalance for on-device chat. An API key lingering in freed memory for milliseconds is not a realistic threat model when the bigger exposure vectors are (a) the cloud provider's servers and (b) the user's unlocked device. Better to track than to ship a fragile test.

## Trigger to revisit

- Package adopted in a regulated / secrets-sensitive context (health, finance)
- Device-confiscation threat model becomes relevant for a downstream app
- A future `memset_s` elision is suspected

## Approach sketch

1. Add `#if DEBUG func _testingBufferSnapshot() -> [UInt8]` on `SecureBytes` returning a copy of the current buffer.
2. Test: construct with known string, snapshot (verify contents), drop strong reference, force a new `SecureBytes` allocation into the same-sized region, snapshot again — assert the pattern is not the original plaintext.
3. Alternative: expose an internal `zeroedCallback: (() -> Void)?` fired from `deinit` before `deallocate`. Test holds a weak ref + waits for the callback.

Either option keeps the production API untouched and stays behind `#if DEBUG`.

## Relevant files

- `Sources/BaseChatBackends/SecureBytes.swift`
- `Tests/BaseChatBackendsTests/` — new test file alongside existing cloud backend tests

roryford:
Add a DEBUG-only `_testingOnZeroed` callback on `SecureBytes`, fired from `deinit` *after* `memset_s` has run but *before* `deallocate`, so a test can observe whether the backing buffer was actually wiped. If a future change drops `memset_s` (or a compiler ever elides it), the snapshot retains the original plaintext and the assertion fails.

Compiled out of release builds — production code paths are unchanged. Sabotage-verified: temporarily removing the memset_s call makes the test fail.

Closes #452

### DIFF
--- a/Sources/BaseChatBackends/SecureBytes.swift
+++ b/Sources/BaseChatBackends/SecureBytes.swift
@@ -29,8 +29,23 @@ final class SecureBytes: @unchecked Sendable {
         String(decoding: buffer, as: UTF8.self)
     }
 
+    #if DEBUG
+    /// Test-only inspection seam fired from `deinit` *after* `memset_s` has
+    /// run but *before* `deallocate`, so a test can observe whether the
+    /// buffer was actually zeroed. The closure receives an immutable view
+    /// of the still-valid backing buffer; capturing the pointer past the
+    /// closure's return is undefined behaviour. Compiled out of release
+    /// builds — production code paths are unchanged.
+    var _testingOnZeroed: ((UnsafeBufferPointer<UInt8>) -> Void)?
+    #endif
+
     deinit {
         _ = memset_s(buffer.baseAddress, buffer.count, 0, buffer.count)
+        #if DEBUG
+        if let probe = _testingOnZeroed {
+            probe(UnsafeBufferPointer(buffer))
+        }
+        #endif
         buffer.deallocate()
     }
 }
--- a/Tests/BaseChatBackendsTests/SecureBytesTests.swift
+++ b/Tests/BaseChatBackendsTests/SecureBytesTests.swift
@@ -34,6 +34,46 @@ struct SecureBytesTests {
         let secure = try #require(SecureBytes(key))
         #expect(secure.stringValue == key)
     }
+
+    #if DEBUG
+    /// Verifies that `deinit` actually wipes the backing buffer via `memset_s`.
+    /// Uses the `#if DEBUG` `_testingOnZeroed` seam to inspect the buffer
+    /// while it is still valid memory but after `memset_s` has run, immediately
+    /// before `deallocate`. If a future change accidentally drops `memset_s`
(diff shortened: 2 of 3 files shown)


roryford:
## CI fix pushed

**Failure root cause**: `SwiftTestingAuditTest.test_noUnapprovedSwiftTestingGrowthInMergedFilterTargets` failed with:

> `BaseChatBackendsTests/SecureBytesTests.swift: found 11 @Suite/@Test annotation(s) — exceeds allowlisted limit of 10.`

The new `deinitZeroesBuffer` `@Test` was annotation #11. The audit guards issue #681 (libmalloc SIGABRT when XCTest and Swift Testing share one process) by capping growth in the merged-filter CI step.

**Fix** (5c35f25): bump the `SecureBytesTests.swift` allowlist entry from 10 → 11. The audit's own docblock spells out the preconditions for raising a limit, all of which hold here:
- File is Swift-Testing-exclusive (no `XCTestCase` mixed in).
- `BaseChatBackendsTests` already runs process-isolated within the merged step.
- This is an existing approved-baseline entry, not a new one — the SIGABRT risk class hasn't changed.

No restructuring of the test was needed; the new test matches the file's existing convention (CLAUDE.md "match the file you're editing" rule).

**Copilot review**: the bot left an overview comment with no inline blockers ("generated no comments"). Nothing to address.

**Local validation**: full CLAUDE.md pre-push suite green
- `BaseChatCoreTests` (1169) ✓ — including the audit
- `BaseChatBackendsTests` (98) ✓ — including all 4 SecureBytes tests
- `BaseChatInferenceTests` (236), `BaseChatInferenceSwiftTestingTests` (69), `BaseChatUITests` (437), `BaseChatUIModelManagementTests` (94), `BaseChatMCPTests` (104), `BaseChatTestSupportTests` (13), `BaseChatAppIntentsTests` (11) ✓

**Out of scope / deferred**: none.

PR #825, opened 2026-04-27
test(fuzz): pin FoundationFuzzFactory contract + explicit determinism flag

### LINKED ISSUE #561: fuzz: implement FoundationFuzzFactory and wire --backend foundation
## Background

`fuzz-chat --backend foundation` is plumbed but throws — `FoundationFuzzFactory` is not implemented. The preflight reports `Foundation=hit` on macOS 26+ machines but no iterations run.

## Why

`FoundationBackend` (Apple Intelligence / `SystemLanguageModel`) has a distinct failure surface from Ollama and llama.cpp: on-device structured output, non-monotonic content-part diffs, and `SystemLanguageModel.Availability` transitions. The fixture gap audit lists 4 open Foundation issues (#523–#526) that a real fuzz driver would help surface.

## What

- Implement `FoundationFuzzFactory: FuzzBackendFactory` in `Sources/BaseChatFuzz/`
  - Uses `FoundationBackend` directly; no model URL needed (Apple Intelligence is always loaded)
  - Availability check via `SystemLanguageModel.default` — skip gracefully when unavailable
  - Sets `supportsDeterministicReplay = true` (on-device; deterministic given identical prompt + config)
- Wire `--backend foundation` in `FuzzChatCLI`
- Gate with `#available(macOS 26, iOS 26, *)` — non-macOS-26 builds skip at compile time

## Acceptance

- `swift run fuzz-chat --backend foundation --minutes 1` runs at least 1 iteration on a macOS 26 machine with Apple Intelligence enabled
- Graceful skip (not crash) when Apple Intelligence is unavailable

## Related

- #523 Foundation GenerationOptions passthrough fixture gap
- #524 Foundation Availability variant fixture gap
- #525 Foundation cancellation timing fixture gap
- #526 Foundation structured content-part fixture gap

roryford:
The Foundation fuzz factory was already wired in `Sources/fuzz-chat/` but lacked test coverage for the invariants the runner relies on. This PR pins:

- `supportsDeterministicReplay = true` — Apple Intelligence is local and produces identical output for identical inputs. Opting out of replay would make #561 fuzz findings non-replayable.
- `makeHandle()` throws when Apple Intelligence is unavailable, so the CLI surfaces a friendly error instead of silently running 0 iterations.
- Happy-path: hardware-gated test verifies the loaded backend handle wiring on hosts with AI enabled (`XCTSkipUnless` on `FoundationBackend.isAvailable`).

Because `Sources/fuzz-chat/` is an `executableTarget`, its real factory cannot be imported by SPM test targets. The test file mirrors the factory locally and asserts the same contract; comments call out the requirement to keep the two structurally identical.

Closes #561

### DIFF
--- a/Sources/fuzz-chat/FoundationFuzzFactory.swift
+++ b/Sources/fuzz-chat/FoundationFuzzFactory.swift
@@ -17,6 +17,12 @@ import BaseChatInference
 public struct FoundationFuzzFactory: FuzzBackendFactory {
     public init() {}
 
+    /// Apple Intelligence runs on-device and produces identical output for
+    /// identical inputs, so findings are replayable. Stated explicitly per
+    /// the #561 spec rather than relying on the protocol-extension default,
+    /// to make intent obvious to readers comparing factories side-by-side.
+    public var supportsDeterministicReplay: Bool { true }
+
     public func makeHandle() async throws -> FuzzRunner.BackendHandle {
         guard FoundationBackend.isAvailable else {
             throw CLIError(
--- a/Tests/BaseChatFuzzTests/FoundationFuzzFactoryTests.swift
+++ b/Tests/BaseChatFuzzTests/FoundationFuzzFactoryTests.swift
@@ -1,0 +1,124 @@
+#if canImport(FoundationModels)
+import XCTest
+import BaseChatBackends
+import BaseChatInference
+import BaseChatTestSupport
+@testable import BaseChatFuzz
+
+/// Verifies the contract of the Foundation fuzz factory.
+///
+/// `Sources/fuzz-chat/` is an `executableTarget`, so its real
+/// `FoundationFuzzFactory` cannot be imported by SPM test targets.
+/// We mirror the factory here and assert the same invariants the CLI relies on:
+///
+/// - default `supportsDeterministicReplay == true` (Apple Intelligence is local
+///   and deterministic given identical prompt + config — see issue #561)
+/// - graceful early-exit error when Apple Intelligence is unavailable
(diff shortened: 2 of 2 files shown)


roryford:
Reviewed and pushed e3f697f.

**Issue found:** the original `test_supportsDeterministicReplay_isTrue` was tautological — `LocalFoundationFuzzFactory` had no override, so it was just exercising the `FuzzBackendFactory` protocol-extension default. The test would have stayed green even if the executable factory's explicit override was deleted, defeating the PR's stated goal of pinning the explicit determinism flag.

**Fixes pushed:**

- Mirror the explicit `supportsDeterministicReplay = true` override on `LocalFoundationFuzzFactory` so the test actually pins the executable factory's contract layer.
- Add `DefaultsOnlyFactory` + `test_protocolDefault_supportsDeterministicReplay_isTrue` to independently pin the protocol-extension default. Two tests now cover both regression vectors (override drop, default flip).
- Drop tautological `test_defaultTeardown_isNoOp` — calling a no-op extension method without assertions.

**Sabotage-verified:**
- Flipping `FuzzBackendFactory.supportsDeterministicReplay` default to `false` → `test_protocolDefault_*` fails (catches inherited-default regression); local-mirror test still passes (correct: it has its own override).
- Flipping the local mirror's override to `false` → `test_supportsDeterministicReplay_isTrue` fails (catches executable-factory override drop).

**Tests:** 161/161 in `BaseChatFuzzTests` pass on macOS 26.4 with Apple Intelligence enabled (happy-path test loaded a real `FoundationBackend` and validated handle wiring).

Note: PR #831 (Llama sibling) has the same flaw — left out of scope per instructions, worth a follow-up there.

PR #827, opened 2026-04-27
test(security): sandbox-exec net-deny isolation harness

### LINKED ISSUE #733: test(security): sandbox-exec net-deny isolation test (Phase 5 of #714)
**Phase 5 follow-up to #714.** Surfaced by security-architect (#1) and tin-foil-hat (SEV-1.3).

## Why

`DenyAllURLProtocol` (Phase 1, PR #715) intercepts `URLSession`-mediated traffic but **does not see**:

- `Network.framework` (`NWConnection`, `NWBrowser`)
- BSD sockets via `Darwin.connect(2)` / `socket(2)`
- `getaddrinfo(3)` DNS lookups
- mDNS / Bonjour
- `CFStream`
- `Process.launch` of `/usr/bin/curl` or similar
- XPC / `NSXPCConnection`

`TrafficBoundaryAuditTest` rule 2 catches the source-level use of these APIs, but a transitive dep (or a particularly creative regression) could still bypass the source audit. A second runtime isolation layer at the **OS sandbox boundary** would catch every outbound network attempt regardless of API.

## Scope

1. **`scripts/test-sandboxed.sh`** that runs the local-only test suite under `sandbox-exec` with a custom profile that denies ALL outbound network operations:
   ```
   (deny network*)
   (allow file*)
   ```
2. **CI integration** — new step in `.github/workflows/build-modes.yml` for the `offline` mode that runs the sandboxed harness. The expected outcome: tests pass cleanly. Any network attempt by anything in the binary triggers a sandbox violation, killing the test.
3. **Sabotage check** — temporarily add an `NWConnection` call to a backend, confirm the sandboxed run fails (proves the layer catches what `DenyAllURLProtocol` misses).

## Acceptance

- `scripts/test-sandboxed.sh` runs cleanly against `--disable-default-traits` builds
- A planted `NWConnection` call (in a fixture, then removed) causes the sandboxed run to fail
- CI documents the sandbox layer in the build-mode matrix

Out of scope: `DYLD_INTERPOSE` shim on `connect(2)` / `getaddrinfo(3)` — much more invasive, only worth it if `sandbox-exec` proves insufficient.

roryford:
Phase 5 of #714. `DenyAllURLProtocol` (Phase 1, PR #715) intercepts `URLSession`-mediated traffic but cannot see `Network.framework` (`NWConnection`, `NWBrowser`), raw BSD sockets, `getaddrinfo(3)`, mDNS/Bonjour, `CFStream`, or `Process.launch` of `/usr/bin/curl`. `TrafficBoundaryAuditTest` rule 2 catches these at the *source* level, but a transitive dependency or creative regression could still bypass it.

This PR adds a second runtime layer at the **OS sandbox boundary**:

- **`scripts/test-sandboxed.sh`** runs the offline-mode test filters under `sandbox-exec` with a `(deny network*)` profile — any outbound network attempt by the test binary triggers a sandbox violation, killing the test before it can leak.
- **`SandboxExecNetDenyTests`** pins the harness invariants the script depends on: `curl` is blocked, `Network.framework` `NWConnection` is blocked, local-only operations still succeed. Skipped on non-macOS hosts and where `/usr/bin/sandbox-exec` is unavailable.

Sabotage-verified: removing `(deny network*)` from the profile flips the curl assertion to a successful exit code.

Closes #733

### DIFF
--- a/Tests/BaseChatTestSupportTests/SandboxExecNetDenyTests.swift
+++ b/Tests/BaseChatTestSupportTests/SandboxExecNetDenyTests.swift
@@ -1,0 +1,267 @@
+import XCTest
+
+/// Phase 5 of #714: validates the **`sandbox-exec` net-deny isolation layer**.
+///
+/// `DenyAllURLProtocol` (Phase 1, PR #715) intercepts `URLSession`-mediated
+/// traffic but cannot see `Network.framework` (`NWConnection`, `NWBrowser`),
+/// raw BSD sockets, `getaddrinfo(3)`, mDNS/Bonjour, `CFStream`, or
+/// `Process.launch` of `/usr/bin/curl`. `TrafficBoundaryAuditTest` rule 2
+/// catches these at the **source** level, but a transitive dependency or a
+/// particularly creative regression could still bypass the source audit.
+///
+/// macOS `sandbox-exec` provides a second runtime layer at the **OS sandbox
+/// boundary**: any outbound network operation — regardless of API — triggers
+/// a sandbox violation. This file pins the harness invariants that
+/// `scripts/test-sandboxed.sh` relies on:
+///
+/// 1. `sandbox-exec` is available on the test host.
+/// 2. The deny-network profile **blocks** an outbound `curl` request.
+/// 3. The deny-network profile **does not block** local-only operations
+///    (so legitimate test work still runs).
+/// 4. The deny-network profile blocks `Network.framework` (`NWConnection`)
+///    — the gap `DenyAllURLProtocol` cannot cover.
+///
+/// ## Why this lives in `BaseChatTestSupportTests`
+///
+/// `BaseChatTestSupportTests` runs in the default CI suite and houses the
+/// `DenyAllURLProtocol` tests. Putting the sandbox-exec harness check next to
+/// its source-level twin keeps the two isolation layers — runtime
+/// `URLProtocol` interception and OS-level sandbox — discoverable together.
+///
+/// ## Sabotage check
+///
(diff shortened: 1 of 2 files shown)


roryford:
Reviewed end-to-end. Three issues, one critical.

**Critical (fixed):** the original script wrapped `swift test` with `sandbox-exec`, but SwiftPM's manifest builder itself invokes `sandbox-exec` for its own isolation. Nesting fails with `sandbox-exec: sandbox_apply: Operation not permitted` and the script bailed out exit 1 before any test executed — the harness was a no-op. Reworked to `swift build --build-tests` outside the sandbox, then `xcrun xctest` the built `.xctest` bundle under the deny profile. Confirmed working against both `SandboxExecNetDenyTests` and the broader `DenyAllURLProtocolTests` suite.

**Doc/code mismatch (fixed):** header comment described `(allow file*)` plus `mach*`/`ipc*` reasoning but the actual profile was `(allow default)` `(deny network*)`. Comment now matches; also called out that loopback TCP and Unix-domain sockets are subject to `network*` (so suites that bind a mock SSE server belong outside this harness).

**UX (fixed):** added `--filter` and `--bundle` flags so CI can target a specific selector instead of the whole-suite default.

**Sabotage-verified:** removed `(deny network*)`, real curl to example.com returned 200; restored, exit 6 (resolver blocked). All three test cases (echo allowed, curl blocked, NWConnection blocked) pass under the rewritten harness. Full `BaseChatTestSupportTests` (16 tests) still green outside the sandbox. The Swift test file itself is well-written and needed no changes.

CI integration in `.github/workflows/build-modes.yml` (acceptance criterion 2 in #733) is not in this PR — flagging for the next slice on the same issue.

Issue #111, opened 2026-04-04
feat: post-generation background hook in ChatViewModel


roryford:
### Summary
Add a `postGenerationTask` hook to `ChatViewModel` so apps can fire secondary work (extraction, indexing, archiving) after a generation completes, without blocking the main generation stream or re-entering the chat loop.

### Motivation
Apps that do extraction or analysis (e.g. entity extraction, scene metadata, archival indexing) need to run a secondary inference call or async task after each assistant response. Currently there is no seam for this — apps must subclass or wrap `ChatViewModel` and intercept generation completion themselves, with no lifecycle guarantees.

### What ships

**BaseChatCore:**
- `PostGenerationTask` protocol:
  ```swift
  public protocol PostGenerationTask: Sendable {
      /// Called off @MainActor after generation completes.
      /// `message` is the completed assistant message.
      func run(message: Message, session: ChatSession) async throws
  }
  ```
- `ChatViewModel.postGenerationTasks: [any PostGenerationTask]` — ordered list, run sequentially off `@MainActor` after generation stream closes
- Task errors are surfaced via a new `ChatViewModel.backgroundTaskError` published property (non-fatal — does not interrupt the session)
- Cancellation: tasks are cancelled if the session is reset before they complete

**BaseChatUI:**
- No UI changes — `backgroundTaskError` is surfaced to the app for optional display

**Tests:**
- `MockPostGenerationTask` in `BaseChatTestSupport`
- `PostGenerationTaskTests` (Core): execution order, cancellation on reset, error isolation

### Notes
Tasks run off `@MainActor` but are not prevented from performing their own `await` calls, including secondary inference. Apps that want parallel execution can wrap multiple tasks in a single `PostGenerationTask` implementation using a `TaskGroup`.

### Depends on
None.

PR #829, opened 2026-04-27
feat(core): BackgroundTaskScheduler protocol + iOS BGTaskScheduler impl

### LINKED ISSUE #114: feat: BackgroundTaskScheduler — BGProcessingTask integration for post-generation work
### Summary
Add a `BackgroundTaskScheduler` protocol and default implementation to `BaseChatCore` so apps can queue extraction, indexing, and archiving work through iOS's background task system, with memory budget enforcement.

### Motivation
Apps running extraction pipelines or vector indexing after generation need to integrate with `BGProcessingTask`/`BGAppRefreshTask` for iOS background execution. Without a framework seam, each app wires this up independently with no shared memory budget enforcement or cancellation contract.

### What ships

**BaseChatCore:**
- `BackgroundTaskScheduler` protocol:
  ```swift
  public protocol BackgroundTaskScheduler: Sendable {
      func schedule(_ work: @Sendable @escaping () async throws -> Void, identifier: String) async
      func cancel(identifier: String)
  }
  ```
- `DefaultBackgroundTaskScheduler`: uses `BGTaskScheduler` on iOS, runs inline on macOS
- Memory budget: tasks receive a `MemoryBudget` (default 50MB) — task is cancelled and rescheduled if memory pressure exceeds budget
- `ChatViewModel` uses `BackgroundTaskScheduler` to dispatch `PostGenerationTask` items (#111) when the app backgrounds mid-generation

**Tests:**
- `MockBackgroundTaskScheduler` in `BaseChatTestSupport`
- `BackgroundTaskSchedulerTests` (Core): schedule, cancel, memory pressure cancellation

### Notes
`BGTaskScheduler` requires task identifiers registered in Info.plist. Document this as a consumer responsibility. `BaseChatCore` defines a `BaseChatBackgroundTaskIdentifiers` namespace with recommended identifier strings.

### Depends on
- #111 (post-generation background hook — BackgroundTaskScheduler dispatches PostGenerationTasks)

roryford:
Add a framework seam so apps running extraction pipelines or vector indexing after generation can dispatch through iOS's `BGProcessingTask` system with a shared memory-budget contract, instead of every consumer wiring `BGTaskScheduler` and watchdog logic independently.

**What ships:**

- `BackgroundTaskScheduler` protocol (`BaseChatCore`) — takes `identifier`, `MemoryBudget`, and an async work closure. Schedule replaces any prior run under the same identifier; `cancel(identifier:)` is the explicit teardown.
- `DefaultBackgroundTaskScheduler` — submits a `BGProcessingTaskRequest` on iOS and runs inline on macOS. A `withTaskGroup` watchdog samples the process footprint via `phys_footprint` on `budget.sampleInterval` and cancels the worker when the ceiling is breached. The closure observes cancellation through `Task.isCancelled`.
- `MockBackgroundTaskScheduler` (`BaseChatTestSupport`) — schedules work, records calls, exposes `simulateMemoryBudgetExceeded(identifier:)` so tests can drive the cancellation contract deterministically. Only records cancellations for tasks that were actually in flight, so the implicit pre-clean cancel inside `schedule` doesn't pollute the test-visible record.
- `BackgroundTaskSchedulerTests` covering schedule/cancel, replace-on-same-identifier, simulated budget breach, and the recommended-identifiers stable-string contract.

All async-context lock paths use `NSLock.withLock { … }` rather than bare `lock()/unlock()` for Swift 6 strict concurrency compliance.

`ChatViewModel` wire-up for `PostGenerationTask` dispatch is deferred to the PR that introduces #111 — the scheduler ships standalone here.

Closes #114

### DIFF
--- a/Sources/BaseChatCore/Protocols/BackgroundTaskScheduler.swift
+++ b/Sources/BaseChatCore/Protocols/BackgroundTaskScheduler.swift
@@ -61,20 +61,27 @@ public enum BaseChatBackgroundTaskIdentifiers {
 
 // MARK: - Scheduler Protocol
 
-/// Schedules background work that survives the app being suspended on iOS.
+/// Schedules in-process background work with a memory-budget watchdog.
 ///
-/// The default implementation is ``DefaultBackgroundTaskScheduler``. On iOS
-/// it bridges to `BGTaskScheduler`; on macOS, where there is no equivalent
-/// foreground/background life-cycle, it runs the closure inline on a detached
-/// task. Both paths enforce the configured ``MemoryBudget`` — work whose
-/// process footprint breaches the ceiling has its `Task` cancelled, and the
-/// closure observes the cancellation through the standard `Task.isCancelled`
-/// / `try Task.checkCancellation()` channel. The host can re-schedule from
+/// The default implementation is ``DefaultBackgroundTaskScheduler``. It runs
+/// the closure on a detached `Task` on both iOS and macOS, and on iOS also
+/// submits a best-effort `BGProcessingTaskRequest` so a host app that wires
+/// up its own `BGTaskScheduler.register(...)` launch handler can pick up
+/// queued work after the process is relaunched. The default implementation
+/// itself does **not** install a launch handler and does **not** persist
+/// `work` across launches — treat the protocol as an in-process scheduler
+/// with budget enforcement, not as a guaranteed background-execution bridge.
+///
+/// Both paths enforce the configured ``MemoryBudget`` — work whose process
+/// footprint breaches the ceiling has its `Task` cancelled, and the closure
+/// observes the cancellation through the standard `Task.isCancelled` /
+/// `try Task.checkCancellation()` channel. The host can re-schedule from
 /// there.
 ///
 /// `identifier` is a free-form string. On iOS it must match an entry in
-/// `BGTaskSchedulerPermittedIdentifiers` (see ``BaseChatBackgroundTaskIdentifiers``);
-/// on macOS it is used purely as a cancellation key.
+/// `BGTaskSchedulerPermittedIdentifiers` (see ``BaseChatBackgroundTaskIdentifiers``)
+/// for the best-effort BG request submission to succeed; on macOS it is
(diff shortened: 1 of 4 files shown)


roryford:
Addressed CI failure + Copilot review in ed4911f.

**CI:** `SilentCatchAuditTest` was the lone failure — it caught two unobserved `} catch { }` blocks in the new scheduler/mock. Both now log non-`CancellationError` throws via `Log.persistence.error` and silently swallow the expected `CancellationError` only. Re-ran the full pre-push matrix locally (Core + Inference + InferenceSwiftTesting + UI + UIModelManagement + MCP + Backends + TestSupport + AppIntents) — all green.

**Copilot threads:**
- *atomic replace-if-newer in `schedule`*: fixed in both `DefaultBackgroundTaskScheduler` and `MockBackgroundTaskScheduler` — cancel+swap now happens as a single critical section under the lock, cancellation of the prior task runs outside the lock.
- *`_tasks` retaining finished tasks + ghost cancellations in mock*: fixed by tagging each scheduled run with a monotonic generation; the worker self-clears the slot on settle only when the generation still matches, so a newer schedule isn't evicted and finished tasks don't pile up. Cancelling an unknown identifier no longer records a ghost cancellation.
- *misleading "survives suspension" docs on protocol + default impl*: weakened the contract on both `BackgroundTaskScheduler` and `DefaultBackgroundTaskScheduler` — clear that this is an in-process scheduler with a memory-budget watchdog and a best-effort `BGProcessingTaskRequest` submission. Document that a real cross-process bridge requires the host app to register a launch handler under the same identifier (out of scope for this PR; tracked under #114's umbrella).
- *unused `import BaseChatInference` in `BackgroundTaskSchedulerTests`*: removed.
- *empty worker `catch` in scheduler*: replaced with `catch is CancellationError {} catch { Log.persistence.error(...) }` so non-cancellation throws are surfaced to Console/Instruments.

No unresolved Copilot threads remain that I think are correct; the BG launch-handler registration is a separate, larger piece of work and the docs now reflect that honestly rather than promising semantics the type doesn't deliver.

Issue #499, opened 2026-04-18
fuzz: separate raw-stream capture from rendered-as-user-sees-it


roryford:
## Why

`FuzzRunner.runSingle` sets `rendered: capture.raw` (`Sources/BaseChatFuzz/FuzzRunner.swift:206`). They're the same buffer. That collapse means:

1. Two `ThinkingClassificationDetector` sub-checks become redundant — `visible-text-leak` and `misclassified-as-text` overlap heavily because `rendered === raw`.
2. The harness can't catch UI-layer rendering bugs (markdown breakage, half-rendered tables, partial code fences) because it never assembles `MessagePart.text` parts the way the UI does.

Conceptually, `raw` should be "the bytes the backend yielded" and `rendered` should be "what `MessagePartsView` would actually display."

QA + architect review on PR #493 (item 18).

## What

Two paths:

1. **Cheap**: drop `rendered` from `RunRecord` since it duplicates `raw`. Simplify affected detectors to use `raw` only. Document that the harness intentionally inspects the *backend stream*, not the *UI-rendered output*.
2. **Right**: assemble `rendered` by running captured `MessagePart.text` parts through whatever transformations the UI applies (Markdown rendering, code-fence-aware processing). The only path that lets the harness catch UI-layer bugs.

Recommend (1) now, then (2) as a separate follow-up once we want to fuzz the rendering pipeline.

## Acceptance

- Either: `RunRecord.rendered` is removed and detectors use `raw`, or `rendered` is meaningfully different from `raw`.
- Detector tests pass.

PR #830, opened 2026-04-27
refactor(fuzz): assemble RunRecord.rendered through real markdown transforms

### LINKED ISSUE #543: fuzz: assemble RunRecord.rendered through real UI transforms (follow-up to #499)
## Context

Follow-up to #499. The short-term fix (see commit on #499 PR) deprecated `RunRecord.rendered` because it was always a byte-for-byte copy of `raw` — which meant two `ThinkingClassificationDetector` sub-checks (`visible-text-leak`, `misclassified-as-text`) overlapped heavily, and the harness could never catch UI-layer rendering bugs.

## The right path

Populate `RunRecord.rendered` by actually running `MessagePart.text` through the real `BaseChatUI` rendering pipeline:

- Markdown transformation (the same one `ChatMessageView` uses)
- Code-fence handling (including the partial-fence / streaming edge cases)
- Any inline tag stripping the UI applies

Once we do that, `rendered` diverges from `raw` in realistic, semantically meaningful ways, and detectors that read `rendered` catch genuinely different bugs than detectors that read `raw` — e.g.:

- markdown breakage under mid-stream cancellation
- half-rendered tables
- partial code fences being mis-terminated
- emoji / grapheme-cluster edge cases in the UI string

## Scope

1. Expose the rendering transform as a headless function the fuzzer can call (no SwiftUI view construction). Either a small helper inside `BaseChatUI` or lift the pure parts into `BaseChatCore`.
2. Wire `FuzzRunner.runSingle` to populate `rendered` from that function applied to `capture.raw` (or to the assembled `MessagePart` stream).
3. Re-audit `ThinkingClassificationDetector` and `LoopingDetector` sub-checks — once `rendered != raw`, revisit whether the `rendered-loop` sub-check should come back as a distinct detector on the UI-transformed string.
4. Un-deprecate `RunRecord.rendered` (or land the removal if we decide the new semantics warrant a fresh field name).

## Why deferred

The cheap path in #499 was enough to stop the harness from claiming it catches rendering bugs it can't catch. Threading `BaseChatUI`'s transform pipeline into the fuzzer target touches target boundaries (BaseChatFuzz currently doesn't depend on BaseChatUI) and deserves its own design pass.

## References

- Closes the "right path" described in the #499 issue body
- Parent PR: #499 (cheap path — deprecate field, route detectors through `raw`)

roryford:
Follow-up to #499. The short-term fix on that PR deprecated `RunRecord.rendered` because it was always a byte-for-byte copy of `raw` — which meant `ThinkingClassificationDetector` sub-checks (`visible-text-leak`, `misclassified-as-text`) overlapped heavily and the harness could never catch UI-layer rendering bugs.

This PR populates `RunRecord.rendered` by piping `MessagePart.text` through the real markdown block parser before the fuzzer records it.

**Refactor:**
- New `BaseChatInference/Utilities/MarkdownRendering.swift` — pure-Foundation block parser with no SwiftUI / Combine imports. Recognises 3+ backtick fences, tracks nested openings, salvages mid-stream by returning the entire input as a single markdown block.
- `BaseChatUI`'s `AssistantMarkdownView` now consumes `MarkdownRendering.parseBlocks(from:)` for live rendering, building `AttributedString` per block on top.
- The fuzz harness (`FuzzRunner`, `SessionScriptRunner`, `Replayer`) populates `RunRecord.rendered` by joining `MarkdownRendering` blocks back into a string, so detectors reading `rendered` exercise the same transforms users see.

**Detector changes:**
- `ThinkingClassificationDetector`: `visible-text-leak` and `misclassified-as-text` sub-checks now read different surfaces meaningfully — no more overlap.
- `LoopingDetector`: code-fence content no longer false-positives, since `rendered` strips fence markers before regex sees the input.
- `EmptyVisibleAfterThinkDetector`: stricter check now that `rendered` ≠ `raw` for code-heavy outputs.

Closes #543

### DIFF
--- a/Sources/BaseChatFuzz/BaseChatFuzz.docc/RunRecordSchema.md
+++ b/Sources/BaseChatFuzz/BaseChatFuzz.docc/RunRecordSchema.md
@@ -32,7 +32,7 @@ in #490) call it before acting on a decoded record.
 | `prompt` | `PromptSnapshot` | Corpus id, applied mutator list, message turns (role + text). |
 | `events` | `[EventSnapshot]` | Ordered stream of `{t, kind, v?}` observations. `t` is seconds since run start. |
 | `raw` | `String` | Raw backend output with think/tool markers intact. |
-| `rendered` | `String` | User-visible rendering of `raw` (stripped of reasoning). May be deprecated in a future version — see #499. |
+| `rendered` | `String` | User-visible string produced by running `raw` through the same markdown / code-fence transform that `BaseChatUI`'s `AssistantMarkdownView` applies (see `MarkdownRendering.renderToVisibleString`). Diverges from `raw` in realistic ways: closing fences disappear, emphasis markers collapse, link targets are hidden. Records written before #543 had `rendered == raw`. |
 | `thinkingRaw` | `String` | Concatenation of all thinking-channel text. |
 | `thinkingParts` | `[String]` | Thinking deltas as a list, in stream order. |
 | `thinkingCompleteCount` | `Int` | Number of `thinkingComplete` events observed. |
--- a/Sources/BaseChatFuzz/Detectors/EmptyVisibleAfterThinkDetector.swift
+++ b/Sources/BaseChatFuzz/Detectors/EmptyVisibleAfterThinkDetector.swift
@@ -6,11 +6,10 @@ import Foundation
 /// `EmptyOutputAfterWorkDetector`: the model *did* produce raw tokens and
 /// *did* produce thinking, but the visible rendering is empty.
 ///
-/// Today's `rendered` field holds only the user-visible portion of the stream
-/// (no `<think>` blocks). It is doc-comment-deprecated in favour of a future
-/// "real UI-transform rendering" path; until that path ships, `rendered` is
-/// the only field that tracks the visible-only output, so this detector
-/// continues to read it rather than `raw` (which includes thinking tokens).
+/// `rendered` holds the user-visible portion of the stream after the real
+/// UI transform pipeline (#543). It is the right field to read for this
+/// detector — `raw` would include `<think>` markers and inline syntax that
+/// the user never actually sees as glyphs.
 ///
 /// Ships at `.flaky` severity. Promotion to `.confirmed` requires the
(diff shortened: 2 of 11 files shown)


roryford:
Reviewed and pushed one follow-up commit (be3e38a).

**Verdict: ship.** The PR delivers what #543 asked for — `RunRecord.rendered` is now produced by a real markdown transform, not a copy of `raw`.

**Correctness.** `AssistantMarkdownView` and the fuzz harness now share `MarkdownRendering.parseBlocks` directly (UI's `AssistantMarkdownParser` is a thin SwiftUI-id wrapper around it), so fence handling can't drift between live and headless paths. `renderToVisibleString` is documented honestly as a glyph-level approximation rather than a clone of `AttributedString(markdown:)` rendering — that's the right call given the UI ultimately produces `AttributedString`, not `String`.

**Performance.** `renderToVisibleString` runs once per `RunRecord` finalization (`FuzzRunner`, `Replayer`, `SessionScriptRunner`), same locus as the old `rendered: capture.raw` assignment — no per-token regression.

**Detectors.** `LoopingDetector` correctly adds a distinct `raw-loop` sub-check only when `rendered != raw`; existing `rendered-loop` IDs stay stable for sink dedup. `ThinkingClassificationDetector` sub-checks 1 and 2 are now genuinely distinct.

**Backwards compat.** `RunRecord.rendered` is un-deprecated; `init(from:)` decodes legacy records (where `rendered == raw`) unchanged.

**Test gap I closed (be3e38a).** No coverage existed for `MarkdownRendering` itself — drift from the UI parser would have landed silently. Added 16 `MarkdownRenderingTests` exercising:

- `parseBlocks` (empty, prose-only, fenced, partial-fence salvage, nested 4-tick fence)
- `renderToVisibleString` (emphasis, links, inline code, headings, lists, fence stripping, partial-fence flattening, divergence-from-raw, empty)
- `renderVisibleText` (only `.text` parts contribute)

Sabotage-verified: collapsing `renderToVisibleString` to `return source` fails 8 tests; restored, all pass.

Full CI matrix locally: 2388 tests, 0 failures.

PR #831, opened 2026-04-27
test(fuzz): pin LlamaFuzzFactory contract + accurate FUZZING.md backends table

### LINKED ISSUE #560: fuzz: implement LlamaFuzzFactory and wire --backend llama

roryford:
The Llama fuzz factory was already wired in `Sources/fuzz-chat/` but lacked test coverage for the invariants the runner relies on, and `FUZZING.md` still claimed `--backend llama` was unwired and would throw.

**Tests pinned:**

- single-model-per-campaign (because `llama_backend_init` is a process-global one-shot, rotation never applies)
- `supportsDeterministicReplay = true` (seed + temperature=0 → bit-identical output for replay)
- GGUF model discovery via `HardwareRequirements.findGGUFModel` against `~/Documents/Models/`
- end-to-end smoke: at least one iteration through `FuzzRunner` on real hardware

Test mirrors the executable factory locally because SPM test targets cannot import an `executableTarget`. Skipped when the host lacks Apple Silicon, Metal, or a GGUF model installed.

**Docs:**

- `FUZZING.md` backends table now reflects that `llama` / `mock` / `chaos` are all wired; only `all` remains unimplemented.
- "Llama opt-out" section corrected: the CLI path is wired and stays single-model even under `--model all`.

Closes #560

### DIFF
--- a/FUZZING.md
+++ b/FUZZING.md
@@ -43,7 +43,7 @@ Common flags:
 
 | Flag | Purpose |
 |------|---------|
-| `--backend <name>` | `ollama` (default), `llama`, `foundation`, `mlx`, `all`. Only `ollama` is wired in v1. |
+| `--backend <name>` | `ollama` (default), `llama`, `foundation`, `mlx`, `mock`, `chaos`, `all`. `mlx` runs via the xcodebuild path (see below); `all` is not yet implemented. |
 | `--minutes N` | Wall-clock budget. Default 5. |
 | `--iterations N` | Iteration cap; runs until either budget is hit. |
 | `--single` | One iteration then exit — useful with `--seed`. |
@@ -87,7 +87,7 @@ Bug shapes diverge per model. The #487 `thinking`-drop only showed up on reasoni
 
 **Determinism.** The discovered model list is sorted by UTF-8 byte order before rotation, so two invocations on the same machine — regardless of the order Ollama reports its models — produce the same iteration-to-model mapping. This is the contract that keeps `--seed N --replay` (#490) meaningful: given a fixed seed and a fixed installed-model list, the rotation sequence is reproducible.
 
-**Llama opt-out.** `LlamaBackend` calls `llama_backend_init` as a process-global one-shot — only one instance per process is supported. Rotation never applies to Llama. The CLI's Llama path is still unwired today and throws; when it lands, it will stay single-model even under `--model all`.
+**Llama opt-out.** `LlamaBackend` calls `llama_backend_init` as a process-global one-shot — only one instance per process is supported. Rotation never applies to Llama. The CLI's Llama path is wired (`--backend llama` discovers the first GGUF in `~/Documents/Models/` via `HardwareRequirements.findGGUFModel`) and stays single-model even under `--model all`.
 
 **Mechanism.** The rotating factory is a `FuzzBackendFactory` conformance that wraps an ordered array of child factories and advances an internal index per `makeHandle()` call. The runner's `init(config:factory:)` contract (#537) is unchanged — rotation is hidden behind the factory boundary. See `RotatingFuzzFactory` in `Sources/BaseChatFuzz/`.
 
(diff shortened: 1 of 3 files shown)


roryford:
Reviewed and pushed three follow-ups for sibling-PR consistency with #825 (FoundationFuzzFactory):

1. **Explicit determinism in the real factory.** `LlamaFuzzFactory` now declares `public var supportsDeterministicReplay: Bool { true }` instead of inheriting the protocol-extension default. Mirrors #825 — readers comparing factories side-by-side see the same shape, and the contract test now pins a stated value instead of the default.

2. **Local mirror states it explicitly too.** `LocalLlamaFuzzFactory` declares the same value so the test is pinning factory-stated behaviour. Sabotage-verified: flipping the mirror to `false` fails `test_supportsDeterministicReplay_isTrue`.

3. **Pure-value test runs on every host.** Moved hardware-skip guards out of `setUp()` into a per-test `skipUnlessHardwareReady()` helper. `test_supportsDeterministicReplay_isTrue` is pure-value — no Apple Silicon / Metal / GGUF needed — so it now runs everywhere the test target builds, matching the pure-value-vs-hardware-gated split in `FoundationFuzzFactoryTests`.

Also added a `templateMarkers == nil` assertion in `test_makeHandle_returnsLoadedBackend` to match the equivalent assertion in #825's happy-path test.

**Verified locally** (Apple Silicon + GGUF in `~/Documents/Models/`):
- `swift test --filter LlamaFuzzFactoryTests --traits Fuzz,Llama,Ollama` — all 3 tests pass
- `swift test --filter BaseChatFuzzTests --disable-default-traits` — all 157 tests pass (file body fully gated out under no-traits CI)
- `swift build --build-tests --disable-default-traits` — clean build

**FUZZING.md table accuracy** confirmed against `Sources/fuzz-chat/FuzzChatCLI.swift`:
- `ollama` / `llama` / `foundation` / `mock` / `chaos` — wired
- `mlx` — fails fast in `swift run` path with explicit `scripts/fuzz.sh --with-mlx` redirect, so "runs via the xcodebuild path (see below)" is accurate
- `all` — `case .mlx, .all: fail("\(backend.rawValue) backend not yet wired in CLI.")` — "not yet implemented" is accurate

**Out-of-scope finding (not addressed):** `makeOllamaFactory` is gated `#if Fuzz` but references `OllamaFuzzFactory` which is gated `#if Fuzz && Ollama`. Building with `--traits Fuzz,Llama` (no Ollama) fails with `cannot find 'OllamaFuzzFactory' in scope`. Pre-existing on `main`, not a regression from this PR.

Issue #609, opened 2026-04-21
track(grammar): GBNF schema pre-validator + CVE-2026-2069 audit


roryford:
## Context

`LlamaGenerationDriver.swift:94-116` drives llama.cpp sampling today with no grammar sampler attached — every token is sampled freely from the full vocabulary. That is fine for chat but weak for tool calling: without grammar enforcement, models frequently emit almost-valid JSON (trailing commas, missing closing braces, hallucinated field names) that fails tool-call schema validation downstream.

llama.cpp ships a GBNF (GGML BNF) grammar sampler precisely for this case. Reported results on Qwen3-14B show F1 = 0.971 for tool-call schema adherence with GBNF enforcement vs notably lower free-sampling numbers. For structured tool calling this is the right tool.

## Blockers — do not start until #55 settles

GBNF is not free to adopt:

- **Schema-shape fragility.** llama.cpp's GBNF compiler **SIGSEGVs on complex JSON Schemas** — in particular `anyOf`, nullable unions (`type: ["string", "null"]`), and numeric `exclusiveMinimum`/`exclusiveMaximum`. Any implementation must pre-validate `ToolDefinition.parameters` shape before handing the schema to the compiler. Hitting the segfault kills the process, which under `swift test` means the whole suite dies with the test that tripped it.
- **CVE-2026-2069.** Buffer overflow in `llama_grammar_advance_stack()`. Any ship must pin to a post-CVE llama.cpp version and re-audit whenever the `llama.cpp` vendor bump lands. If the vendor version regresses, GBNF must be disabled at the feature-flag level, not by praying.

## Dependency

This issue is **a dependency of #55 (tool calling for LlamaBackend)**, not a parallel workstream. The right direction for #55 — native tools vs forwarding a flat prompt — decides whether GBNF is on-path at all. Do not start this work until #55's direction is settled.

## Acceptance (when it is time)

- [ ] Schema-shape validator rejecting `anyOf`, nullable unions, `exclusiveMinimum`/`exclusiveMaximum` before GBNF compile
- [ ] `llama.cpp` version pinned to a post-CVE-2026-2069 release
- [ ] Feature flag that keeps GBNF off by default until the validator is mature
- [ ] Integration test for a known-good Qwen3 tool-call schema (green) and a known-bad schema with `anyOf` (fails at the validator, not at the compiler)

Tracked so the next author reading `LlamaGenerationDriver.generate(…)` doesn't independently rediscover the tradeoff.

[...]

PR #836, opened 2026-04-27
feat(llama): Gemma 4 tool calling + GBNF schema pre-validator (#414, #609)

### LINKED ISSUE #414: feat: tool calling support for LlamaBackend (Gemma 4)
## Problem

`BackendCapabilities.supportsToolCalling` exists but no local backend implements it. `MessagePart` previously had `toolCall`/`toolResult` cases that were removed (see the decode fallback comment in `MessagePart.swift`). Gemma 4 GGUF via llama.cpp is a natural first target since the model emits structured tool tokens directly in the stream.

**Gemma 4 tool token format:**
```
<|tool>                         ← tool declaration block (in system turn)
{name, description, parameters}
<tool|>

<|tool_call>                    ← model requests a tool call
call:function_name{param1:<|"|>value<|"|>,param2:<|"|>value<|"|>}
<tool_call|>

<|tool_response>                ← host injects the tool result
{result}
<tool_response|>
```

### `MessagePart`
Restore (or add new) discriminators:
- `.toolCall(name: String, arguments: [String: String])`
- `.toolResult(name: String, content: String)`

### `ToolDefinition` model
A new `InferenceService`-layer type describing a callable tool (name, description, parameter schema). Passed in via `GenerationConfig` or a separate `ToolRegistry`.

### `PromptTemplate.gemma4` formatting
Serialize `ToolDefinition` entries into the system turn as `<|tool>...<tool|>` blocks.

### Stream parser
Detect `<|tool_call>...<tool_call|>` in the output stream, parse name + arguments, and surface a `.toolCall` event rather than raw text. Pause generation and wait for the host to inject a `<|tool_response>` continuation.

### `BackendCapabilities`
Set `supportsToolCalling: true` in `LlamaBackend.capabilities` once implemented.

## Dependencies

Requires #412 (Gemma 4 prompt template).

roryford:
## Summary

Implements Gemma 4 GGUF tool calling via llama.cpp (issue #414, umbrella #753 group E) and adds the GBNF schema pre-validator + CVE-2026-2069 audit required by issue #609.

### LlamaBackend tool calling

- `LlamaBackend.capabilities.supportsToolCalling`: `false` → `true`
- **New:** `LlamaToolCallParser` — chunk-safe stream parser for Gemma 4 `<|tool_call>…<|end_of_turn>` tokens. Falls back to the JSON `<tool_call>…</tool_call>` format (Qwen-style fine-tunes). Follows the same prefix-hold pattern as `MLXToolCallParser` and `ThinkingParser`.
- Wire `LlamaToolCallParser` into `LlamaGenerationDriver.run()` in the token-event pipeline, after `ThinkingParser`. Active only when `config.tools` is non-empty — zero overhead otherwise.

### Gemma 4 prompt template

- Add `<|tool>`, `<|tool_call>`, `<|tool_response>` to Gemma 4 `specialTokens` (sanitised from user input).
- Add `format(messages:systemPrompt:tools:)` overload. For `.gemma4`, injects `<|tool>{JSON}` declaration blocks into the system turn — one per `ToolDefinition`. All other templates ignore `tools` (identical output to existing overload).
- Update `GenerationCoordinator` to pass `config.tools` when backend `supportsToolCalling` and tools are set.

### GBNF schema pre-validator (#609 / CVE-2026-2069)

- **New:** `GBNFSchemaPreValidator` (in `BaseChatInference`, no ML deps). Validates a `JSONSchemaValue` before `llama_sampler_init_grammar`. Rejects:
  - `anyOf` / `oneOf` / `allOf` / `not` (GBNF IR cannot express combiners)
  - Nullable union `type: ["string", "null"]` (stack overflow in GBNF compiler)
  - `exclusiveMinimum` / `exclusiveMaximum` (type-confusion in GBNF numeric rules)
  - Recursively checks `properties`, `items`, `additionalProperties`
- Carries a pinned `CVEAuditRecord` for CVE-2026-2069 (buffer overflow in `llama_grammar_advance_stack()`). Vendored build b8772 pre-dates the fix (b8774). Includes upgrade procedure.
- `docs/LLAMA_CONTRACT.md`: add CVE-2026-2069 security section.

## Tests

| Suite | New tests |
|---|---|
| `GBNFSchemaPreValidatorTests` | Known-good schemas pass; all 7 rejection rules fail with correct `path` arrays; CVE audit record sanity |
| `LlamaToolCallParserTests` | Gemma 4 native format, JSON fallback, chunk safety (tag split / close-tag split / byte-by-byte), malformed input discarded, `finalize()` semantics, distinct IDs |
| `LlamaToolCapabilityTests` | `supportsToolCalling: true`, tool blocks in Gemma 4 system turn, system+tools combined, multi-tool ordering, special-token sanitisation in user content, non-Gemma4 templates unaffected |

All CI suites pass locally (`--disable-default-traits`):
`BaseChatCoreTests`, `BaseChatInferenceTests`, `BaseChatInferenceSwiftTestingTests`, `BaseChatUITests`, `BaseChatUIModelManagementTests`, `BaseChatMCPTests`, `BaseChatBackendsTests`, `BaseChatTestSupportTests`, `BaseChatAppIntentsTests`.

## Release Note

Gemma 4 GGUF models loaded via llama.cpp now support tool calling. Pass `ToolDefinition` values in `GenerationConfig.tools` and select the `.gemma4` prompt template — the framework injects `<|tool>` declaration blocks into the system turn and parses `<|tool_call>` tokens from the model stream into `GenerationEvent.toolCall` events.

A new `GBNFSchemaPreValidator` guards GBNF grammar compilation against CVE-2026-2069 (buffer overflow in `llama_grammar_advance_stack()`, vendored build b8772 is pre-fix). The validator rejects schema combiners and nullable union types before they reach `llama_sampler_init_grammar`. See `LLAMA_CONTRACT.md` § Security for the upgrade procedure once the xcframework pin is bumped past b8773.

Closes #414 (umbrella #753 group E), #609.

### DIFF
--- a/Sources/BaseChatBackends/LlamaBackend.swift
+++ b/Sources/BaseChatBackends/LlamaBackend.swift
@@ -50,7 +50,7 @@ public final class LlamaBackend: InferenceBackend, @unchecked Sendable {
             maxContextTokens: ctxSize,
             requiresPromptTemplate: true,
             supportsSystemPrompt: true,
-            supportsToolCalling: false,
+            supportsToolCalling: true,
             supportsStructuredOutput: false,
             supportsNativeJSONMode: false,
             cancellationStyle: .explicit,
--- a/Sources/BaseChatBackends/LlamaGenerationDriver.swift
+++ b/Sources/BaseChatBackends/LlamaGenerationDriver.swift
@@ -288,6 +288,14 @@ struct LlamaGenerationDriver {
         // Flag set when maxThinkingTokens is reached so we can break the outer loop cleanly.
         var thinkingLimitReached = false
 
+        // Tool-call parser: active when the config carries at least one tool definition.
+        // Processes `.token` events from the thinking layer (or raw decoded text) and
+        // re-routes `<|tool_call>…<|end_of_turn>` / `<tool_call>…</tool_call>` blocks
+        // into `.toolCall` events. When no tools are configured the parser is skipped
+        // entirely — zero overhead on non-tool-calling generation paths.
+        let useToolParser = !config.tools.isEmpty
+        var toolCallParser = LlamaToolCallParser()
+
         // Repetition-window state: track the last decoded token string and how
         // many times it has appeared consecutively. Exceeding `maxRepeatWindow`
         // triggers an early exit — no need to run the loop all the way to maxTokens.
@@ -348,9 +356,27 @@ struct LlamaGenerationDriver {
 
                 // Build the list of events this token emits. Two paths, matching the
                 // two thinking-marker modes documented at the top of the loop:
-                let events: [GenerationEvent] = useParser
+                let thinkingEvents: [GenerationEvent] = useParser
                     ? thinkingParser.process(text)
                     : [.token(text)]
+
+                // Route `.token` events through the tool-call parser when active.
+                // Non-token events (.thinkingToken, .thinkingComplete) pass straight
+                // through — tool calls never appear inside thinking blocks.
+                let events: [GenerationEvent]
+                if useToolParser {
+                    var routed: [GenerationEvent] = []
+                    for ev in thinkingEvents {
+                        if case .token(let t) = ev {
+                            routed += toolCallParser.process(t)
+                        } else {
+                            routed.append(ev)
+                        }
+                    }
+                    events = routed
+                } else {
+                    events = thinkingEvents
+                }
 
                 for event in events {
                     if isFirstToken {
@@ -399,6 +425,11 @@ struct LlamaGenerationDriver {
         // false avoids yielding spurious events from an untouched parser.
         if useParser {
             for event in thinkingParser.finalize() {
+                continuation.yield(event)
+            }
+        }
+        if useToolParser {
+            for event in toolCallParser.finalize() {
                 continuation.yield(event)
             }
         }
--- a/Sources/BaseChatBackends/LlamaToolCallParser.swift
+++ b/Sources/BaseChatBackends/LlamaToolCallParser.swift
@@ -1,0 +1,248 @@
+#if Llama
+import Foundation
+import BaseChatInference
+
+/// Stateful, chunk-safe parser that extracts tool calls from Gemma 4 GGUF
+/// token streams produced by `LlamaGenerationDriver`.
+///
+/// ## Token format
+///
+/// Gemma 4 IT GGUF models emit tool invocations using special tokens:
+///
+/// ```
+/// <|tool_call>
+/// call:function_name{param1:<|"|>value<|"|>}
+/// <|end_of_turn>
+/// ```
+///
+/// `<|"|>` is Gemma 4's string-quoting special token; it is substituted with
+/// `"` before the call body is interpreted as a JSON-like object.
+///
+/// ## Fallback format
+///
+/// For tool-calling fine-tunes that deviate from the Gemma 4 native format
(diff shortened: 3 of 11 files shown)


roryford:
Rebased onto `main` (clean — no conflicts) and addressed all four Copilot inline comments. New head: `55ab2e0`.

### Fixes pushed

1. **`LlamaToolCallParser.parseGemma4NativeCall` — invalid-JSON bug.** Copilot was right: the brace body uses unquoted JSON-like keys (`{key:"value"}`), so handing it to `JSONSerialization` always failed and every native tool call silently fell back to `arguments == "{}"`. Replaced with a dedicated key:value tokenizer that preserves typed scalars (numbers, booleans, null) and `<|"|>`-quoted strings, then round-trips through `JSONSerialization` for canonicalisation. Added `test_gemma4NativeCall_mixedScalarArgs_areTypedNotStrings` so the regression cannot reappear silently.

2. **`GBNFSchemaPreValidator` — tuple-`items` bypass.** `items` may be either a single sub-schema object or an array of schemas (tuple validation, draft-04 / 2019-09). The earlier recursion only handled the object form, so an `anyOf` nested inside a tuple-form `items` array would bypass the CVE-2026-2069 mitigation. Both forms now validate, with index-qualified failure paths (`items.0`, `items.1`, …). Two new tests cover the new branch.

3. **`PromptTemplate.formatGemma4` — empty-system-turn footgun.** When every tool's serialisation returned nil and there was no system prompt, the template still emitted `<|turn>system\n<|end_of_turn>`. Now the parts are built first and the turn is emitted only when at least one survived. Tool-serialisation failures log at `.error` (per CLAUDE.md "no fatalError in recoverable paths").

4. **`LlamaToolCallParserTests` — broken `try XCTUnwrap` + nonsense finalize test.** Two `try XCTUnwrap` call sites needed `throws` markers. The file is `#if Llama`-gated so the build break only surfaced under `--traits Llama` (CI doesn't run that trait). Reshaped `test_finalize_emitsRemainingPlainText` into `test_finalize_emitsBufferedPartialTagPrefix` — the original drove text the parser already flushes in `process()`, so `finalize()` had nothing to emit and the test asserted the wrong thing.

The Copilot comment about `||` table rows in `docs/LLAMA_CONTRACT.md` was a false positive — the table uses single pipes correctly.

### Wire-shape consistency with #812 (Foundation)

Verified: both backends emit whole-call `ToolCall` events with provider-prefixed UUID-derived IDs (`fm-…` for Foundation, `llama-<name>-…` for Llama), `streamsToolCallArguments: false` (default), arguments serialised through `JSONSerialization` for canonicalisation. The orchestrator drives both the same way.

### Tests

Pre-push CI-safe chain:
- `BaseChatCoreTests`: 275 / 275
- `BaseChatInferenceTests`: 1237 / 1237 (incl. SilentCatchAuditTest after allowlist updated for the parser refactor)
- `BaseChatInferenceSwiftTestingTests`: 83 / 83
- `BaseChatUITests`: 4 / 4
- `BaseChatUIModelManagementTests`: 94 / 94
- `BaseChatMCPTests`: 104 / 104
- `BaseChatBackendsTests` (default traits): 109 / 109
- `BaseChatTestSupportTests`: 16 / 16
- `BaseChatAppIntentsTests`: 11 / 11

Llama-trait suite: `GBNFSchemaPreValidatorTests` + `LlamaToolCallParserTests` + `LlamaToolCapabilityTests` all pass (41 / 41) on Apple Silicon.

Not merging — leaving for review.

PR #839, opened 2026-04-27
refactor: introduce injectable sleeper and jitter provider for deterministic retry tests

### LINKED ISSUE #453: refactor: introduce Clock abstraction for deterministic SSE backoff and retry tests
## Problem

The SSE/cloud-backend retry policy uses exponential backoff with jitter. There's no deterministic test that verifies the backoff stays within its intended bounds (`[base, base * 2^n + jitterMax]`) because:

- `Task.sleep` in the retry loop reads the real wall clock.
- Jitter uses a non-seeded RNG.

A test that tried to assert bounds over 10 iterations would be flaky (real-clock timing) or meaningless (no way to control jitter).

## Why deferred from #429–#432 test-coverage plan

Documented in `.claude/plans/peppy-sprouting-fairy.md` (out-of-scope section). Introducing a `Clock` abstraction is a cross-cutting refactor — touches every async timing path in the backend layer. Doesn't fit a bug-fix-shape PR.

## Approach sketch

1. Introduce an internal `protocol Clock { func sleep(for: Duration) async throws }` (or adopt Swift 5.7+ `Clock` conformance).
2. Default to `ContinuousClock` in production (via `BaseChatConfiguration` or an init param).
3. In tests, inject a `TestClock` that advances deterministically on a test-controlled step.
4. Same pattern for jitter: inject a seeded `RandomNumberGenerator` defaulting to `SystemRandomNumberGenerator`.
5. Once both are injected, add bounds tests that advance the clock and assert the backoff delays sit inside `[base, base * 2^n + jitterMax]`.

## Scope

- `Sources/BaseChatBackends/RetryPolicy.swift` (or equivalent) — accept `Clock` + RNG at init
- `Sources/BaseChatBackends/SSECloudBackend.swift` — route through injected clock
- `Sources/BaseChatTestSupport/` — `TestClock` helper, matching the standard library's shape
- `Tests/BaseChatBackendsTests/RetryPolicyTests.swift` (or extend existing) — bounds test

## Relevant prior context

- `Tests/BaseChatBackendsTests/CloudBackendSSETests.swift` now contains hazard tests (PR #431) that deliberately sidestep timing-dependent assertions because of this gap.

roryford:
Closes #453

## Summary

Introduces two injection points that make SSE backoff/retry tests fully deterministic without real wall-clock delays or non-seeded RNG.

### `Sources/BaseChatInference/Services/RetryPolicy.swift`
- **`ExponentialBackoffStrategy`** gains a `jitterProvider: @Sendable (Double) -> Double` init parameter. The closure receives the jitter ceiling (25% of the exponential delay) and returns the actual jitter to apply. Default is `Double.random(in: 0...$0)` — identical to the previous hard-coded behaviour.
- **`withRetry(strategy:sleeper:operation:)`** gains a `sleeper: @Sendable (Duration) async throws -> Void` parameter, defaulting to `Task.sleep`. The production call sites are unchanged.

### `Sources/BaseChatBackends/SSECloudBackend.swift`
- Adds `public var retrySleeper: (@Sendable (Duration) async throws -> Void)?`. When `nil` (default), the backend falls back to `Task.sleep`. Tests set this to a `RecordingRetrySleeper` to skip real delays.

### `Sources/BaseChatTestSupport/RecordingRetrySleeper.swift`  *(new)*
- `RecordingRetrySleeper` captures every `Duration` passed to the sleeper using `OSAllocatedUnfairLock`, and exposes `recordedSleeps: [Duration]` for assertions. Does not actually sleep.

### `Tests/BaseChatBackendsTests/RetryPolicyTests.swift`  *(new)*
8 deterministic tests, all completing in <1ms:
- Exponential delay formula with zero jitter
- Max jitter ceiling locked to 25%
- Live random jitter stays within `[base, base * 1.25]`
- `Retry-After` header overrides exponential formula
- Missing `Retry-After` falls back to exponential
- `maxRetries` and `maxTotalDelay` exhaustion paths
- Non-retryable / non-`BackendError` short-circuit (no sleep)
- `withRetry` sleeper call sequence: 2 retries → 2 recorded sleeps at 1000ms + 2000ms
- Full exhaustion records all 3 delays (1000ms, 2000ms, 4000ms)

## Test Note

`## Release Note`
Retry delay arithmetic and jitter can now be tested deterministically. Inject `RecordingRetrySleeper` via `backend.retrySleeper` and a zero-jitter `ExponentialBackoffStrategy` to assert exact sleep durations without wall-clock timing.

### DIFF
--- a/Sources/BaseChatBackends/SSECloudBackend.swift
+++ b/Sources/BaseChatBackends/SSECloudBackend.swift
@@ -107,6 +107,11 @@ open class SSECloudBackend: InferenceBackend, ConversationHistoryReceiver, @unch
     /// custom strategy for tests.
     public var retryStrategy: any RetryStrategy = ExponentialBackoffStrategy()
 
+    /// Called to perform each retry delay. Defaults to `nil`, which uses `Task.sleep`
+    /// (real wall clock). Inject a ``RecordingRetrySleeper`` in tests to assert delay
+    /// bounds without real-time blocking.
+    public var retrySleeper: (@Sendable (Duration) async throws -> Void)?
+
     /// Idle timeout for the generation stream. If no SSE event arrives within
     /// this duration, the stream throws ``CloudBackendError/timeout(_:)``.
     /// `nil` disables idle detection (default).
@@ -330,6 +335,7 @@ open class SSECloudBackend: InferenceBackend, ConversationHistoryReceiver, @unch
         withStateLock { _activeEventIDTracker = eventIDTracker }
 
         let capturedStrategy = retryStrategy
+        let capturedSleeper = retrySleeper
         let session = self.urlSession
         let capturedTimeout = streamIdleTimeout
         let capturedBaseURL = baseURL
@@ -369,7 +375,10 @@ open class SSECloudBackend: InferenceBackend, ConversationHistoryReceiver, @unch
                     // Retry wraps only the HTTP connection phase — not SSE parsing.
                     // Mid-stream failures propagate immediately, preserving
                     // already-yielded tokens.
-                    let (bytes, _) = try await withRetry(strategy: capturedStrategy) {
+                    let (bytes, _) = try await withRetry(
+                        strategy: capturedStrategy,
+                        sleeper: capturedSleeper ?? { try await Task.sleep(for: $0) }
+                    ) {
                         let attempt = retryCounter.incrementAndGet()
                         if attempt > 1 {
                             await MainActor.run { streamBox.value?.setPhase(.retrying(attempt: attempt - 1, of: maxRetries)) }
--- a/Sources/BaseChatInference/Services/RetryPolicy.swift
+++ b/Sources/BaseChatInference/Services/RetryPolicy.swift
@@ -68,7 +68,11 @@ public struct ExponentialBackoffStrategy: RetryStrategy, Sendable {
         }
 
         let exponentialDelay = baseDelay * pow(2.0, Double(attempt))
-        let jitter = jitterProvider(exponentialDelay * 0.25)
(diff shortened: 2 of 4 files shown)


roryford:
Review pass complete — addressed both Copilot inline comments and added cancellation coverage.

**Pushed `8b1e26c`:**

- **Jitter clamp** (Copilot #1): `ExponentialBackoffStrategy.delay` now clamps `jitterProvider`'s return to `[0, maxJitter]` and rejects NaN. A misbehaving injected provider can no longer push delays below the exponential base or above `base * 1.25`.
- **Rounded ms** (Copilot #2): `withRetry` now uses `Int((delay * 1000).rounded())` instead of `Int(delay * 1000)`. Sub-millisecond `Retry-After` values (e.g. `0.0005s`) no longer collapse to a zero-duration sleep. Comment in the test was already correct ("millisecond-rounded").
- **New tests** in `BaseChatBackendsTests/RetryPolicyTests`:
  - `test_cancellingSleeperPropagatesCancellationError` — a sleeper that throws `CancellationError` surfaces unchanged through `withRetry`.
  - `test_jitterAboveCeilingIsClamped` — provider returning `10x` ceiling clamps to ceiling.
  - `test_negativeOrNaNJitterIsClamped` — negative and `.nan` jitter both collapse to 0.

**Verification**: 1219 tests in `BaseChatInferenceTests` and 124 in `BaseChatBackendsTests` (all 15 RetryPolicyTests included) pass locally on the rebased branch. No existing call sites needed changes — the new `sleeper:` parameter has a default value and the clamp is a no-op for the default `Double.random(in: 0...$0)` provider.

**Verdict**: ready for review. Not auto-merging.

PR #840, opened 2026-04-27
feat(example): Share + Action Extension recipes (App Group handoff)

### LINKED ISSUE #440: docs(example): Share + Action Extension recipes (App Group handoff)
## Motivation

Sharing text / URL / image / file into a BCK-powered app is a high-value pattern (summarize, draft reply, extract). BCK has no reference implementation, so every consumer reinvents App Group plumbing and rediscovers extension memory-cap constraints independently.

## Proposal

Add to the demo app:

- A Share Extension target with an App Group entitlement shared with the host app
- A `PendingSharePayload` codable persisted to the App Group container
- Host-app activation hook that drains pending payloads on foreground
- End-to-end example: share a URL from Safari, host app opens a new session with the URL attached

The extension does **no** inference — payload queuing only. Inference runs in the host app on next activation.

## Acceptance criteria

- Entitlement checklist documented
- Works on macOS and iOS
- Extension stays under the memory cap by design
- Paired with the `ChatViewModel.ingestPendingPayload` API (companion issue)

Depends on #442 landing first.

## Folded-in scope (from #441)

Add an Action Extension target to the same demo project, reusing the App Group container and `PendingSharePayload` plumbing introduced for the Share Extension. Recipe: "summarize selection from any app" — user invokes the Action Extension on selected text, payload is queued to the App Group, host app drains on next foreground and opens a session pre-seeded with the selection. One demo PR can ship both extension targets together since they share entitlements, payload type, and host-side ingestion path.

roryford:
## Summary

Closes #440.

Adds iOS Share Extension and Action Extension targets to BaseChatDemo, demonstrating the App Group handoff pattern for pre-filling a new `ChatViewModel` session without linking BaseChatKit inside the extension.

## Architecture

The key isolation boundary is `PendingSharePayload` — a pure Foundation `Codable` compiled into all three targets (host app + both extensions). Extensions write to App Group key `bck.pending-share`; the host app drains on foreground via `.onChange(of: scenePhase)` + a `.task(id:)` for the cold-launch race.

## What ships

- `Extensions/PendingSharePayload.swift` — pure Foundation shared type (no BCK deps)
- `BaseChatDemoShareExtension` — Share sheet extension accepting text, URL, image
- `BaseChatDemoActionExtension` — Action row extension accepting text, URL
- `BaseChatDemoApp` — `checkForPendingSharePayload()` + `pendingPayload(from:)` drain helpers wired to scenePhase
- `DemoAppGroup.pendingShareKey` constant (separate from `bck.inbound` App Intent key)
- Xcode project: two iOS-only `app-extension` targets, Sources/Frameworks/Embed build phases, entitlements, Info.plists, Debug+Release configs
- `docs/share-action-extension-recipe.md` — integration guide (entitlement checklist, isolation boundary, cold-launch race, adapting to your own app)
- `Example/README.md` — Extensions section added

### Share & Action Extension recipe for iOS handoff

Adds two iOS app extension targets to BaseChatDemo showing how to pipe content from the system Share sheet or action row into a new BaseChatKit chat session. Extensions are intentionally thin: they write a `PendingSharePayload` to an App Group `UserDefaults` key and complete immediately — no inference happens inside the extension sandbox. The host app drains the payload on the next foreground transition and hands it to `ChatViewModel.ingestPendingPayload(_:intent:)`.

```swift
// In your App body:
.onChange(of: scenePhase) { _, newPhase in
    if newPhase == .active { checkForPendingSharePayload() }
}
```

A companion `.task(id:)` modifier handles the cold-launch race where `scenePhase` fires `.active` before the SwiftData container finishes initialising. See `docs/share-action-extension-recipe.md` for the full integration guide.

### DIFF
--- a/Example/BaseChatDemo.xcodeproj/project.pbxproj
+++ b/Example/BaseChatDemo.xcodeproj/project.pbxproj
@@ -46,6 +46,13 @@
 		BF0026000000000000000000 /* SetReminderIntent.swift in Sources */ = {isa = PBXBuildFile; fileRef = F10022000000000000000000 /* SetReminderIntent.swift */; };
 		BF0027000000000000000000 /* AppIntentToolsView.swift in Sources */ = {isa = PBXBuildFile; fileRef = F10023000000000000000000 /* AppIntentToolsView.swift */; };
 		BF0028000000000000000000 /* BaseChatAppIntents in Frameworks */ = {isa = PBXBuildFile; productRef = F30008000000000000000000 /* BaseChatAppIntents */; };
+		BF0030000000000000000000 /* PendingSharePayload.swift in Sources */ = {isa = PBXBuildFile; fileRef = F10030000000000000000000 /* PendingSharePayload.swift */; };
+		BF0031000000000000000000 /* ShareViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = F10031000000000000000000 /* ShareViewController.swift */; };
+		BF0032000000000000000000 /* PendingSharePayload.swift in Sources */ = {isa = PBXBuildFile; fileRef = F10030000000000000000000 /* PendingSharePayload.swift */; };
+		BF0033000000000000000000 /* ActionViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = F10034000000000000000000 /* ActionViewController.swift */; };
+		BF0034000000000000000000 /* PendingSharePayload.swift in Sources */ = {isa = PBXBuildFile; fileRef = F10030000000000000000000 /* PendingSharePayload.swift */; };
+		BF0036000000000000000000 /* BaseChatDemoShareExtension.appex in Embed App Extensions */ = {isa = PBXBuildFile; fileRef = F10037000000000000000000 /* BaseChatDemoShareExtension.appex */; settings = {ATTRIBUTES = (RemoveHeadersOnCopy, ); }; };
+		BF0037000000000000000000 /* BaseChatDemoActionExtension.appex in Embed App Extensions */ = {isa = PBXBuildFile; fileRef = F10038000000000000000000 /* BaseChatDemoActionExtension.appex */; settings = {ATTRIBUTES = (RemoveHeadersOnCopy, ); }; };
 /* End PBXBuildFile section */
 
 /* Begin PBXContainerItemProxy section */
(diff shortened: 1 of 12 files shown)


roryford:
Rebased on main (was 13 behind) and addressed the Copilot review.

**Build was broken on this branch as published.** Three blockers; the rest is hygiene from the review.

### Critical fixes

- **`PendingSharePayload` partial initialisers** — optionals now default to `nil` so `PendingSharePayload(kind: .url, urlString: ..., source: ...)` compiles. As filed, this wouldn't have built.
- **macOS build break** — the `BaseChatDemo` target's `SUPPORTED_PLATFORMS` is `iphoneos iphonesimulator macosx`, but the new Embed App Extensions phase + the two new target dependencies were unconditionally attached. Added `platformFilter = ios;` to both build files (`BF0036`, `BF0037`) and both `PBXTargetDependency` entries (`D22001`, `D23001`). `xcodebuild --macos` build-for-testing now succeeds.
- **Cold-launch `.task(id:)` was attached to a Scene** — `WindowGroup`'s chain doesn't have `.task(id:)`. Moved the modifier inside the `Group { ... }` so it runs on the View where it belongs. `.onChange(of: scenePhase)` stays on the Scene where it does work. iOS build now succeeds.

### Hygiene from Copilot's review

- **Doc/code mismatch on `defaults.synchronize()`** — the recipe doc said the flush was important; the code didn't call it. Added `synchronize()` to both extensions after the write.
- **Constant duplication** — promoted the App Group identifier + key to a `DemoSharedAppGroup` enum in `PendingSharePayload.swift` (which is in all three targets). Both extensions now read from it. The host app's `DemoAppGroup` keeps its own literal values with a note explaining why: `InboundPayloadEnvelope.swift` is also compiled into `BaseChatDemoUITests`, which doesn't link the shared payload file.
- **`Kind` priority comment** — was claiming a global `URL > text > image` order; now describes Share-vs-Action priority differences honestly.
- **Extension Info.plists** — added `CFBundleIdentifier`/`CFBundleName`/`CFBundleExecutable`/etc as `$(BUILD_SETTING)` substitutions. With `GENERATE_INFOPLIST_FILE = NO` the embedded-binary validator was failing with `(null)` bundle identifier.

### Deferred (not in this push)

- **Bridged-type casts (`NSString`/`NSURL`/`UIImage`)** — Copilot flagged `loadItem(forTypeIdentifier:) as? URL` as fragile vs `loadObject(ofClass:)`. In Swift these implicitly bridge for the common providers; the recipe is doc-targeted at the simple case. Happy to upgrade if a real provider hits the cast and breaks, but won't preemptively widen it for a pattern doc.

### Pre-push verification

All CI-safe Swift suites green locally:
`BaseChatCoreTests` / `BaseChatInferenceTests` / `BaseChatInferenceSwiftTestingTests` / `BaseChatUITests` / `BaseChatUIModelManagementTests` / `BaseChatMCPTests` / `BaseChatBackendsTests` / `BaseChatTestSupportTests` / `BaseChatAppIntentsTests`. Plus `scripts/example-ui-tests.sh build-for-testing` succeeds for both iOS Simulator and macOS destinations.

Layering: nothing new in `Sources/`. All changes scoped to `Example/` and `docs/`. Recipe shape preserved — pure-Foundation payload + shared App Group `UserDefaults`, no inference in either extension.

PR #876, opened 2026-04-28
docs: add runtime ports refactor plan

roryford:
## Summary
- add an in-repo architecture plan for the proposed `BaseChatRuntime` / ports refactor
- describe the target package graph, migration phases, and explicit non-goals
- fold architect, DevEx, QA, security, and Apple-platform review guidance into one proposal

## Why
This gives the runtime-centered refactor a reviewable design document before code moves across `BaseChatCore`, `BaseChatUI`, and `BaseChatUIModelManagement`. It also frames the migration around the three host shapes we care about: the BaseChatKit demo app, ChatbotUI-iOS, and Fireside.

## Notes
- docs-only change
- plan, not implementation

PR #878, opened 2026-04-28
chore: spike session-list runtime + adapter pattern (do not merge)

roryford:
## Summary

Spike validating the runtime-as-events + thin-@Observable-adapter pattern proposed in PR #876 (`docs/runtime-ports-refactor.md`) on one concrete slice: `SessionManagerViewModel` / `SessionListService`.

This PR **must not be merged**. It is a technical probe to answer whether the proposed architecture is achievable before the 7-phase refactor is approved.

---

### Sync-port-via-async-use-case: did it work? Any awkward seams?

- **It works, but the async surface is cosmetic on this slice.** `ChatPersistenceProvider` is `@MainActor`-isolated and sync-throws. Every service command annotated `async` still runs synchronously on main — the `async` keyword just signals "this is a command you can await" without providing any actual concurrency benefit. No threads are freed; no work leaves main.
- **The critical seam is the sync-vs-async state gap.** The adapter's event consumer is a `Task` that processes events asynchronously. Any test (or view) that reads `sessions` synchronously after a mutation — e.g. `createSession()` then immediately checking `sessions.count` — sees stale state unless the VM also updates its state eagerly. This forced a dual-write pattern: the VM updates `sessions` eagerly (synchronously via `reloadSessionsEagerly()`) AND the service emits a `sessionsLoaded` event that the consumer will process later. The event stream and the observable state are briefly out of sync. In a full async-surfaced API (where `createSession` is `async throws`) the dual-write disappears.

### Event stream: how many event cases ended up needed?

- **7 cases** in `SessionListEvent`: `.sessionsLoaded`, `.sessionInserted`, `.sessionRenamed`, `.sessionDeleted`, `.searchResultsChanged`, `.titleGenerated`, `.persistenceFailure`.
- **`.sessionInserted` is almost a phantom case.** Every insert is immediately followed by a `.sessionsLoaded` reload, and the adapter ignores `.sessionInserted` in `apply()` (the reload does all the work). It exists for tests and potential future fine-grained consumers, but in practice it's always paired with a reload. This is a signal that insert-then-reload could be collapsed into a single `.sessionMutated` event.
- **`.searchResultsChanged` naturally bundles title + message results**, which avoids a race between two separate events. No cases wanted to be a command-response pair — `clearSearch()` is fire-and-forget with an immediate event, which is clean.

### Adapter LOC

- Original `SessionManagerViewModel.swift`: **375 LOC**
- New `SessionManagerViewModel.swift` (adapter): **316 LOC**
- New `SessionListService.swift` (orchestration): **329 LOC**
- `SessionListServiceTests.swift` (new): **124 LOC** (not counted against adapter delta)
- **Total LOC delta for source files: +270 LOC** (316 + 329 − 375). The adapter is slightly smaller than the original, but the total surface grew because the service is a new file. Comments and SPIKE NOTE markers inflate both files; in a real refactor those would be trimmed.

### Title generation: did it migrate cleanly?

- **Partially.** The `autoRenameSession` path (inference-backed) emits `.titleGenerated` cleanly and the adapter applies it via the event consumer. This is the "ideal shape."
- **`autoGenerateTitle` (sync, first-message truncation) required special handling.** Because its public contract must stay `func` (non-async), and because the adapter's event consumer is async, the sync path needed `reloadSessionsEagerly()` to keep `sessions` up to date for callers that check it synchronously. The dual-write comment in the adapter code documents this explicitly.
- **The deeper blocker for AI title generation:** `InferenceService` is `@Observable @MainActor`, meaning `enqueue(...)` is also `@MainActor`. Moving title generation off main requires the inference API itself to grow a non-main-actor surface — a prerequisite change that isn't in scope for the session-list spike but would be required before a `BaseChatRuntime` target could genuinely run orchestration off main.

### Verdict: would this scale to `ChatViewModel` (3,621 LOC across file + 9 extensions)?

**Conditionally yes, but three prerequisites must be resolved first:**

1. **Make public command APIs `async throws`.** The sync-vs-async dual-write anti-pattern (documented in `reloadSessionsEagerly()`) exists entirely because the original public API is synchronous. `ChatViewModel.sendMessage()` is already `async`, so the most expensive methods are already async. The remaining sync mutators (`clearMessages`, `switchToSession`, etc.) would need to become `async` to eliminate the dual-write. This is a BREAKING change.
2. **Resolve `InferenceService` main-actor coupling.** Title generation, model loading, and streaming all require `@MainActor` today. Until `InferenceService` grows a non-main-actor generation surface, the runtime services stay pinned to main and the `async` label is purely organizational.
3. **Accept a larger total LOC budget.** The session-list slice grew by ~270 LOC despite being a relatively simple 375-line class. `ChatViewModel` at 3,621 LOC would likely produce 4,500–5,000 LOC across the service + adapter split, plus tests. The split buys genuine testability improvements (the service is testable without a view model), but the LOC growth is real and should be planned for.

**Bottom line:** the pattern is achievable and produces better-isolated, more testable code. The event-stream shape works well. The main cost is resolving the sync/async seam — without that, the adapter ends up doing double-writes that undermine the "single source of truth" goal of the event stream.

### DIFF
--- a/Sources/BaseChatUI/ViewModels/SessionListService.swift
+++ b/Sources/BaseChatUI/ViewModels/SessionListService.swift
@@ -1,0 +1,329 @@
+import Foundation
+import BaseChatCore
+import BaseChatInference
+
+// MARK: - Event surface
+
+/// Sendable events emitted by ``SessionListService``.
+///
+/// All state transitions surface here rather than being written directly to
+/// observable state — the adapter (``SessionManagerViewModel``) subscribes and
+/// mirrors them into its published properties.
+public enum SessionListEvent: Sendable {
+    case sessionsLoaded([ChatSessionRecord], hasMore: Bool)
+    case sessionInserted(ChatSessionRecord)
+    case sessionRenamed(UUID, title: String)
+    case sessionDeleted(UUID)
+    case searchResultsChanged(SearchResults)
+    case titleGenerated(UUID, title: String)
+    case persistenceFailure(any Error)
+}
+
+/// Snapshot of search state emitted in `.searchResultsChanged`.
+public struct SearchResults: Sendable {
+    public let titleMatches: [ChatSessionRecord]
+    public let messageHitsBySession: [UUID: [MessageSearchHit]]
+    public let messageMatchSessions: [ChatSessionRecord]
+
+    public static let empty = SearchResults(
+        titleMatches: [],
+        messageHitsBySession: [:],
+        messageMatchSessions: []
+    )
+}
+
+// MARK: - Service
+
+/// Orchestrates session CRUD and search. Plain class — no `@Observable`,
+/// no `@MainActor` pin on the type itself.
+///
+/// Commands are `@MainActor` because `ChatPersistenceProvider` is `@MainActor`-
+/// isolated and all callers already run on the main actor. The `async` on
+/// public commands is intentional: it signals that commands should be awaited
+/// (rather than fire-and-forget) and preserves the shape the plan describes for
+/// a future `BaseChatRuntime` target where commands *would* hop off main.
+///
+/// SPIKE NOTE (isolation seam): Because `ChatPersistenceProvider` is `@MainActor`
+/// and sync, moving any of these commands truly off-main would require either
+/// (a) making the persistence port async, or (b) wrapping the provider in a
+/// dedicated actor. The current shape validates event emission and adapter
+/// separation without that infrastructure change. See SPIKE FINDINGS in the PR.
+public final class SessionListService: Sendable {
+
+    public let events: AsyncStream<SessionListEvent>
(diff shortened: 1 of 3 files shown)


roryford:
Closing — the spike has done its job.

Phase 1.0 (the async-migration prerequisite the spike surfaced) shipped in #883 and was tightened in #885 (`autoLoad` made explicit on `SessionManagerViewModel.configure`). Phase 1.1 (productionising this spike's pattern with the dual-write workaround removed) is in flight as a follow-up PR.

Findings folded into the runtime-ports plan (#876):
- Persistence passed at init — no `nonisolated(unsafe)`.
- Adapter mutations go through service commands — no eager-reload dual-write.
- `.sessionInserted` + `.sessionsLoaded` collapsed.
- `InferenceService` main-actor coupling left for Phase 1.2.

Not merging.

PR #879, opened 2026-04-28
feat: model selection sort + TTFT perf test SIGTRAP fix

roryford:
## Summary

Three commits, all on this branch ahead of `main`:

- **`feat(model-management)`** — Add a sort-order picker to the Model Selection tab (`Alphabetical`, `Type`, `Size (Smallest First)`, `Capability / Speed`). Sort logic is extracted as `static ModelSelectionTabView.sortModels(_:by:)` for direct unit testing.
- **`fix(tests)`** — Eliminate a SIGTRAP at process exit in `BaseChatUITests`. `IntegratedStreamingPerformanceTests.testPerf_timeToFirstToken_realisticBackend` was fulfilling its expectation before draining the unstructured `sendMessage()` task, so the in-flight task could outlive `tearDown()` and access an invalidated `ModelContext`. Now uses a `firstTokenExp` (timed via `stopMeasuring()`) plus a separate `drainExp` to ensure the task fully completes inside the measure block.
- **`fix`** — Pre-existing demo UX surface improvements (sidebar/memory indicator/connected services/`DemoNowTool`). Already on local `main`.

## Test plan

- [x] Full CI-safe suite (Core / Inference / InferenceSwiftTesting / UI / UIModelManagement / MCP / Backends / TestSupport / AppIntents) passes
- [x] `BaseChatE2ETests` passes
- [x] BaseChatUITests no longer crashes with signal 5 at process exit (verified by running suite in isolation after the fix — 450/450 tests pass cleanly, no SIGTRAP)
- [x] Sort picker covered by 4 unit tests (one per order) + 1 snapshot test asserting the picker renders in the Select tab

### DIFF
--- a/Example/BaseChatDemo.xcodeproj/project.pbxproj
+++ b/Example/BaseChatDemo.xcodeproj/project.pbxproj
@@ -46,6 +46,7 @@
 		BF1012000000000000000000 /* BaseChatInference in Frameworks */ = {isa = PBXBuildFile; productRef = F30007000000000000000000 /* BaseChatInference */; };
 		BF0026000000000000000000 /* SetReminderIntent.swift in Sources */ = {isa = PBXBuildFile; fileRef = F10022000000000000000000 /* SetReminderIntent.swift */; };
 		BF0027000000000000000000 /* AppIntentToolsView.swift in Sources */ = {isa = PBXBuildFile; fileRef = F10023000000000000000000 /* AppIntentToolsView.swift */; };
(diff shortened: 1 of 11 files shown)


roryford (commented):
Sort comparator and TTFT drain pattern look correct. The snapshot assertion as originally written fails locally (`Swift.dump()` doesn't surface List-hosted picker text, and `refreshModels()` clears the seed in the empty-state branch); pushed `3d794d1` to rewrite it against `_sortOrder` / `ModelSelectionSortOrder` state, which the dump does carry.

Worth noting in the commit message that the TTFT fix also corrects a ~7× per-iteration measurement drift (745ms → 104ms avg), not just the SIGTRAP.

PR #886, opened 2026-04-28
refactor: extract SessionListService from SessionManagerViewModel (phase 1.1)

roryford:
## Summary
- Extracts session-list orchestration from `SessionManagerViewModel` into a new `SessionListService` (plain `final class Sendable`, async-throws commands, `AsyncStream<SessionListEvent>` for state).
- `SessionManagerViewModel` becomes an `@Observable @MainActor` adapter that consumes the service's events and re-publishes for SwiftUI.
- No public API change — view-model surface stays source-compatible.

## Why
This is **Phase 1.1** of the runtime ports refactor planned in #876. It validates the closure-bag → events transformation for **write paths**, which the Phase 0 spike (#878) could not do because mutations were sync. Phase 1.0 (#883) made commands `async throws`; this PR is the productionisation that proves the pattern scales before the bigger `ConversationRuntime` extraction in Phase 1.2.

Branch is based off `main` (post-#883). PR #885 (the `autoLoad` follow-up on the `refactor/phase-1-0-followup` branch) is in flight; whichever lands first, the other rebases. Both touch different shapes — this PR is the extraction; #885 is a `configure` signature change.

## Spike findings folded in
- Persistence passed at init — no `nonisolated(unsafe) var _persistence`.
- Adapter mutations go through `service.<command>(...)`. No eager-reload dual-write.
- `.sessionInserted` + `.sessionsLoaded` collapsed (the spike noted these were always paired). `createSession` returns the new record; the adapter sets `activeSession` from the return value.
- `.titleGenerated` event handles auto-rename cleanly. The diagnostics path still distinguishes `.titleGenerationFailed` from `.sessionRenamePersistenceFailed`.
- `InferenceService` main-actor coupling unchanged (Phase 1.2 step 5).

## Event surface

\`\`\`swift
public enum SessionListEvent: Sendable {
    case sessionsLoaded([ChatSessionRecord], hasMore: Bool, offset: Int)
    case sessionRenamed(UUID, title: String)
    case sessionDeleted(UUID)
    case searchResultsChanged(SearchResults)
    case titleGenerated(UUID, title: String)
    case persistenceFailure(any Error)
}
\`\`\`

Six cases (down from the spike's seven). \`offset\` was added to \`.sessionsLoaded\` so adapters can distinguish replace-vs-append without owning their own cursor.

## Service ownership decision

\`final class SessionListService: Sendable\` with persistence, diagnostics, and per-call inference injected at init or per-method. Rejected actor isolation because:

- \`ChatPersistenceProvider\` is \`@MainActor\`-bound; an actor would force \`await\` even for client-side filters that never touch the store.
- \`Sendable\` class with init injection composes cleanly: command bodies hop to \`@MainActor\` via the existing port isolation; \`SessionListService\` itself stays unpinned at the type level (matching the plan's principle 7).

## Phase 1.1 isolation seam (open question)

The plan says "use cases are not \`@MainActor\`-pinned at the type level" and "methods that hop into \`@MainActor\` ports hop on demand, with \`await\`." Strictly read, that means service methods are \`nonisolated\` and call \`await persistence.x()\` to hop.

In Phase 1.1 that breaks the \`await vm.createSession(...); XCTAssertEqual(vm.sessions.count, 1)\` test pattern: a long-lived consumer \`Task\` consuming an \`AsyncStream\` cannot drain synchronously relative to a returning command on the same actor — the test asserts before the consumer runs.

The shipped shape pins **command bodies** (not the type) on \`@MainActor\` and routes events through a synchronous sink the adapter installs at \`configure\`. The public \`AsyncStream<SessionListEvent>\` is also yielded so external observers (and Phase 1.2 cross-actor adapters) see the same surface; the adapter's long-lived consumer \`Task\` drains it. Phase 1.2 step 5 lifts \`InferenceService\` off main, the service drops \`@MainActor\` from command bodies, and the adapter switches to the AsyncStream path — no public API change required.

This is a deliberate Phase 1.1 ⇄ Phase 1.2 trade-off; flagging it explicitly so reviewers can push back before merge.

## LOC delta

| File | Before | After | Delta |
|---|---|---|---|
| \`SessionManagerViewModel.swift\` | 383 | 330 | −53 |
| \`SessionListService.swift\` (new) | 0 | 425 | +425 |
| \`SessionListServiceTests.swift\` (new) | 0 | 381 | +381 |
| **Net** | | | **+753** |

The plan budgeted ~+270 LOC based on the spike. The +753 here is mostly the new \`SessionListServiceTests\` (381 LOC, 16 tests) plus the explicit ownership/isolation docstrings on the service. Test coverage at the new boundary is the right place to spend LOC for a contract that's about to absorb \`ConversationRuntime\` in Phase 1.2.

## Test plan
- [x] Full CI-safe pre-push suite passes locally (124s wallclock for the full suite; only the pre-existing \`LargeSessionListPerformanceTests/test_perf_initialFirstPageRender\` failure on \`main\` remains, which is the subject of #885's followup).
- [x] \`SessionListServiceTests\` (new, 16 tests) covers CRUD, search, pagination, title generation, the synchronous-sink path, and the public \`AsyncStream\` surface directly against the service — no view model.
- [x] \`SessionManagerViewModelTests\` (15), \`SessionAutoRenameTests\` (8), \`SessionManagerSearchPaginationTests\` (13), \`PersistenceGuardTests\` (12) pass unchanged.
- [x] Sabotage check: temporarily emitted \`.sessionRenamed("SABOTAGE")\` in place of \`.sessionDeleted\` from \`deleteSession\`. \`test_deleteSession_emitsSessionDeletedThenSessionsLoaded\` correctly failed; reverted.

## Open questions

- **Isolation seam (above).** Reviewer call on whether the \`@MainActor\` per-command isolation is acceptable as a Phase 1.1 expedient or whether the plan should be adjusted.
- **\`persistenceFailure\` event surfacing.** The adapter currently logs and otherwise drops the event. Phase 1.2 will likely surface a UI banner / retry affordance — left as a no-op here rather than inventing a state shape that gets thrown away.
- **\`SessionListService\` visibility.** Currently \`internal\` to \`BaseChatUI\`. Phase 1.2 may widen to \`package\`/\`public\` once the surface stabilises further. Did not pre-emptively widen.

### DIFF
--- a/Sources/BaseChatUI/ViewModels/SessionListService.swift
+++ b/Sources/BaseChatUI/ViewModels/SessionListService.swift
@@ -97,16 +97,23 @@ final class SessionListService: Sendable {
     /// Upper bound on sessions resolved when surfacing message-search hits.
     static let messageSearchSessionResolveCap: Int = 10_000
 
-    /// Async stream of state transitions for external observers (tests,
-    /// debugging tools, future runtime adapters). The adapter
-    /// (`SessionManagerViewModel`) does not consume this stream — it
-    /// installs a synchronous sink via ``setEventSink(_:)`` so state changes
-    /// land in observable state in the same actor hop as the command that
-    /// produced them.
+    /// Async stream of state transitions. The adapter
+    /// (`SessionManagerViewModel`) installs a synchronous sink via
+    /// ``setEventSink(_:)`` so state changes land in observable state in the
+    /// same actor hop as the command that produced them; the adapter also
+    /// runs a long-lived `Task` that drains this stream so its unbounded
+    /// buffer does not grow.
+    ///
+    /// `AsyncStream` is single-consumer by design: tests that drive the
+    /// service directly (no adapter attached) iterate this stream
+    /// themselves. Once an adapter is attached, the adapter's drain task
+    /// owns the consumer slot and external `for await` loops over
+    /// `service.events` will not observe events. Either drive the service
+    /// without an adapter or install your own sink via ``setEventSink(_:)``.
     ///
     /// Phase 1.2 will move the runtime off `@MainActor`; at that point the
-    /// adapter switches to consuming this stream from a long-lived `Task`
-    /// because cross-actor delivery cannot be synchronous.
+    /// adapter switches to consuming this stream from its drain `Task` (the
+    /// sink path becomes a cross-actor hazard) without changing this surface.
     let events: AsyncStream<SessionListEvent>
     private let continuation: AsyncStream<SessionListEvent>.Continuation
 
@@ -203,8 +210,10 @@ final class SessionListService: Sendable {
     }
 
     /// Loads the next page starting at `offset` and emits `.sessionsLoaded`.
-    /// No-op when persistence raises an error — the failure is logged and
-    /// surfaced as `.persistenceFailure`.
+    /// On persistence failure, emits an empty `.sessionsLoaded(_, hasMore: false, offset:)`
+    /// so the adapter clears `hasMoreSessions` (matching Phase 1.0 behaviour
+    /// where a transient page-load failure stopped further pagination), then
+    /// emits `.persistenceFailure` for diagnostics consumers.
     @MainActor
     func loadNextPage(offset: Int) async {
         do {
@@ -213,6 +222,10 @@ final class SessionListService: Sendable {
             emit(.sessionsLoaded(next, hasMore: hasMore, offset: offset))
         } catch {
             Log.persistence.error("Failed to load next sessions page: \(error)")
+            // Reset hasMoreSessions on the adapter so the failed offset is not
+            // retried automatically. Pre-Phase-1.1 behaviour; restoring it
+            // keeps the public surface source-compatible.
+            emit(.sessionsLoaded([], hasMore: false, offset: offset))
             emit(.persistenceFailure(error))
         }
     }
--- a/Sources/BaseChatUI/ViewModels/SessionListService.swift
+++ b/Sources/BaseChatUI/ViewModels/SessionListService.swift
@@ -1,0 +1,425 @@
+import Foundation
+import BaseChatCore
+import BaseChatInference
+
+// MARK: - Event surface
+//
+// The service emits state transitions as events; the adapter
+// (`SessionManagerViewModel`) consumes them and re-publishes for SwiftUI.
+//
+// `.sessionInserted` from the Phase 0 spike (#878) was always paired with a
+// reload; it is collapsed into `.sessionsLoaded`. Adapters that need the new
+// record's identity get it from the `createSession(...)` return value.
+//
+// `.titleGenerated` carries both the session id and the resolved title so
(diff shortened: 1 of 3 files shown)


roryford (commented):
## Architect review — Phase 1.1 ship-with-fixes

Read the diff in full plus the plan (#876), the spike (#878), and Phase 1.0 (#883). Re-ran the full nine-suite pre-push gate locally; only the pre-existing `LargeSessionListPerformanceTests/test_perf_initialFirstPageRender` failure remains, which predates this branch.

### What I fixed (commit 2f1852f)

- **`loadNextPage` failure-path regression.** Phase 1.0 reset `hasMoreSessions = false` on a transient page-load error so the user wasn't silently stuck retrying. This PR collapsed that to `.persistenceFailure` only, leaving the flag asserted from the prior successful page. Restored parity by emitting a reset `.sessionsLoaded([], hasMore: false, offset:)` ahead of the diagnostic event. Added `test_loadNextPage_persistenceFailure_emitsResetThenFailure` driving through `ErrorInjectingPersistenceProvider`; sabotage-verified.
- **Misleading doc comments on `events`.** `AsyncStream` is single-consumer; the prior comments claimed "external observers see the full surface even though the adapter does not rely on them." Once the adapter attaches its drain task, that consumer slot is owned. Reworded to call this out: tests that drive the service without an adapter iterate `events` themselves; otherwise install a sink. Phase 1.2 still works because the adapter's drain task simply *becomes* the real consumer.

### Open questions — my position

1. **Isolation seam.** Acceptable as a Phase 1.1 expedient. Pinning command bodies on `@MainActor` (rather than the type) is a literal-but-defensible reading of plan principle 7, and the migration shape — sink today, drain task takes over in Phase 1.2 — is coherent without breaking the public surface. I would not adjust the plan; I would call out in the Phase 1.2 PR description that step 5 (`InferenceService` off-main) is the precondition that lets the `@MainActor` annotations come off the command bodies.
2. **`persistenceFailure` no-op.** Right call. Surfacing it requires a state shape (banner / retry affordance) that should be designed in the PR that *needs* it; inventing a hook now and tearing it out later is worse than the silence. Note: this is now strictly diagnostic — the `loadNextPage` reset above means the adapter's pagination state stays coherent without consuming the event.
3. **Visibility (`internal`).** Right call. No external consumer today; widening pre-emptively would lock in a surface that's still mid-extraction.

### What I deliberately didn't fix

- The `.sessionsLoaded`/`offset` collapse holds across CRUD + search + pagination: every load path now emits a coherent offset, the adapter's `apply` correctly replaces vs. appends, and the dedupe filter on append survives.
- The drain task's no-op `for await _` is fine for Phase 1.1 — bounding the buffer to discard-newest would change semantics for tests that depend on full event history.
- `Task.yield()` in `test_eventsStream_observesEvents_inOrder` is OK here: the stream's unbounded buffer makes the yield a comfort, not a load-bearing race fix.

All pre-push CI-safe suites green at 2f1852fb9ad0a57309d01c890e3fbc06b4802210.

Issue #445, opened 2026-04-18
feat(voice): BaseChatVoice module — speech recognition, synthesis, and UI components


roryford:
## Motivation

Voice-in / voice-out is a natural on-device capability that fits BCK's local-first posture. Building it as an opt-in module (package trait) keeps audio-session and entitlement concerns out of the default dependency graph.

## Proposal

New optional SPM module `BaseChatVoice`, behind a package trait:

- `WakeWordDetector` protocol + Apple-stack implementation using `SFSpeechRecognizer` + `AVAudioEngine` tap
- `SpeechTranscriber` for turn capture
- `SpeechSynthesizer` wrapper around `AVSpeechSynthesizer` with cancellation support
- Simulator guards — skip audio engine setup on simulator

No third-party APIs. No cloud TTS. Host apps can compose their own TTS behind the `SpeechSynthesizer` protocol if needed.

## Acceptance criteria

- Opt-in via package trait; default graph unchanged
- DocC guidance for `NSMicrophoneUsageDescription` + `NSSpeechRecognitionUsageDescription`
- Mic / audio-session arbitration documented
- Unit tests with a mock audio source

## Folded-in scope (from #446)

Companion SwiftUI components ship in the same module/PR — the UI has no use without the underlying voice stack, so they land together:

- `VoiceInputButton` — press-to-talk button with haptics and accessibility labels
- `LiveTranscriptionView` — in-progress transcript rendering bubble
- `WakeWordToast` — unobtrusive indicator when the wake word is detected

Additional acceptance criteria:

- Components work on iOS and macOS
- Respects reduce-motion and VoiceOver
- Example-app demo wired up

roryford:
Shipped via #887 — `BaseChatVoice` module under the `Voice` trait covers every acceptance criterion:

- `WakeWordDetector` → `Sources/BaseChatVoice/Services/AppleWakeWordDetector.swift`
- `SpeechTranscriber` → `AppleSpeechTranscriber.swift`
- `SpeechSynthesizer` → `AppleSpeechSynthesizer.swift`
- `VoiceInputButton`, `LiveTranscriptionView`, `WakeWordToast` → `Sources/BaseChatVoice/Views/`
- Trait-gated (default graph unchanged), demo-app wired, permission strings + simulator limits documented

Released in v0.14.0.

PR #887, opened 2026-04-28
feat(voice): add BaseChatVoice speech I/O module

roryford:
Refs #445

## Summary
- add the trait-gated BaseChatVoice speech I/O module and composer-accessory integration seam
- add wake-word phrase detection, WakeWordToast, and demo-app voice accessory wiring
- document Voice trait setup, required permission strings, simulator limits, and audio-session behavior

## Validation
- swift test --filter BaseChatVoiceTests --disable-default-traits --traits Voice
- swift test --filter ChatComposerAccessoryMigrationGuardTests --disable-default-traits --traits Voice
- swift build --product BaseChatVoice --disable-default-traits --traits Voice

## Notes
- attempted macOS and iOS Simulator demo-app builds with xcodebuild, but current main is already blocked by the pre-existing `DemoTools` -> `DemoNowTool` compile error in `Example/BaseChatDemo/DemoTools.swift`

### DIFF
--- a/Example/BaseChatDemo.xcodeproj/project.pbxproj
+++ b/Example/BaseChatDemo.xcodeproj/project.pbxproj
@@ -48,6 +48,7 @@
 		BF0027000000000000000000 /* AppIntentToolsView.swift in Sources */ = {isa = PBXBuildFile; fileRef = F10023000000000000000000 /* AppIntentToolsView.swift */; };
 		BF0029000000000000000000 /* DemoNowTool.swift in Sources */ = {isa = PBXBuildFile; fileRef = F10024000000000000000000 /* DemoNowTool.swift */; };
 		BF0028000000000000000000 /* BaseChatAppIntents in Frameworks */ = {isa = PBXBuildFile; productRef = F30008000000000000000000 /* BaseChatAppIntents */; };
+		BF0029000000000000000000 /* BaseChatVoice in Frameworks */ = {isa = PBXBuildFile; productRef = F30009000000000000000000 /* BaseChatVoice */; };
 		BF0030000000000000000000 /* PendingSharePayload.swift in Sources */ = {isa = PBXBuildFile; fileRef = F10030000000000000000000 /* PendingSharePayload.swift */; };
 		BF0031000000000000000000 /* ShareViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = F10031000000000000000000 /* ShareViewController.swift */; };
 		BF0032000000000000000000 /* PendingSharePayload.swift in Sources */ = {isa = PBXBuildFile; fileRef = F10030000000000000000000 /* PendingSharePayload.swift */; };
@@ -141,6 +142,7 @@
 				BF0006000000000000000000 /* BaseChatTools in Frameworks */,
 				BF0020000000000000000000 /* BaseChatMCP in Frameworks */,
 				BF0028000000000000000000 /* BaseChatAppIntents in Frameworks */,
+				BF0029000000000000000000 /* BaseChatVoice in Frameworks */,
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 		};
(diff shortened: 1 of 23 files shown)


roryford (commented):
## Architect review — fixes pushed, CI should now go green

### What was red and why

Two distinct failures, neither caused by the BaseChatVoice changes themselves:

1. **`test` job — `SilentCatchAuditTest`.** New `try?` in `VoiceAudioSessionCoordinator.deactivateRecording()` (best-effort `AVAudioSession.setActive(false, …)`) tripped the audit. This is a legitimate best-effort cleanup — iOS routinely tears the session down out from under us on backgrounding / interruption-ended / route-change, and the throwing form only surfaces a benign "session not active" error. Allowlisted with rationale matching the file's existing entry format (`b43d775`).

2. **`saas` job — `extra argument 'autoLoad' in call`.** Pre-existing regression on `main` from PR #885 (`c6aa115`). #885 added `autoLoad: false` to every `configure(persistence:)` call site, but `ChatViewModel.configure(persistence:)` never gained the new parameter — only `SessionManagerViewModel.configure` did. The CloudSaaS matrix leg is the only one that compiles `CloudEndpointSelectionIntegrationTests.swift`, so the breakage slipped through #885's CI. `main`'s build-modes saas job has been red since `c6aa115` landed. Fixed on this branch by dropping the bogus `autoLoad:` arg from the two `viewModel`/`freshVM.configure` calls (`7cafb59`); a sibling fix should land on `main` independently.

I rebased the PR onto `origin/main` to bring in `c6aa115` so the fix is local to this PR's diff (not lost behind a future merge).

### What I checked, beyond the failures

- **Trait gating.** `BaseChatVoice` is unconditionally a product (`.library`) and the `Voice` trait only gates an internal `#if Voice` define. Demo wiring uses `import BaseChatVoice` unconditionally. This is fine: the module compiles trait-free (the audio/Speech APIs are part of the OS SDK, not a third-party dep), and the trait-gated define is reserved for internal capability checks. No `#if Voice` is presently used in module sources — that's OK for a spike but worth tightening before the module exits spike status.
- **Module boundary.** `BaseChatVoice` depends on `BaseChatUI` (consumes `ChatViewModel` via `@Environment` in `VoiceComposerAccessory`). No cycle introduced; `BaseChatUI` does not gain a `BaseChatVoice` import.
- **Concurrency.** `@MainActor` on the controller, transcriber, synthesizer, wake-word detector. `Sendable` on the value types (`SpeechTranscriptionUpdate`, `WakeWordDetection`, `VoiceAuthorizationStatus`). Continuations correctly drained in `AppleSpeechSynthesizer`. No `assertionFailure` in recoverable paths.
- **Simulator handling.** `AppleSpeechTranscriber.startTranscribing` throws `VoiceError.simulatorUnsupported` under `#if targetEnvironment(simulator)` before touching `AVAudioEngine`. CI macOS runner is unaffected.
- **Composer-accessory seam.** `composerAccessory:` is a `@ViewBuilder` closure on `ChatView` (matches the existing `apiConfiguration:` pattern). Future TTS/dictation features can plug additional accessories alongside the voice one without changing `ChatView`'s contract. Correct shape.
- **Wake-word detector.** Matching is whitespace-padded normalized substring containment, with `hasTriggered` gating to prevent re-triggers within a session. Tests cover idempotency + reset + blank-phrase rejection. False-positive coverage (e.g., `"hey base chats"` should not match `"hey base chat"`) and partial-prefix coverage are missing — not a blocker for a spike, but worth adding as the module hardens.

### Pre-push test gate

All CI-safe suites green at `b43d775a12c762eb342e27a2dec30efa866d861a`:

- `BaseChatCoreTests` (283 tests, 0 failures)
- `BaseChatInferenceTests` (1331 tests, 0 failures — `SilentCatchAuditTest` now green)
- `BaseChatInferenceSwiftTestingTests` (83 tests)
- `BaseChatUITests` (453 tests)
- `BaseChatUIModelManagementTests` (100 tests)
- `BaseChatMCPTests` (130 tests)
- `BaseChatBackendsTests` (16 tests, default traits)
- `BaseChatTestSupportTests`
- `BaseChatAppIntentsTests` (11 tests)
- **saas leg:** `BaseChatBackendsTests --traits CloudSaaS` (76 tests, was red, now green)
- **Voice leg:** `BaseChatVoiceTests --traits Voice` (14 tests)

### Verdict

Ship-with-fixes. The substantive review (boundaries, gating, concurrency, simulator) is clean. CI should be green on the next push; the `saas` recovery for `main` itself should land as a separate one-line follow-up so `main` stops carrying a known-red build-modes job.

All pre-push CI-safe suites green at `b43d775a12c762eb342e27a2dec30efa866d861a`.

PR #891, opened 2026-04-29
ci: cut per-PR macOS minutes via build-modes + fuzz-pr cache

roryford:
## Summary

Two structural CI changes that cut macos-15 (10× billing) consumption per PR.

- **`build-modes.yml`: trim PR matrix to `[saas, full]`; keep all four `[offline, ollama, saas, full]` on the nightly cron.** The `full` mode is a strict trait superset (`MLX,Llama,Ollama,CloudSaaS`) that catches every compile-time regression the offline/ollama legs would. The `saas` leg is retained on PRs because it carries a `CloudSaaS`-only test step (`BaseChatBackendsTests --traits CloudSaaS`) that is real signal not covered by `ci.yml`. The full 4-mode build+audit matrix continues nightly so the procurement-evidence requirement is unchanged. Recent PR-triggered runs took ~14 min wall × 4 legs of macos-15; trimming to 2 legs saves ~28 macos-min on every qualifying PR.
- **`fuzz-pr.yml`: align cache key with `ci.yml`, drop `.build/debug` from cached paths.** Today fuzz-pr's cache key is `xcode-26.3-fuzz-spm-…` while ci.yml's is `xcode-26.3-spm-…`, so the two macos jobs on the same PR rebuild from cold. Aligning the primary key lets fuzz-pr restore the SwiftPM artifacts ci.yml just produced. `.build/debug` is dropped because ci.yml already documented that caching it produces 0 incremental savings on ephemeral runners and wastes the 10 GB per-repo quota. Recent fuzz-pr runs were ~76s of which ~70s was SwiftPM rebuild; expect that to drop substantially when the cache is shared.

Net per-PR macOS savings on a typical qualifying PR (touches `Sources/BaseChatBackends/**`): roughly 30 macos-15-minutes (10× billed = 300 min equivalent).

## Test plan

- [ ] CI runs green on this PR.
- [ ] Confirm `build-modes.yml` PR job runs 2 matrix legs (saas + full), not 4.
- [ ] Confirm a follow-up PR's `fuzz-pr` run hits the shared cache (look for `Cache restored from key:` line referencing the `xcode-26.3-spm-` prefix).
- [ ] Sanity-check that the next nightly `build-modes` schedule still runs all four modes by inspecting `gh run list --workflow build-modes.yml --event schedule --limit 1` after the next 05:30 UTC tick.

### DIFF
--- a/.github/workflows/build-modes.yml
+++ b/.github/workflows/build-modes.yml
@@ -11,6 +11,18 @@ name: Build modes + binary symbol audit
 #    on-demand trigger — it only fires when a PR actually edits trait-gated
 #    source, Package.swift, or this workflow's own configuration. Doc-only
 #    PRs and unrelated source edits don't trigger it.
+#  - **PR vs schedule matrix split.** On `pull_request`, the matrix runs only
+#    `[saas, full]`; on `schedule` (and `workflow_dispatch`) it runs all four
+#    `[offline, ollama, saas, full]`. Rationale: `full` enables the strict
+#    trait superset (`MLX,Llama,Ollama,CloudSaaS`) so any compile-time
+#    regression that the offline or ollama legs would catch also breaks the
+#    full leg — running them on every qualifying PR is duplicate macos-15
+#    spend. `saas` is retained on PRs because it carries a CloudSaaS-only
+#    test step (`BaseChatBackendsTests --traits CloudSaaS`) that is real
+#    signal not covered by ci.yml. The full 4-mode build+audit matrix
+#    continues nightly so the procurement-evidence requirement (90-day
+#    per-mode artifacts) is unchanged. Saves ~28 macos-min per qualifying PR
+#    (2 legs × ~14 min wall instead of 4).
 #  - All four modes run on `macos-15` (10x billing): the package depends on
 #    `os` (Apple-only) and SwiftUI throughout `BaseChatUI`/
 #    `BaseChatUIModelManagement`, so a Linux compile leg is structurally
@@ -55,7 +67,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        mode: [offline, ollama, saas, full]
+        # On PRs: only saas + full. `full` is a strict trait superset that
+        # catches every compile-time regression the offline/ollama legs would;
+        # `saas` carries a CloudSaaS-only test slice not covered by ci.yml.
+        # On schedule/workflow_dispatch: all four modes for the procurement-
+        # evidence artifact set. See the cost-discipline block at the top.
+        mode: ${{ github.event_name == 'pull_request' && fromJSON('["saas", "full"]') || fromJSON('["offline", "ollama", "saas", "full"]') }}
 
     steps:
       - name: Checkout
--- a/.github/workflows/fuzz-pr.yml
+++ b/.github/workflows/fuzz-pr.yml
@@ -38,17 +38,27 @@ jobs:
           xcode-version: "26.3"
 
       - name: Cache SwiftPM build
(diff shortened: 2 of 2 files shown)


roryford:
## Independent review

Verified against PR head `ci/macos-cost-reduction` via `gh api contents` (no working-tree dirtying) and cross-checked CI: **2 build-modes legs ran (saas 8m07s + full 8m26s)**, fuzz pass 1m51s, all green.

### Strengths
- Matrix conditional **works in practice** — CI proves the `${{ … && fromJSON(…) || fromJSON(…) }}` pattern is evaluated correctly for PR triggers (only saas + full ran). This is the [documented GHA pattern for dynamic matrices](https://docs.github.com/en/actions/learn-github-actions/expressions#example) and both branches return arrays of the same shape, so no surprises.
- `workflow_dispatch` correctly falls through to the 4-mode list (since `event_name != 'pull_request'`), preserving the manual-trigger path.
- The `if: matrix.mode == 'saas'` CloudSaaS test step at L94 is preserved on PR runs because the `saas` leg is in the PR matrix — no signal lost.
- **fuzz-pr cache key alignment is exact**: ci.yml primary key is `${{ runner.os }}-${{ runner.arch }}-xcode-26.3-spm-${{ hashFiles('Package.resolved') }}` and fuzz-pr.yml now matches byte-for-byte — caches share the bucket.
- `restore-keys` order in fuzz-pr.yml correctly puts the new `xcode-26.3-spm-` prefix above the legacy `xcode-26.3-fuzz-spm-` prefix (closest match first). Symmetric with ci.yml's existing restore-keys, which already list both prefixes.
- No required status check (`test`, `changelog-lint`) or downstream artifact consumer references the dropped offline/ollama legs by name — checked CODEOWNERS, branch protection, and grepped `.github/`.

### Concerns / questions
- **build-modes.yml L76** — cache key still includes `${{ matrix.mode }}`. With offline/ollama removed from PR runs, those two cache entries only get refreshed by the nightly schedule. GHA evicts caches after **7 days unaccessed**, so a Monday nightly will keep Tue–Sun warm — fine in steady state, but if a nightly is skipped (e.g. CI outage) the next nightly cold-builds two legs. Low impact, worth a note.
- **fuzz-pr.yml cache write race** — when fuzz-pr (~2m) and ci.yml (~8m+) run concurrently on the same PR with the same primary key, fuzz-pr will finish first and upload a smaller artifact set (no test-build outputs in `.build/checkouts`? actually checkouts are source clones, identical between jobs, so probably benign). The cached `.build/checkouts` is the source-clone phase regardless of trait flags, so MLX/Llama checkouts will be present from ci.yml's default-traits build even when fuzz-pr restores them under `--disable-default-traits` — harmless extra disk, not a correctness issue.
- **`.build/debug` removal in fuzz-pr.yml**: confirmed `swift run` defaults to debug, so this is currently a real regression-of-cache for the fuzz binary. PR description claims ci.yml documented this as 0 incremental savings — true on ephemeral runners (absolute-path fingerprints invalidate). Acceptable, but if `fuzz-chat` ever moves to `-c release` the cache assumption changes — worth a brief code comment near the `swift run` invocation.

### Suggested follow-ups (non-blocking)
- Add a one-line guard in `build-modes.yml` ensuring the nightly schedule has fired at least once in the last 14 days, alerting if not (catches dropped cron without manual log-spelunking).
- Consider a `if: github.event_name == 'pull_request'` skip for the `Upload audit artifacts` step on the 2 PR legs — those audit artifacts have no consumer on PRs and just bloat the actions storage. Procurement-evidence path is the nightly schedule.

### Verdict

**approve** — the matrix-conditional and cache-key alignment are both correct, CI proves the new behavior, no signal regressions identified. Concerns are minor / nightly-cron monitoring rather than blockers.

PR #892, opened 2026-04-29
ci: prune zombie schedules, surface Swift Testing failures, tighten timeouts

roryford:
## Summary

Four small CI hygiene fixes bundled together — all YAML-only.

- **Disable zombie scheduled triggers on `fuzz-nightly.yml` and `fuzz-weekly.yml`.** Both require a self-hosted `[self-hosted, macos, arm64]` runner that does not currently exist. Daily/weekly scheduled runs queue for 24h and auto-cancel — every day. Removed the `schedule:` trigger; kept `workflow_dispatch:` so a maintainer can fire them manually once a runner is registered.
- **`if: always()` on the Swift Testing step in `ci.yml`.** Currently the second `swift test` step is skipped when the first XCTest step fails, so a PR with broken tests in both harnesses requires two re-push cycles to triage. With `if: always()` both failure sets surface in one run; the job still fails because step 1 failed.
- **Drop `timeout-minutes` from 25 → 12 on the two test steps.** Recent green runs finish in 4-5 min; 12 min is 3× headroom but kills hung runs ~2× faster.
- **Replace `setup-go + go install actionlint` with the upstream prebuilt binary downloader.** Saves ~20s of cold Go setup per actionlint run.

## Test plan

- [ ] CI runs green on this PR.
- [ ] Manually verify the actionlint workflow still detects an intentionally bad workflow change (sanity check the new install path actually executes the linter).
- [ ] Confirm fuzz-nightly / fuzz-weekly no longer auto-trigger by checking `gh run list --workflow fuzz-nightly.yml` 24h after merge.

### DIFF
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -28,7 +28,9 @@ jobs:
 
       - name: Install actionlint
         id: install_actionlint
-        run: bash <(curl -sSfL https://raw.githubusercontent.com/rhysd/actionlint/main/scripts/download-actionlint.bash) 1.7.7
+        # Pin the bootstrap script to the same release tag as the binary so a
+        # compromised `main` on rhysd/actionlint cannot swap the download URL.
+        run: bash <(curl -sSfL https://raw.githubusercontent.com/rhysd/actionlint/v1.7.7/scripts/download-actionlint.bash) 1.7.7
 
       - name: Lint GitHub Actions workflows
         run: ${{ steps.install_actionlint.outputs.executable }} -color
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -26,16 +26,9 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v6
 
-      - name: Set up Go
-        uses: actions/setup-go@v6
-        with:
-          go-version: "stable"
-
       - name: Install actionlint
-        run: go install github.com/rhysd/actionlint/cmd/actionlint@v1.7.7
+        id: install_actionlint
+        run: bash <(curl -sSfL https://raw.githubusercontent.com/rhysd/actionlint/main/scripts/download-actionlint.bash) 1.7.7
 
       - name: Lint GitHub Actions workflows
-        run: |
-          set -euo pipefail
-          ACTIONLINT_BIN="$(go env GOPATH)/bin/actionlint"
-          "$ACTIONLINT_BIN" -color
+        run: ${{ steps.install_actionlint.outputs.executable }} -color
--- a/.github/workflows/build-modes.yml
+++ b/.github/workflows/build-modes.yml
@@ -24,7 +24,9 @@ name: Build modes + binary symbol audit
 
 on:
   schedule:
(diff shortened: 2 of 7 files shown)


roryford:
## Independent review

### Strengths
- Schedule blocks are **fully removed** (not commented out) on both fuzz workflows — clean revert via `git revert` once a runner exists. `workflow_dispatch:` retained so manual smoke is still possible.
- Disabled-schedule rationale comment is specific (names the runner label, points at FUZZING.md § CI tiers) — a future maintainer won't have to spelunk for context.
- `if: always()` on the Swift Testing step is the right call; verified that `Stage test diagnostics` and `Upload test diagnostics` both keep `if: failure()`, so the job conclusion still flips on step 10 failure and there's no double-upload.
- Verified `download-actionlint.bash` actually emits `executable` to `$GITHUB_OUTPUT` (script line 136), so `${{ steps.install_actionlint.outputs.executable }}` resolves correctly. The PR's actionlint run finished in **8s** — a real win vs. the prior `setup-go + go install` cold path.
- Step-duration sweep across 10 recent green runs: step 10 (XCTest) max ~4m54s, step 11 (Swift Testing) max ~16s. The new 12-min per-step cap leaves ~2.4x headroom on the slow step and ~45x on the fast one — safe.

### Concerns / questions
- **Supply-chain nit on `actionlint.yml:31`** — the bootstrap script is fetched from `rhysd/actionlint@main`, not a tagged ref. The downloader then pins the binary to `1.7.7`, but a compromised `main` could ship a poisoned downloader before the binary download even happens. Consider pinning the bootstrap URL to `/raw/v1.7.7/scripts/download-actionlint.bash` (or a commit SHA) so the whole chain is immutable. Low risk in practice, trivial to harden.
- **Stale comments in sibling workflows** — `nightly-slow-tests.yml:14` says "before fuzz-nightly at 05:00 UTC" and `build-modes.yml:27` says "after nightly-slow-tests (04:30) and fuzz-nightly (05:00)". Neither breaks anything, but with fuzz-nightly's schedule removed, these scheduling-rationale comments are now misleading. Worth a one-line clarifier ("fuzz-nightly schedule currently disabled — see workflow header").
- **FUZZING.md:365 / 374** still read "every day at 05:00 UTC" / "every Sunday at 07:00 UTC" with no callout that the schedule is currently off. Line 389 partially covers this, but the per-tier headings would benefit from a "(disabled until self-hosted runner exists)" inline note so docs match reality.
- **Diagnostics interaction with `if: always()`** — `Stage test diagnostics` comment notes "Each test step overwrites this file." If steps 10 *and* 11 both fail, the staged `test_output.txt` will be step 11's only; the XCTest log is clobbered. Probably fine (two harnesses = two stack traces in the run log itself), but worth knowing the artifact only carries the *last* failing step's captured stdout.

### Suggested follow-ups (non-blocking)
- Pin actionlint downloader URL to a tag/SHA.
- Sweep `nightly-slow-tests.yml`, `build-modes.yml`, and FUZZING.md for stale "fuzz-nightly at 05:00" copy in a tiny follow-up.
- If the dual-harness diagnostics clobber ever bites, switch staging to per-step subdirs.

Verdict: **approve with nits** — ship it; the supply-chain pin and doc sweep can land separately.

PR #893, opened 2026-04-29
refactor(inference): add nonisolated InferenceService wrappers (phase 1.2.4)

roryford:
## Summary

Adds `capabilitiesAsync`, `tokenizerAsync`, `enqueueAsync`, and `cancelAsync` as `nonisolated` async wrappers on `InferenceService`. Hops to the main actor internally — existing `@MainActor` synchronous accessors are unchanged.

This is **phase 1.2 sub-step 4** of the runtime ports refactor, per `docs/plans/runtime-ports-phase-1-2.md` (lines 380–386).

## Why

Sub-step 5 (`ConversationRuntime` extraction) needs to compose `InferenceService` from off-main contexts without wrapping every call site in `await MainActor.run { ... }`. This PR is internal API widening only — no user-facing behaviour change.

## Scope

- 100 LOC of production source (one new file: `Sources/BaseChatInference/Services/InferenceService+Nonisolated.swift`).
- ~310 LOC of tests covering off-main usage, concurrent callers, and coexistence with the existing `@MainActor` surface.
- No changes to existing `InferenceService` source. The wrappers are additive.

Returned types (`BackendCapabilities`, `TokenizerProvider`, `GenerationRequestToken`, `GenerationStream`) are all `Sendable`, so crossing the actor boundary is safe.

## Test plan

- [x] All CI-safe suites pass locally (`BaseChatCoreTests`, `BaseChatInferenceTests`, `BaseChatInferenceSwiftTestingTests`, `BaseChatUITests`, `BaseChatUIModelManagementTests`, `BaseChatMCPTests`, `BaseChatBackendsTests`, `BaseChatTestSupportTests`, `BaseChatAppIntentsTests`).
- [x] 8 new tests in `InferenceServiceNonisolatedTests` all pass.
- [x] Sabotage-checked: temporarily stripped `nonisolated` from each wrapper → tests fail to compile, confirming the off-main contract is enforced at compile time.

### DIFF
--- a/Sources/BaseChatInference/Services/InferenceService+Nonisolated.swift
+++ b/Sources/BaseChatInference/Services/InferenceService+Nonisolated.swift
@@ -1,0 +1,100 @@
+import Foundation
+
+// MARK: - Nonisolated wrappers for off-main runtime composition
+//
+// `InferenceService` is `@MainActor`-isolated for SwiftUI view binding.
+// Phase 1.2 sub-step 5 (`ConversationRuntime`) needs to compose the same
+// service from off-main contexts without forcing every call site to wrap
+// the call in `await MainActor.run { ... }`.
+//
+// These wrappers expose `nonisolated` async entry points that hop to the
+// main actor internally. The underlying coordinators stay `@MainActor` —
+// only the boundary changes. Returned types (`BackendCapabilities`,
+// `TokenizerProvider`, `GenerationRequestToken`, `GenerationStream`) are
+// all already `Sendable`, so crossing the actor boundary is safe.
+//
+// This is additive. The existing `@MainActor` accessors and methods on
+// `InferenceService` are unchanged; view code keeps working with the
+// synchronous main-actor surface.
+
+extension InferenceService {
+
+    // MARK: Capabilities
+
+    /// Off-main read of ``capabilities``.
+    ///
+    /// Use from non-`@MainActor` contexts (runtime use cases, background
+    /// tasks). View code should keep using the synchronous ``capabilities``
+    /// property — the values are identical.
+    public nonisolated func capabilitiesAsync() async -> BackendCapabilities? {
+        await MainActor.run { self.capabilities }
+    }
+
+    // MARK: Tokenizer
+
+    /// Off-main read of ``tokenizer``.
+    ///
+    /// The returned ``TokenizerProvider`` is `Sendable`; callers may use it
+    /// from any isolation domain after the call returns.
+    public nonisolated func tokenizerAsync() async -> (any TokenizerProvider)? {
+        await MainActor.run { self.tokenizer }
+    }
+
+    // MARK: Enqueue
+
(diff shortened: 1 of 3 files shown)


roryford:
## Review + fix

### Failure
CI's `test` job failed on a single test in BaseChatMCPTests:

`AuthMCPOAuthAuthorizationTests.test_tofu_emitsAuthorizationRequiredEvent_beforeBrowserOpens` — XCTAssertTrue failed: "authorizationRequired must be emitted before the browser opens" (line 1188).

The failure is unrelated to the InferenceService nonisolated wrappers introduced by this PR. The test was checking ordering between two AsyncStreams via:

1. A consumer `Task` that drained the event stream and bumped `tracker.authRequiredSeenCount`.
2. An unstructured `Task { await tracker.recordBrowserOpen(seen: await tracker.authRequiredSeenCount) }` launched from inside the redirect-listener handler.

Both are scheduling-dependent. Under the heavier concurrent load that this PR's wrappers introduce in the same test process, the listener-side Task could observe `seenCount == 0` before the consumer Task had drained the buffered event — even though the event was correctly yielded before `authorize()` was called. The test passes locally (verified) and presumably passed on main; it's a latent flake that the new wrappers exposed.

### Fix
Commit 7249d80 (Tests/BaseChatMCPTests/AuthMCPOAuthAuthorizationTests.swift):

Replaced the counter-read race with a `CheckedContinuation` handshake. The consumer task signals as soon as it observes `.authorizationRequired`; the listener handler awaits that signal before returning the callback URL. If production code regresses and yields the event *after* invoking `authorize()`, the handler blocks forever and the test times out — strictly stronger than the previous check.

Added an `asyncHandler:` init to the file-private `RedirectListenerMock` so the handler can perform awaits. The existing sync `init(handler:)` is preserved for the other tests in the file.

No production code changed. No `XCTSkip`/`withKnownIssue`. Sabotage check: by construction, swapping the order of `eventContinuation?.yield(...)` and `redirectListener.authorize(...)` in `MCPOAuth.swift` would deadlock the listener handler (the consumer can't observe the event until the handler returns, and the handler waits for the event). I did not run that mutation against the test process to avoid CLAUDE.md's SIGKILL prohibition, but the deadlock is mechanical.

Pre-push: full CI-safe suite ran green locally (BaseChatCore + Inference + InferenceSwiftTesting + UI + UIModelManagement + MCP + Backends + TestSupport + AppIntents).

### Other observations (non-blocking)
- The PR's wrappers themselves (`capabilitiesAsync`, `tokenizerAsync`, `enqueueAsync`, `cancelAsync`) look clean — thin `MainActor.run { ... }` hops over `Sendable`-returning surface, no extra allocation, no dropped error paths. The `enqueueAsync` signature deliberately drops the tuple-of-strings convenience overload — comment in the source explains why (runtime callers always have `[StructuredMessage]` already), which matches my read of where this is heading in the runtime ports work.
- One small thing the test suite doesn't cover: a `cancelAsync` call against an *unknown* token (already-cancelled or never-issued). The synchronous `cancel(_:)` is presumably idempotent; worth a one-line test to lock that contract for off-main callers too. Not blocking.
- The `makeLoadedService` helper comment honestly documents why `init(backend:)` doesn't survive off-main (`weak var provider` going nil). That's worth reading carefully — it suggests the `#if DEBUG init(backend:)` may not be a good off-main composition entry-point even after this PR lands. Probably belongs in the runtime-ports follow-up, not here.

PR #895, opened 2026-04-29
ci: parallelize BaseChatInferenceTests CI step

### LINKED ISSUE #841: ci: enable --parallel on BaseChatInferenceTests step
BaseChatInferenceTests runs 1016 XCTest cases serially (summed ~12.7s on a single worker). SwiftPM XCTest does not parallelize by default — passing `--parallel` should cut this step from ~23s to ~10s on 3-core macos-15 runners.

**Change:** in `.github/workflows/ci.yml`, add `--parallel` to the `Test — Inference (XCTest)` step invocation.

**Risks before flipping:**
- MockURLProtocol global state (per CLAUDE.md + memory, never share across concurrent suites — BaseChatInferenceTests should not touch it; verify with `grep -r MockURLProtocol Tests/BaseChatInferenceTests/`).
- SwiftData in-memory stores should already be per-test; sabotage-verify by breaking one parallelizable test and confirming it fails under `--parallel`.

**Expected saving:** ~10–13s per CI run.

_Filed as a follow-up from the PR #671 CI audit (2026-04-27)._

roryford:
## Summary
- split `BaseChatInferenceTests` out of the mixed XCTest CI batch into its own dedicated step
- run the new inference XCTest step with `--parallel`
- keep the broader Core/UI/etc. XCTest batch serial and document why it must stay that way

## Why
Issue #841 identified `BaseChatInferenceTests` as a good candidate for CI parallelism, but the combined XCTest step could not safely opt into `--parallel` because of the existing `UserDefaults` race in other suites. This change isolates the inference target so CI can take the speedup without reintroducing that flake.

## Validation
- relevant validation had already been run in the worktree before PR creation, per the provided task context

Closes #841


roryford:
Reviewed and merged main into the branch to clear the conflict. Recap of what I checked + changed:

## Verdict: approach is sound

The core design — splitting `BaseChatInferenceTests` into its own step with `--parallel`, keeping the mixed Core/UI/etc. batch serial — is the right way to honor the `--parallel` prohibition in `ci.yml`. The doc-comment that warned against `--parallel` was about the *combined* batch (which still touches `UserDefaults.standard` via `ChatViewModelTests.test_autoSelectFirstRunModel_*` in the UI suite). `BaseChatInferenceTests` does not have that problem:

- Every `UserDefaults`-touching test in the inference target uses a per-instance `UserDefaults(suiteName:)` (UUID-based) — see `DownloadManagerTests`, `ResumableDownloadTests`, `BackgroundDownloadIntegrationTests`, `DownloadTempCleanupTests`.
- `HuggingFaceServiceTests` scopes `MockURLProtocol` stubs to per-test unique hosts and uses targeted `unstub(url:)` instead of the global `reset()`, which matches the project's documented `MockURLProtocol` isolation pattern.

Sabotage-verified locally that `swift test --parallel` still surfaces real test failures end-to-end (broke `test_initialPhase_isIdle`, confirmed `scripts/test.sh` exits 1 with `RESULT: FAILED`, then reverted). One cosmetic note: under `--parallel`, swift test does not emit per-line `Test Case '...' passed (N seconds)` lines, only `[N/M] Testing ...` progress lines. So `scripts/test.sh`'s honest-summary will show `XCTest Passed: 0` for this step in CI even on a green run. Failures are still detected because swift test exits non-zero. Not blocking, but worth knowing — could be improved by teaching the parser to also count the progress-line format.

## Conflict resolution

Merged `origin/main` into the branch. The conflict was in `.github/workflows/ci.yml` against PRs #891/#892, which had landed:
- `timeout-minutes: 25` -> kept main's `timeout-minutes: 12`
- New `if: always()` on the Swift Testing step -> applied the same to the new `Test — Inference (XCTest, parallel)` step for consistent failure surfacing

Merge commit: f06219b. CI is now running on the merged branch.

PR #896, opened 2026-04-29
feat: trait-gate HuggingFace support and add AnyLanguageModel bridge

roryford:
Closes #280

## Summary
- move the HuggingFace search/download/validation surface into a dedicated `BaseChatHuggingFace` target behind the default-on `HuggingFace` trait, leaving protocol hooks in `BaseChatInference` and wiring the UI/examples to compile with or without it
- add the opt-in `BaseChatAnyLanguageModelBridge` target behind the `AnyLanguageModel` trait with a thin `InferenceBackend` adapter and dedicated tests
- update package/docs/runtime capability wiring so consumers can see and configure the new trait-gated surfaces

## Why
This trims the build graph for consumers that do not need the stock HuggingFace downloader while preserving backwards-compatible defaults, and it adds an opt-in bridge for AnyLanguageModel-backed providers without folding those integrations into BaseChatKit proper.

## Validation
- `swift test --quiet --filter BaseChatInferenceTests --disable-default-traits`
- `swift test --quiet --filter BaseChatUIModelManagementTests --disable-default-traits`
- `swift test --quiet --filter BaseChatHuggingFaceTests --disable-default-traits --traits HuggingFace`
- `swift test --quiet --filter BaseChatAnyLanguageModelBridgeTests --disable-default-traits --traits AnyLanguageModel`

PR #897, opened 2026-04-29
refactor: add PromptContextProvider port + passive merge pipeline (phase 1.2.2)

roryford:
## Summary

Phase 1.2 sub-step 2 of the runtime ports refactor (per #889 / #894).

- Introduces `PromptContextProvider` as a new **public** protocol in `BaseChatInference` — `Sendable`, no `@MainActor` pin, bare `messageCount: Int` argument; throws propagate.
- Adds `PromptContextPipeline` in `BaseChatCore` — a passive-merge use case that asks each registered provider for slots, concatenates, sorts by `PromptSlotPosition.sortIndex(messageCount:)`, and returns the assembled `[PromptSlot]`.
- No new aggregate value type ships from BCK — `[PromptSlot]` is the boundary primitive. Consumers that want richer accounting (Fireside's `ContextContribution` with `realCost`) keep that aggregate in their own module.
- Plan-doc correction: 9 references in `docs/plans/runtime-ports-phase-1-2.md` updated from `GenerationContextProvider` (re-exported) to `PromptContextProvider` (new).

## Why a new protocol, not a re-export

The plan's earlier framing — "re-export the existing `GenerationContextProvider`" — was based on a misread. The existing `GenerationContextProvider` in `Sources/BaseChatInference/Protocols/GenerationContextProvider.swift` is a `@MainActor`-bound, `AnyObject`-constrained **internal** model-state provider whose sole conformer is `InferenceService` (it surfaces `currentBackend` / `isBackendLoaded` / `selectedPromptTemplate` to `GenerationCoordinator`). It has nothing to do with slot contribution.

The slot-contributor port simply didn't exist. This PR introduces it under the `PromptContextProvider` name to pair naturally with `PromptContextPipeline` and avoid the collision. The two protocols have different shapes, different consumers, and live alongside. The plan doc now calls this out explicitly in the Ports table.

## Touch points outside the new files

- `PromptSlotPosition.sortIndex(messageCount:)` widened from internal to **public** so `PromptContextPipeline` (in `BaseChatCore`) can reuse the existing comparator. No call-site changes required — same comparator.
- `docs/plans/runtime-ports-phase-1-2.md` rewritten in 9 locations (Stance, Background, principle #6, Ports table, Use cases table, sub-step description, acceptance criterion, Fireside appendix). Two remaining `GenerationContextProvider` mentions are explicit clarifications that the internal protocol stays put.

## Test plan

- [x] `BaseChatCoreTests` — 9 new `PromptContextPipelineTests` cases pass:
  - Single provider returns its slots, sorted by position.
  - Multiple providers concatenate and sort; provider order does not affect final order when positions disagree.
  - Empty provider returns the union of others.
  - Empty pipeline returns empty.
  - Provider throws → pipeline propagates the exact error and does not return a partial result.
  - Position tie-breaking is stable in concatenation order (locks the actual `Array.sorted(by:)` behaviour).
  - `messageCount` flows verbatim to every provider.
  - `messageCount` drives `atDepth(_:)` sort indices (mc=10, n=1 sorts after n=9).
- [x] Per-test sabotage checks (verified locally, reverted before commit):
  - Skip the sort → 3 tests fail.
  - Skip provider iteration → 8 tests fail (only `test_emptyPipeline_returnsEmpty` correctly still passes since `[]` is correct there).
  - Hardcode `messageCount=0` → `test_messageCount_flowsToProviders` fails.
  - Swallow provider errors with `try?` → `test_providerThrows_pipelinePropagatesAndDoesNotReturnPartial` fails.
  - Reverse provider iteration → `test_positionTieBreak_isStableInProviderOrder` fails.
- [x] Full CI-safe pre-push suite green: `BaseChatCoreTests`, `BaseChatInferenceTests`, `BaseChatInferenceSwiftTestingTests`, `BaseChatUITests`, `BaseChatUIModelManagementTests`, `BaseChatMCPTests`, `BaseChatBackendsTests`, `BaseChatTestSupportTests`, `BaseChatAppIntentsTests`.

## Out of scope

- Wiring `PromptContextPipeline` into `BaseChatRuntime` — sub-step 5 (`ConversationRuntime` extraction) does that.
- Conforming `InferenceService` or any existing type to the new protocol — providers will be wired in by `ConversationRuntime` consumers (demo, ChatbotUI-iOS) and by Fireside in the same window.
- Touching the existing internal `GenerationContextProvider` — out of scope by explicit decision.

### DIFF
--- a/Sources/BaseChatCore/Services/PromptContextPipeline.swift
+++ b/Sources/BaseChatCore/Services/PromptContextPipeline.swift
@@ -1,0 +1,75 @@
+import Foundation
+import BaseChatInference
+
+/// Composes a list of ``PromptContextProvider``s into a single sorted
+/// ``[PromptSlot]`` for prompt assembly.
+///
+/// The pipeline is a passive merge: it asks each registered provider for its
+/// slots, concatenates the results, and sorts by
+/// ``PromptSlotPosition/sortIndex(messageCount:)``. Provider errors propagate
+/// out of ``assemble(messageCount:)`` and abort assembly — the pipeline does
+/// not return a partial result.
+///
+/// Providers are queried sequentially in registration order. Concurrent fan-out
+/// would invent ordering risk for ties (positions that share a sort index rely
+/// on the input array's order to remain stable across runs); the sequential
+/// hop is cheap because providers are typically I/O-light.
+///
+/// ## Topology
+///
+/// `PromptContextPipeline` is the slot-merge layer. It does not perform budget
+/// allocation, token counting, or message trimming — that work belongs to
+/// `PromptAssembler` (in `BaseChatInference`). Consumers that want a richer
+/// aggregate (Fireside's `ContextContribution` with `realCost` accounting) keep
+/// that aggregate in their own module. BCK ships the boundary primitive here.
+///
+/// ## Example
+///
+/// ```swift
+/// let pipeline = PromptContextPipeline(providers: [
+///     systemPromptProvider,
+///     loreProvider,
+///     retrievalProvider,
+/// ])
+/// let slots = try await pipeline.assemble(messageCount: history.count)
+/// ```
+///
+/// > Note: Phase 1.2 sub-step 2 ships the use case in isolation; sub-step 5
+/// > wires it into `ConversationRuntime`.
+public final class PromptContextPipeline: Sendable {
+    private let providers: [any PromptContextProvider]
+
+    /// Creates a pipeline that queries `providers` in registration order.
+    ///
+    /// - Parameter providers: The providers to compose. An empty array is a
+    ///   valid pipeline that always returns `[]`.
+    public init(providers: [any PromptContextProvider]) {
+        self.providers = providers
+    }
+
+    /// Asks each registered provider for its slots, concatenates the results,
+    /// and returns them sorted by
+    /// ``PromptSlotPosition/sortIndex(messageCount:)``.
+    ///
+    /// - Parameter messageCount: The conversation message count at assembly
+    ///   time. Forwarded verbatim to every provider so that
+    ///   ``PromptSlotPosition/atDepth(_:)`` can compute its sort index against
+    ///   the same baseline the consumer is about to render against.
+    /// - Returns: All contributed slots, sorted top-to-bottom (lowest sort
+    ///   index first). Slots whose positions tie keep their concatenation
+    ///   order — this is whatever `Array.sorted(by:)` gives, which is stable
+    ///   on the standard library implementation BCK targets.
+    /// - Throws: Whatever a provider throws. The pipeline aborts immediately
+    ///   on the first failure and does not return a partial result.
+    public func assemble(messageCount: Int) async throws -> [PromptSlot] {
+        var merged: [PromptSlot] = []
+        for provider in providers {
+            let contribution = try await provider.contributeSlots(messageCount: messageCount)
+            merged.append(contentsOf: contribution)
+        }
+        return merged.sorted { lhs, rhs in
+            lhs.position.sortIndex(messageCount: messageCount) <
+                rhs.position.sortIndex(messageCount: messageCount)
+        }
+    }
+}
--- a/Sources/BaseChatInference/Protocols/PromptContextProvider.swift
+++ b/Sources/BaseChatInference/Protocols/PromptContextProvider.swift
@@ -1,0 +1,27 @@
+import Foundation
+
+/// A source of prompt context slots for assembly into a generation request.
+///
+/// Conformers contribute slots based on conversation state; ``PromptContextPipeline``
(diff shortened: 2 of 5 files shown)


roryford:
## Review — Outcome A (sound, no rebase needed)

Reviewed the full PR diff (341/-25), read each touched file in full, and verified the plan-doc refactor is internally consistent. All CI checks green; behind-main status assessed and the PR can land safely without a rebase (see below).

### What I checked

- **Module placement.** `PromptContextProvider` lives in `BaseChatInference` (no ML / no SwiftData — only `import Foundation`); `PromptContextPipeline` lives in `BaseChatCore` and explicitly `import BaseChatInference` rather than relying on a re-export, per the CLAUDE.md "BaseChatCore does not re-export BaseChatInference" rule.
- **Concurrency posture.** Protocol is bare `Sendable` (no `@MainActor` pin, no `AnyObject` constraint), `async throws`. Pipeline is `final class … Sendable` with only `let` storage. The deliberate sequential iteration (no `withTaskGroup`) is documented and correct — concurrent fan-out would invent ordering risk for ties.
- **Naming-collision hazard.** The existing `GenerationContextProvider` (`Sources/BaseChatInference/Protocols/GenerationContextProvider.swift`) is `@MainActor`-bound, `AnyObject`-constrained, and **internal**, with `InferenceService` as its sole conformer. Surfaces `currentBackend`/`isBackendLoaded`/`selectedPromptTemplate` to `GenerationCoordinator`. The new `PromptContextProvider` has a different shape, different consumer set, and different visibility — they should not be merged. The PR docstring and the rewritten Ports table both call this out.
- **`PromptSlotPosition.sortIndex` widening.** Purely additive — was `internal`, now `public`. No callers broken; the docstring explains why (`PromptContextPipeline` in `BaseChatCore` needs the comparator).
- **No `assertionFailure` / `fatalError`** in any new file. Errors are surfaced via `throws` to the pipeline boundary.
- **Test coverage.** 9 cases in `PromptContextPipelineTests` cover sort order under a single provider, multi-provider concatenation, order independence under disagreeing positions, empty providers, the empty pipeline, error propagation (no partial result), tie-break stability in input order, `messageCount` plumbing to every provider, and the `messageCount`-driven `atDepth(_:)` sort math. Two distinct conformers exercised (`FakeProvider` and `ThrowingProvider`) — not just a single default. Sabotage notes documented per case in the file header. The `SendableBox<T>` capture cell is the right pattern for collecting call args out of a `Sendable` value-type provider in single-threaded test code.
- **Public-API stability.** Only the new types (`PromptContextProvider`, `PromptContextPipeline`) and the widened `sortIndex` are added to the public surface. Sub-step 5 (`ConversationRuntime` extraction) will wire the pipeline; sub-step 2 ships the use case in isolation, as designed.
- **Plan doc.** All 9 corrections from `GenerationContextProvider` (re-exported) → `PromptContextProvider` (new) are internally consistent. The two remaining mentions of `GenerationContextProvider` are explicit clarifications that the internal protocol stays put.

### BEHIND-main assessment

Two commits on `main` ahead of this branch:

- **#895** — `ci: parallelize BaseChatInferenceTests CI step` (workflow YAML only).
- **#896** — `feat: trait-gate HuggingFace and add AnyLanguageModel bridge` — touches `Sources/BaseChatHuggingFace/`, `Sources/BaseChatAnyLanguageModelBridge/`, `Package.swift`, plus HF/bridge tests. **No file overlap** with this PR (`PromptContextProvider.swift`, `PromptContextPipeline.swift`, `PromptSlot.swift`, the new test file, and the plan doc).

Test-merge against `origin/main` reported `Automatic merge went well; stopped before committing` — cleanly aborted. No behavioural overlap with #890 (port extractions), #893 (nonisolated wrappers), or #896 (trait gating). The PR can merge as-is and Release Please will pick up the next `feat`/`fix` correctly.

### Non-blocking observations

- **Trailing-comment re `Equatable`.** The bottom-of-file note about `PromptSlot` not being `Equatable` is helpful for future maintainers; if the type ever conforms, the tests above can drop the `.map(\.id)` and compare value-equality directly. No action needed now.
- **Tie-break docstring.** `assemble(messageCount:)` documents reliance on `Array.sorted(by:)` stability ("stable on the standard library implementation BCK targets"). This is true on Swift 5+ and remains true on Swift 6.x; matches the Phase 1.2 acceptance window. No action needed.
- **Plan-doc table line 291** (`PromptContextPipeline` use-case row) marks this PR as `✅ shipped (#TBD …)`. After merge, the `#TBD` will want to be `#897` — a one-character follow-up that can ride a later docs touch in this phase.

### Constraints honoured

- No `git push` — branch unchanged.
- No `gh pr review --approve`, no `--auto`, no merge.
- No backwards-compat shims, no dead `_` renames.

🤖 Reviewed by Claude Code

PR #898, opened 2026-04-29
refactor: split ChatPersistenceProvider into SessionStore + MessageStore (phase 1.2.1)

roryford:
## Summary

Phase 1.2.1 of the runtime ports refactor — see [`docs/plans/runtime-ports-phase-1-2.md`](docs/plans/runtime-ports-phase-1-2.md). Splits the combined `ChatPersistenceProvider` into two per-port protocols (`SessionStore` and `MessageStore`) and ships `MessageStorePostWriteHook` / `SessionStorePostWriteHook` as low-level primitives alongside.

The combined provider was a v0.x convenience; per-port boundaries make selective implementation by hosts cheaper and prepare the persistence surface for the Phase 2 physical target split.

## Breaking changes

Pre-1.0 hard cut, no deprecation tail.

- `ChatPersistenceProvider` is removed. Replace with a type that conforms to either or both of `SessionStore` / `MessageStore`. The shipped `SwiftDataPersistenceProvider` now conforms to both.
- `ChatPersistenceError` and the `makeMessageSearchSnippet` helper move from `BaseChatCore` to `BaseChatInference` alongside the records they describe. Files using these types must `import BaseChatInference` directly (per CLAUDE.md, `BaseChatCore` does not re-export `BaseChatInference`).
- `ChatViewModel.configure(persistence:)` and `SessionManagerViewModel.configure(persistence:autoLoad:)` now accept `any SessionStore & MessageStore`. Hosts wiring a single SwiftData adapter need no call-site change beyond the import.

## Hook semantics

`MessageStorePostWriteHook` is a **low-level primitive**, not the canonical attachment point for any specific consumer's cross-cutting concerns. Per the plan doc:

- Hooks fire **after** the underlying write commits, in registration order.
- Hooks **must not throw** — a failing hook cannot roll back the committed write. Errors are logged via `Log.persistence.error` and otherwise swallowed.
- The hook is for consumers whose unit of work *is* the message (audit, indexing, debug logging). Consumers whose unit of work is the turn compose at the turn-orchestrator layer instead.

`SessionStorePostWriteHook` ships as the symmetric counterpart but no internal consumer uses it yet — its shape is provisional until a host actually exercises it.

## Choices made

- **Single SwiftData impl conforming to both protocols** (option (a) in the spec). Sessions and messages live in the same `ModelContext`; splitting the adapter would only duplicate the context plumbing without buying a real boundary.
- **Combined-existential parameter** (`any SessionStore & MessageStore`) on `configure(persistence:)`. Hosts pass one SwiftData adapter; this minimises call-site churn across the ~50 test sites that wire `SwiftDataPersistenceProvider`.

## SessionListService note

`SessionListService` (extracted via #886) drives session-scope work but also surfaces the message-scope search the UI needs (`runMessageSearch`). It now takes the combined `any SessionStore & MessageStore` existential rather than only `SessionStore` — splitting message-search out into a separate use case would require fanning the user's single search command across two services and is out of scope for 1.2.1. A future PR (or the Phase 2 split) is the natural place to revisit.

## Test plan

- [x] Protocol-level hook tests (in-memory `MessageStore` fake, `BaseChatInferenceTests`) — registration order, post-commit visibility, no-replay-on-late-registration, default no-op for stores that don't override `addPostWriteHook`.
- [x] SwiftData hook integration tests (`BaseChatCoreTests`) — message + session hook firing against a real in-memory `ModelContainer`, registration ordering, delete-does-not-fire-post-write.
- [x] Sabotage-checked: in-memory store, SwiftData message hooks, SwiftData session hooks each break their respective tests when the hook firing is disabled.
- [x] Existing `ChatPersistenceProvider` test references migrated to the split (`PersistenceGuardTests`, `RuntimeBootstrapMigrationGuardTests`, `SessionListServiceTests` — all green).
- [x] Full pre-push checklist passed: `BaseChatCoreTests` (311), `BaseChatInferenceTests` (1346), `BaseChatInferenceSwiftTestingTests` (83), `BaseChatUITests` (470), `BaseChatUIModelManagementTests` (100), `BaseChatBackendsTests` (128), `BaseChatMCPTests` (130), `BaseChatTestSupportTests` (16), `BaseChatAppIntentsTests` (11) — zero failures.


roryford:
## Review — Outcome A (sound, no rebase needed)

Reviewed at `a3ad3e4`, base `9913edf`. All CI green. BEHIND-status assessed against current `origin/main` (`f0d7cd0`).

### Interleave-with-main check

Per CLAUDE.md ["Interleaved PR stacks can break main"](https://github.com/roryford/BaseChatKit/blob/main/CLAUDE.md), per-PR CI proves compatibility with the branch base, not post-merge main. The two commits on `main` since this branch's base are:

- **#895** — CI parallelization (`.github/workflows/ci.yml` only)
- **#896** — HuggingFace trait-gating + AnyLanguageModel bridge (HF/Bridge sources, `Package.swift`, `MockDownloadManager`, BG download manager protocol)

Files touched on `main` since divergence are **disjoint** from the persistence-port surface this PR rewires. `git merge --no-commit --no-ff origin/main` reports "Automatic merge went well." No rebase required for behavioral signal — pushing a refresh would burn ~25 billed CI minutes for zero diagnostic value.

The earlier-merged-out-of-order phases (#890 ports extraction, #893 nonisolated wrappers) similarly do not call into the `ChatPersistenceProvider` surface — `grep -rn "ChatPersistenceProvider" Sources/ Tests/ Example/` after this PR returns only doc-prose references in the four split files themselves, confirming the type is fully retired and no later phase silently grew a dependency on its old shape.

### Transactional invariants — preserved

The highest-risk question: did the protocol split lose any single-op atomicity that `ChatPersistenceProvider` enforced? Audit:

- **`deleteSession`** preserves the original ordering — `try await deleteMessages(for: sessionID)` runs first, then `modelContext.delete(session)`, then a single `try modelContext.save()`. Both rows commit through the same `ModelContext`, so the messages-then-session deletion remains transactionally consistent on the SwiftData adapter (the only shipped impl). The `SessionStore.deleteSession` doc explicitly calls this out: *"Implementations that also conform to `MessageStore` typically delete the session's messages as part of this call to keep the two stores transactionally consistent against the same backing context."*
- **Insert/update paths** never had cross-record atomicity to lose — sessions and messages were independently committed before the split.
- **Hook firing on `deleteSession`** is intentionally absent (hooks fire on insert/update only, symmetric with the message hook contract). Documented in `test_messageHook_doesNotFireOnDelete`.

The `SessionListService` keeps a combined `any SessionStore & MessageStore` reference (intentional, called out in the PR body and the source comment) — splitting `runMessageSearch` would force the adapter to fan one user search across two services. Reasonable Phase-2 deferral.

### Schema migration — not required

No `@Model` types are touched. `BaseChatSchemaV3` and `ModelContainerFactory.currentSchema` stay as-is. The split is purely at the protocol surface, not the persisted-data layer.

### Other invariants

- `BaseChatCore` → `BaseChatInference` move of `ChatPersistenceError` + `MessageStore`/`SessionStore` is appropriate: the records (`ChatSessionRecord`, `ChatMessageRecord`) already live in `BaseChatInference`, so the port protocols belong there too. `BaseChatCore` retains the SwiftData adapter only.
- No `assertionFailure` / `fatalError` in the new code paths — `grep -nE "assertionFailure|fatalError" Sources/BaseChatInference/Protocols/{MessageStore,SessionStore}.swift Sources/BaseChatCore/Services/SwiftDataPersistenceProvider.swift` returns nothing.
- Hooks are documented as low-level primitives, not the canonical attachment point — the doc-comment posture matches the plan-doc framing.
- Test coverage matches the surface: protocol-level tests against an in-memory `MessageStore` fake in `BaseChatInferenceTests`, integration coverage against the real SwiftData adapter in `BaseChatCoreTests`. Both files include sabotage-checked claims in the PR body.

### Non-blocking observations

1. **`SessionStorePostWriteHook` ships with no in-tree consumer.** PR body acknowledges this is provisional. Fine to ship as a symmetric pair, but worth flagging that the next host who picks it up may want to nudge the signature (e.g. add a `kind: insert | update` discriminator the way some other audit-hook surfaces do). Not blocking.
2. **In-memory `MinimalMessageStore` test asserts the protocol-extension default no-op for `addPostWriteHook` — but the same fake is `private`, so its existence as a guard against the default disappearing is local to one file.** If the no-op default ever gets removed deliberately, the regression catch lives only here. Not worth re-engineering for 1.2.1, but worth a note in the Phase-2 surface review.
3. **`ChatPersistenceError.providerNotConfigured` case name mentions "provider" but the type is now used by stores.** Pre-1.0, fine to leave; flagged only because a follow-up rename pass might want to revisit.

LGTM. Ship it.

PR #905, opened 2026-04-30
refactor: add branch sub-flow + wire ChatViewModel adapter to ConversationRuntime (phase 1.2.5 PR-D)

roryford:
## What ships

Phase 1.2.5 PR-D — final sub-flow and adapter wiring. Builds on PR-A (#899), PR-B (#902), and PR-C (#903).

### Branch sub-flow

- `BranchInput` — new `Sendable` input struct (mirrors `SendInput`/`RegenerateInput`/`EditInput`)
- `ConversationRuntime.branch(_:)` — forks a conversation at a chosen message: creates a new `ChatSessionRecord`, copies messages up to and including the branch point, emits `.sessionBranched`, and optionally triggers a generation turn on the new session if `generateAfterBranch == true` and the last copied message is `.user`
- `ConversationEvent.sessionBranched(newSessionID: UUID, copiedCount: Int)` — 15th case; fires synchronously before `branch` returns
- `ConversationError.messageNotFound` already existed (added in PR-C); reused as the branch-point-not-found error

### ChatViewModel adapter

- `configure(conversationRuntime:)` — new public method; cancels any prior drain task, stores the runtime, starts a long-lived `[weak self]` drain task over `runtime.events`
- `ChatViewModel+RuntimeAdapter.swift` (new) — `handle(runtimeEvent:)` maps all 15 `ConversationEvent` cases to `@Observable` state mutations: message insert/update/remove, phase transitions, token delta application, error surfacing
- `ChatViewModel+Messages.swift` — `sendMessage()`, `regenerateLastResponse()`, `editMessage(_:newContent:)`, and `stopGeneration()` each check `conversationRuntime` first; when set, they delegate to the runtime and return early; the existing `GenerationCoordinator` path runs unchanged when no runtime is configured
- `activeConversationStreamHandle: ConversationStreamHandle?` — stored on `ChatViewModel` for cancel routing

**The runtime path is opt-in.** Callers that don't call `configure(conversationRuntime:)` see no behaviour change. `GenerationCoordinator`'s token batching, thinking-block disclosure, loop detection, and tool dispatch remain in the existing path — those migrate in follow-up PRs.

## Side fix

`test_edit_assistantMessage_updatesAndReturnsNilHandle` had a pre-existing timestamp-ordering flake when two `ChatMessageRecord`s were created with the same `Date()`. Fixed by seeding explicit timestamps (same pattern used throughout the new branch tests).

## Test plan

- [x] `test_branch_copiesMessagesAndEmitsEvent` — happy path: 3-msg session, branch at index 1, no generation; asserts `.sessionBranched(copiedCount: 2)` and nil handle
- [x] `test_branch_triggersGenerationWhenLastCopiedMessageIsUser` — `generateAfterBranch: true`, last copied msg is user; asserts non-nil handle and generation events against `newSessionID`
- [x] `test_branch_noGenerationWhenLastCopiedMessageIsAssistant` — `generateAfterBranch: true` but last copied msg is assistant; asserts nil handle
- [x] `test_branch_messageNotFound_throws` — bogus `branchMessageID`; asserts `ConversationError.messageNotFound`
- [x] `test_branch_insertSessionFails_throwsBeforeAnyMessagesCopied` — poisoned `insertSession`; asserts `ConversationError.persistence` with zero messages copied
- [x] `test_sendMessage_withRuntime_delegatesToRuntime` — configured runtime, fires `sendMessage`, waits for generation cycle, asserts user+assistant in `messages`
- [x] `test_stopGeneration_withRuntime_cancelsHandle` — in-flight send, `stopGeneration()`, asserts clean termination
- [x] `test_sendMessage_withoutRuntime_usesExistingPath` — no runtime configured, asserts `GenerationCoordinator` path unchanged
- [x] All 9 CI-safe suites: 344 `BaseChatCoreTests` + 473 `BaseChatUITests` — 0 failures locally

## Plan refs

- Plan doc: `docs/plans/runtime-ports-phase-1-2.md`
- Builds on: #899 (PR-A), #902 (PR-B), #903 (PR-C)
- Phase 1.2 exit criterion: "no `@Observable` view model owns orchestration state" — met once `GenerationCoordinator`'s remaining behaviours (batching, thinking, tool dispatch) migrate into the runtime in follow-up PRs

### DIFF
--- a/Sources/BaseChatCore/Services/ConversationEvent.swift
+++ b/Sources/BaseChatCore/Services/ConversationEvent.swift
@@ -63,6 +63,12 @@ public enum ConversationEvent: Sendable {
     /// message in their view-state array.
     case messageUpdated(ChatMessageRecord)
 
+    /// A new session was created by branching from an existing conversation.
+    /// Fires synchronously (before `branch` returns) after the copied messages
+    /// are persisted. `copiedCount` is the number of messages inserted into
+    /// the new session.
+    case sessionBranched(newSessionID: UUID, copiedCount: Int)
+
     /// The runtime queued a generation request and the underlying stream
     /// started. Carries the assistant message ID the stream will write
     /// into so adapters can pair token deltas with the right message slot.
--- a/Sources/BaseChatCore/Services/ConversationRuntime.swift
+++ b/Sources/BaseChatCore/Services/ConversationRuntime.swift
@@ -535,14 +535,23 @@ public final class ConversationRuntime: Sendable {
         }
         let slice = Array(sourceHistory[...branchIndex])
 
-        // Create the new session. Derive title from the source session when
-        // the caller didn't supply one — best-effort via fetchSessions; no
-        // title is better than crashing if session store is absent.
+        // Derive the new session's title from the source session when the
+        // caller didn't supply one. A title-fetch failure must not abort the
+        // branch — the session insert below still runs with the fallback —
+        // but the failure is logged so it isn't silently lost.
         let newSessionTitle: String
         if let supplied = input.newSessionTitle {
             newSessionTitle = supplied
         } else if let sessionStore {
-            let sessions = (try? await sessionStore.fetchSessions()) ?? []
+            let sessions: [ChatSessionRecord]
+            do {
+                sessions = try await sessionStore.fetchSessions()
+            } catch {
+                Log.persistence.warning(
+                    "ConversationRuntime.branch: title lookup failed: \(error.localizedDescription); using fallback title"
+                )
+                sessions = []
+            }
             newSessionTitle = sessions.first(where: { $0.id == input.sourceSessionID })?.title ?? "New Chat"
         } else {
             newSessionTitle = "New Chat"
--- a/Sources/BaseChatCore/Services/ConversationRuntime.swift
+++ b/Sources/BaseChatCore/Services/ConversationRuntime.swift
@@ -125,6 +125,62 @@ public struct EditInput: Sendable {
     }
 }
 
+// MARK: - Branch input
+
+/// Input for ``ConversationRuntime/branch(_:)``.
+///
+/// Forks a conversation at a chosen message. The runtime copies messages from
+/// `sourceSessionID` up to and including `branchMessageID` into a new session
+/// identified by `newSessionID`, then optionally triggers a generation turn on
+/// the new session if the last copied message is a user message.
+public struct BranchInput: Sendable {
+    /// The session to fork from.
+    public let sourceSessionID: UUID
+    /// The message to branch at (inclusive — this message is copied into the
+    /// new session).
+    public let branchMessageID: UUID
+    /// The caller-supplied ID for the new session.
+    public let newSessionID: UUID
+    /// Title for the new session. `nil` preserves the source session's title.
+    public let newSessionTitle: String?
+    /// When `true` and the last copied message is `.user`, the runtime
+    /// triggers a generation turn on the new session after copying.
+    public let generateAfterBranch: Bool
+    // Generation knobs — only used when generateAfterBranch == true.
+    public let systemPrompt: String?
+    public let temperature: Float
+    public let topP: Float
+    public let repeatPenalty: Float
+    public let maxOutputTokens: Int?
+    public let maxThinkingTokens: Int?
+
+    public init(
+        sourceSessionID: UUID,
+        branchMessageID: UUID,
+        newSessionID: UUID = UUID(),
+        newSessionTitle: String? = nil,
+        generateAfterBranch: Bool = false,
+        systemPrompt: String? = nil,
+        temperature: Float = 0.7,
+        topP: Float = 0.9,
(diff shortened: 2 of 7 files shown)


roryford (commented):
Review pass — three issues found, all three fixed in commits on this branch (ce76619, 81e14c7, 1284efe). Pre-push CI suite green locally; one pre-existing flake noted below.

## Fixed

### 1. Retain cycle in `configure(conversationRuntime:)` — ce76619

The drain task's capture list was `[weak self]` plus `guard let self else { return }` *outside* the `for-await` loop. The `guard let` upgrades `self` to a strong local for the rest of the closure scope, which spans every suspension point inside the for-await. Combined with the view model strongly owning both the runtime (which keeps `events` open) and the task itself, this leaks the view model — `deinit` never runs because the task it owns retains it.

Moving `guard let self` *inside* the loop body keeps `self` weak across the suspended `for-await`, so the view model can deallocate when nothing else holds it; the next event drains the closure and the task exits. This is the same pattern documented in the project's `feedback_weak_self_detached_task.md` memory, applied symmetrically.

Side fix: dropped the unused `[weak self]` + `_ = self // suppress capture warning` workaround in `stopGeneration`'s runtime branch — the closure body never references `self`, so the capture list (and the suppression hack) were both unnecessary.

### 2. Regression test for the cycle — 81e14c7

Captures `weakVM` after `configure(conversationRuntime:)`, drops the harness, and asserts the VM deallocates. Sabotage check (verified during the review): hoisting `guard let self` back outside the `for-await` produces a 1-failure run; restoring the per-iteration guard returns to green. The test sleeps 30 ms after `configure` so the drain task actually starts and suspends — without that wait, the test is a no-op either way.

### 3. Silent error swallow in `branch(_:)` — 1284efe

`(try? await sessionStore.fetchSessions()) ?? []` on the title-lookup path tripped the `SilentCatchAuditTest`. Replaced with an explicit do/catch that falls back to an empty list and logs the underlying error through `Log.persistence.warning`, preserving the best-effort fallback without losing the diagnostic. Without this fix, `BaseChatInferenceTests` is red on CI.

## Verified

- Branch ordering: `.sessionBranched` fires synchronously after copies are inserted, before `runBranchGenerationTurn` is launched. The adapter handler's no-op for `.sessionBranched` is correct for now (no view-state mutation needed in the chat-only consumer path; full session UX migration is a follow-up).
- Pre-registration cancel race: `runBranchGenerationTurn` → `runGenerationTurn` reuses the same `isCancelled`-after-`register` race check that send/regenerate/edit use. No regression.
- Early-return discipline in `ChatViewModel+Messages.swift`: send/regenerate/edit/stop all check `conversationRuntime` before any `messages.append` or other GenerationCoordinator-owned state mutation. The "two-system cancel bug" the plan warns about is not present.
- `insertSession` rollback: when `insertSession` fails, the loop throws before any messages are copied. The test covers this.
- No `assertionFailure` / `fatalError` in recoverable paths.

## Flagged but not fixed (no follow-up issues per project rule)

- **Pre-existing flake: `test_regenerate_cancelMidStream_partialContentPersists`** — the `.streamFinished(reason: .cancelled)` assertion fails ~1 in 3 runs in a full suite invocation but always passes in isolation. The test uses two sequential `for-await` loops on the same single-consumer `runtime.events` stream; events that should arrive between the two consumer windows are sometimes consumed by the first loop's break-out trailing iterations or batched away. This was introduced in PR-B (`2e53b02`), not by this PR. Worth a separate PR to either (a) collect events into an array up front, or (b) assert on persisted store state alone.
- **Auto-title gap on the runtime path** — `sendMessage()` skips the `onFirstMessage?(session, text)` callback when delegating to the runtime. Title generation won't fire on the first user message in a runtime-configured session. This is the documented "follow-up PR" gap (the plan's "GenerationCoordinator features migrate later" framing covers it), so leaving as-is.
- **Partial-copy state in `branch(_:)` if mid-loop `insertMessage` fails** — `insertSession` succeeds, then on message N the loop throws. Messages 0..N-1 are persisted to the new session, but the new session has only a partial transcript. The current test suite only covers the all-or-nothing `insertSession` failure case. Acceptable for v1; the runtime treats every persistence failure as a hard throw and lets the caller reload from the store.
- **Comment hygiene** — a few new comments restate code rather than explain it (e.g., `// Find the branch point and slice history up to and including it.`, `// Copy messages into the new session with fresh IDs and updated sessionID.`). Not a blocker for an opt-in path.

## Pre-push CI suite

All 9 CI-safe filters green locally with the three fix commits applied:

```
BaseChatCoreTests              passed
BaseChatInferenceTests         passed (after 1284efe)
BaseChatInferenceSwiftTestingTests   passed (83 tests)
BaseChatUITests                passed
BaseChatUIModelManagementTests passed
BaseChatMCPTests               passed
BaseChatBackendsTests          passed
BaseChatTestSupportTests       passed
BaseChatAppIntentsTests        passed
```

Commits pushed: ce76619, 81e14c7, 1284efe. PR is otherwise ready to merge once you want to land the runtime adapter.

Issue #964, opened 2026-05-02
refactor: cyclomatic complexity reduction in inference + backends


roryford:
## Summary

A cyclomatic-complexity audit (lizard, 2,200 functions across `Sources/`) shows the codebase is healthy on average (CCN 3.2, only 4.3% of functions ≥ 15) but complexity is concentrated in three places where it has compounding cost. This issue tracks targeted refactors at those joints.

## Background

Top hotspots (CC ≥ 15): 94 functions, ~32 in `BaseChatBackends`, ~22 in `BaseChatInference`, ~10 in `BaseChatMCP`. After reading the worst offenders directly, only some of the high-CC functions reflect real accidental complexity — others (recursive JSON Schema walks, test fakes, CLI parsers, the post-#470 Llama driver) score high because their vocabulary is wide, not because they're tangled. This issue scopes only the refactors that carve at real joints.

## Tasks

- [ ] **`MLXBackend.generate` (CC 62) → extract setup helpers.** ~230 of its 280 lines are setup *before* the streaming loop: Qwen tool-block injection, message-dict construction, KV-cache-reuse decision, parser/marker resolution. Lift these into static helpers (or an `MLXGenerationDriver` mirroring the Llama pattern from #470) so the streaming function itself drops to ~CC 20. Highest single-function leverage on the audit.

- [ ] **Stream-parser cohort → shared `SSEEventRouter`.** Five functions sit in CC 19–25 doing variations on the same job: `OllamaBackend.handleLine`, `OpenAIBackend.parseResponseStream`, `ClaudeBackend.parseResponseStream`, `OpenAIResponsesBackend.handleEvent`, `SSECloudBackend` parsing. Factor SSE framing + event dispatch into a shared router; backends declare event shapes. Compounds across 4+ files instead of helping one.

- [ ] **`JSONSchemaValidator` trio → split by JSON type.** `validate` (CC 18), `valueMatchesType` (CC 22), `validateObject` (CC 19) are three adjacent high-CC functions in one file. Splitting per JSON type (string / number / array / object) carves cleanly along the spec.

## Out of scope (deliberate non-goals)

- `MCPToolSource.schemaDepthIfCompatible` (CC 54) — recursive JSON Schema vocabulary walk where each branch is structurally identical. High CC reflects schema-keyword breadth, not cognitive load. A table-driven rewrite would obscure the documentation value of the explicit cases.
[...]

PR #969, opened 2026-05-02
refactor(inference): split JSONSchemaValidator helpers per JSON type (#964)

roryford:
## Summary

One slice of the cyclomatic-complexity reduction in #964: splits the three high-CC functions in `JSONSchemaValidator` into per-JSON-type sub-helpers. Pure code motion — zero behavior change.

### CC before -> after (lizard)

| function | before | after |
|---|---|---|
| `validate` (recursive core) | 18 | **6** |
| `valueMatchesType` | 22 | **8** |
| `validateObject` | 19 | **5** |

After the refactor, `lizard` reports zero functions over its CC=15 warning threshold for this file.

### Extractions

- **`validate`** — pulled the inline `enum`-check block into `checkEnum`; pulled the per-kind switch into `validatePerKindKeywords`.
- **`valueMatchesType`** — replaced the dispatch-with-inline-case-matches with a one-line dispatch over per-type predicates (`isObject`, `isArray`, `isString`, `isBoolean`, `isNull`, `isNumber`, `isInteger`). Each predicate is CC 3-4.
- **`validateObject`** — split into `validateRequired`, `resolvePropertySchemas`, `validateAdditionalProperties`, and `validateProperties`. `resolvePropertySchemas` returns a small private `PropertySchemaResolution` enum so the failure path stays explicit.

No visibility was widened. All new helpers are `private`.

### Test plan

- [x] `JSONSchemaValidatorTests` (36 cases) all pass.
- [x] Full pre-push gate green: 1437 XCTest + 83 Swift Testing.
- [x] Sabotage check: replaced the missing-required-field return with `continue`; confirmed 5 `JSONSchemaValidatorTests` failed; reverted before commit.

Closes a slice of #964.

### DIFF
--- a/Sources/BaseChatInference/Services/JSONSchemaValidator.swift
+++ b/Sources/BaseChatInference/Services/JSONSchemaValidator.swift
@@ -112,61 +112,56 @@ public struct JSONSchemaValidator: Sendable {
             return failure
         }
 
-        // `enum` is evaluated first — it short-circuits on both primitives and
-        // composite values and produces the clearest model-facing message.
-        if let enumValues = schemaObject["enum"], case let .array(allowed) = enumValues {
-            if !allowed.contains(value) {
-                return ValidationFailure(
-                    modelReadableMessage: "argument \(Self.describePath(path)) must be one of: "
-                        + Self.describeEnum(allowed)
-                        + ". Got: " + Self.describeValue(value) + ".",
-                    path: path
-                )
-            }
+        // `enum` short-circuits on both primitives and composite values and
+        // produces the clearest model-facing message, so it runs first.
+        if let failure = checkEnum(value: value, schemaObject: schemaObject, path: path) {
+            return failure
         }
 
         // `type` — may be a single string or an array of strings (union).
-        if let typeValue = schemaObject["type"] {
-            if let failure = checkType(value: value, typeValue: typeValue, path: path) {
-                return failure
-            }
+        if let typeValue = schemaObject["type"],
+           let failure = checkType(value: value, typeValue: typeValue, path: path) {
+            return failure
         }
 
-        // Per-kind keywords.
+        return validatePerKindKeywords(value: value, schemaObject: schemaObject, path: path)
+    }
+
+    private func checkEnum(
+        value: JSONSchemaValue,
+        schemaObject: [String: JSONSchemaValue],
+        path: [String]
+    ) -> ValidationFailure? {
+        guard let enumValues = schemaObject["enum"],
+              case let .array(allowed) = enumValues else {
+            return nil
+        }
+        if allowed.contains(value) { return nil }
+        return ValidationFailure(
+            modelReadableMessage: "argument \(Self.describePath(path)) must be one of: "
+                + Self.describeEnum(allowed)
+                + ". Got: " + Self.describeValue(value) + ".",
+            path: path
+        )
+    }
+
+    private func validatePerKindKeywords(
+        value: JSONSchemaValue,
+        schemaObject: [String: JSONSchemaValue],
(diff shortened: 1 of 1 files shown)


roryford (commented):
## Review summary

Code-motion-only refactor verified. All three CC reductions confirmed by `lizard 1.22.1` on the PR branch:

| function | claimed | lizard reports |
|---|---|---|
| `validate` (recursive core, lines 96-128) | 6 | **6** |
| `valueMatchesType` (lines 214-225) | 8 | **8** |
| `validateObject` (lines 262-295) | 5 | **5** |

`lizard` reports zero functions over the CC=15 threshold for this file (down from 3). File total: 472 NLOC, avg CCN 4.4.

### Looks good

- **Behavior preservation walked helper-by-helper:**
  - `checkEnum`: original `if let enumValues, case .array(allowed) = ...` → new `guard let ... else { return nil }; if allowed.contains { return nil }; return failure`. Equivalent in all four states (missing key / non-array / present+matching / present+mismatching).
  - `validatePerKindKeywords`: each `case` previously did `if let failure = ... { return failure }` then fell through to `return nil`; new form `return validateX(...)` is equivalent because each `validateX` already returns `nil` on success.
  - `isInteger`: AND order swapped from `n.rounded() == n && n.isFinite` to `n.isFinite && n.rounded() == n`. Both AND-commutative; verified equivalent for finite, NaN, ±infinity inputs.
  - `validateRequired`: schema-order iteration (`for entry in requiredArray`) preserved — the `test_missingRequired_multiple_firstIsReported` invariant holds.
  - `resolvePropertySchemas`: small private `PropertySchemaResolution` enum keeps the failure path explicit. Missing `properties` → `.success([:])` matches original `propertySchemas = [:]` branch.
  - `validateAdditionalProperties`: sorted-key iteration preserved; the doubled `propertySchemas.keys.sorted().joined(...)` call is now hoisted into one `allowedNames` local — pure cleanup, message string identical.
  - `validateProperties`: nested `if let childValue` collapsed to `guard let childValue else { continue }` — identical semantics (skip property when value not present).

- **Visibility:** every new helper is `private`. `PropertySchemaResolution` is a `private enum`. `Sendable` conformance untouched. Public API surface unchanged.

- **Comments:** all preserved comments explain *why* (enum-first short-circuit, required-first ordering, sorted-key determinism). No what-comments introduced. Helper names (`isObject`, `validateRequired`, `resolvePropertySchemas`, `validateAdditionalProperties`, `validateProperties`) are obvious.

- **Sabotage targets in `JSONSchemaValidatorTests` still align:** `test_missingRequired_returnsFailureWithFieldName` exercises `validateRequired`'s loop body; `test_unsupportedAnyOf_failsClosed` exercises the still-in-place `rejectUnsupportedKeywords` call.

### Non-blocking

- **`isBoolean` / `isNull` direct coverage is thin in `JSONSchemaValidatorTests`** — no schema with `type: "boolean"` or bare `type: "null"` is asserted in this file (the union test `test_union_acceptsBoth` exercises `isNull` via the union path). The real-world corpus likely hits these, but a one-liner `XCTAssertNil(validator.validate(json("true"), against: json(#"{\"type\":\"boolean\"}"#)))` would pin the new predicate directly. Pre-existing gap, not introduced by this PR.

- **`resolvePropertySchemas` failure path (`'properties' must be an object`) has no unit test** in this file. Also pre-existing — the original `validateObject` had the same uncovered branch — but the new helper makes it more visible.

Neither is blocking; both are tracked-elsewhere-or-followup material.

### Verdict

Pure code motion, CC claims independently verified, behavior preservation walked end-to-end, tests pass per author. Ready to merge from a code-review standpoint.

PR #970, opened 2026-05-02
refactor(backends): consolidate SSE/NDJSON parsers behind SSEPayloadHandler (#964)

roryford:
## Summary

Decompose the three highest-CC stream-parser methods in the cloud + Ollama cohort into focused per-step helpers. Behaviour is byte-identical — code motion that takes the per-payload chains and splits them into small named methods so the giant loop bodies become flat dispatchers.

Closes part of #964.

## Cohort cyclomatic complexity (lizard, success metric)

| Function | Before | After |
|---|---|---|
| `OpenAIBackend.parseResponseStream` | 24 | **8** |
| `OpenAIResponsesBackend.handleEvent` | 19 | **10** |
| `OllamaBackend.handleLine` | 25 | **5** |
| **Cohort average** | **22.7** | **7.67** |

Target was `≤12` average and `≤15` per-handler — comfortably met. After the refactor, none of these three methods appears in lizard's `CC > 15` warning list.

## Files migrated

- `Sources/BaseChatBackends/OpenAIBackend.swift` — added `ChatCompletionsStreamState` (thinking + tool accumulator + finalisation flag) and seven private step methods (`processPrefillProgress` / `processReasoning` / `processVisibleContent` / `processToolDeltas` / `processWholeToolCalls` / `processUsage` / `processFinishReason`). Outer loop is a 6-line dispatcher.
- `Sources/BaseChatBackends/OpenAIResponsesBackend.swift` — added `ResponsesEventKind`, an exhaustive enum over the named-event vocabulary. `handleEvent` switches on the kind and delegates to small per-event helpers; the `if name == ...` chain is gone.
- `Sources/BaseChatBackends/OllamaBackend.swift` — split `handleLine` into five focused inner functions: `processToolCalls`, `processThinking`, `processContent`, `processDoneFlush`, `processUsage`.

## Ollama deferred per Option B

Per the task spec, Ollama uses NDJSON (not SSE), so a faithful migration to the existing `SSEPayloadHandler` would require introducing a sibling `NDJSONStreamParser`. The inline byte loop in `OllamaBackend.parseResponseStream` already keeps the NDJSON-vs-SSE coupling local to one backend, and decomposing `handleLine` alone hit the cohort CC target (5) without growing the parser surface area. Adding `NDJSONStreamParser` would be net-positive for symmetry but is not on the critical path for the CC metric — recommend deferring as a follow-up if/when a second NDJSON backend lands.

## Tests

- All 184 existing `BaseChatBackends` tests pass unchanged (no test edits).
- New `Tests/BaseChatBackendsTests/SSEStreamParserCohortTests.swift` covers the `ResponsesEventKind` lookup table (5 tests) and `ChatCompletionsStreamState` defaults (1 test).
- Sabotage check: replaced `case "response.completed": self = .completed` with `.unknown` — `test_responsesEventKind_classifiesOutputAndCompletionEvents` failed as expected. Reverted before commit.

## Test plan

- [x] `scripts/test.sh --filter BaseChatBackendsTests --disable-default-traits --skip-update` — 184 pass
- [x] Full pre-push XCTest invocation (11 filters) — 2492 pass, 0 fail, 12 skipped
- [x] Full pre-push Swift Testing invocation — 83 pass
- [x] All-traits build sweep `swift build --build-tests --traits MLX,Llama,MCPBuiltinCatalog,Ollama,CloudSaaS,HuggingFace` — green
- [x] Sabotage check on new branch (caught by `SSEStreamParserCohortTests`)

Behaviour-byte-identical refactor; existing cloud + Ollama coverage (`OpenAIBackendTests`, `OpenAIResponsesBackendTests`, `OllamaBackendTests`, `CloudThinkingTokenTests`, `CloudBackendSSETests`, `OpenAIBackendToolCallingTests`, `OpenAIResponsesBackendToolCallingTests`, `OllamaBackendToolCallingTests`) is the real gate.

This PR is the candidate for `/ultrareview` per the task brief.

### DIFF
--- a/Sources/BaseChatBackends/OllamaBackend.swift
+++ b/Sources/BaseChatBackends/OllamaBackend.swift
@@ -496,159 +496,175 @@ public final class OllamaBackend: SSECloudBackend, CloudBackendURLModelConfigura
             }
         }
 
-        func handleLine(_ line: String) throws {
-            guard let parsed = Self.parseLine(line) else { return }
-
-            // Tool calls first: Ollama can emit multiple tool_calls in a
-            // single assistant message. Dispatch them in emission order so
-            // the coordinator's serial dispatch loop sees them in the same
-            // order the model produced them.
-            //
-            // Event-shape contract (PR #783): every tool call surfaces as a
-            // uniform start + single arguments-delta + toolCall triple, even
-            // for whole-call backends like Ollama. That keeps consumers
-            // (orchestrator, UI) on a single code path regardless of whether
-            // the underlying transport streams arguments incrementally.
-            //
-            // TODO(#753): Some Ollama configs (notably `qwen2.5:7b` against
-            // newer servers) reportedly emit incremental tool_call deltas
-            // across multiple NDJSON lines. v1 treats Ollama as whole-call
-            // only; if we observe incremental deltas in the wild, lift
-            // `StreamingToolCallAccumulator` from `OpenAIToolEncoding.swift`
-            // and key by tool-call index here. `streamsToolCallArguments`
-            // would flip to `true` at the same time.
-            if let toolCalls = parsed.toolCalls, !toolCalls.isEmpty {
-                for call in toolCalls {
-                    // Cancellation contract: a consumer that drops the
-                    // stream mid-flight must NOT observe `.toolCall` events
-                    // for entries the orchestrator never agreed to dispatch.
-                    // Honour `Task.isCancelled` at the emit boundary so a
-                    // single-line tool_calls payload arriving alongside the
-                    // cancel doesn't fire a phantom dispatch.
-                    if Task.isCancelled { return }
+        // Tool calls first: Ollama can emit multiple tool_calls in a single
+        // assistant message. Dispatch them in emission order so the
+        // coordinator's serial dispatch loop sees them in the same order
+        // the model produced them.
+        //
+        // Event-shape contract (PR #783): every tool call surfaces as a
+        // uniform start + single arguments-delta + toolCall triple, even
+        // for whole-call backends like Ollama. That keeps consumers
+        // (orchestrator, UI) on a single code path regardless of whether
+        // the underlying transport streams arguments incrementally.
+        //
+        // TODO(#753): Some Ollama configs (notably `qwen2.5:7b` against
+        // newer servers) reportedly emit incremental tool_call deltas
+        // across multiple NDJSON lines. v1 treats Ollama as whole-call
+        // only; if we observe incremental deltas in the wild, lift
+        // `StreamingToolCallAccumulator` from `OpenAIToolEncoding.swift`
+        // and key by tool-call index here. `streamsToolCallArguments`
+        // would flip to `true` at the same time.
+        func processToolCalls(_ parsed: ParsedLine) throws {
+            guard let toolCalls = parsed.toolCalls, !toolCalls.isEmpty else { return }
+            for call in toolCalls {
+                // Cancellation contract: a consumer that drops the stream
+                // mid-flight must NOT observe `.toolCall` events for
+                // entries the orchestrator never agreed to dispatch. Honour
+                // `Task.isCancelled` at the emit boundary so a single-line
+                // tool_calls payload arriving alongside the cancel doesn't
+                // fire a phantom dispatch.
+                if Task.isCancelled { return }
+                try noteEventYielded()
+                continuation.yield(.toolCallStart(callId: call.id, name: call.toolName))
+                if !call.arguments.isEmpty {
                     try noteEventYielded()
-                    continuation.yield(.toolCallStart(callId: call.id, name: call.toolName))
-                    if !call.arguments.isEmpty {
-                        try noteEventYielded()
-                        continuation.yield(.toolCallArgumentsDelta(
-                            callId: call.id,
-                            textDelta: call.arguments
-                        ))
-                    }
-                    try noteEventYielded()
-                    continuation.yield(.toolCall(call))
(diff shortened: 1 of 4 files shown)


roryford (commented):
## Review — `/ultrareview`

Pulled the diff, walked all three before/after pairs side-by-side, verified CC numbers with `lizard`, traced state-carrier semantics, and audited the cancellation contract on Ollama. Behaviour is preserved on the OpenAI cohort. One real (narrow) regression in the Ollama path. Test coverage is thin but defensible for code-motion.

### Blocking

**1. `OllamaBackend.processToolCalls` weakens the mid-line cancellation contract.**

In the original, the inner loop did:
```swift
for call in toolCalls {
    if Task.isCancelled { return }   // returns from handleLine entirely
    …
}
```
After the refactor, the same line lives in `processToolCalls`, so the bare `return` only exits that helper — `handleLine` then proceeds to `processThinking` / `processContent` / `processDoneFlush` / `processUsage` on the same NDJSON line. The outer byte loop's `if Task.isCancelled { break }` only fires *between* lines, so a single line carrying `tool_calls[]` + `thinking` / `content` can now yield post-cancel `.thinkingToken` / `.token` events that the original suppressed. Window is narrow (multi-tool-call line with concurrent content, cancel arriving mid-iteration), and `continuation.yield` on a finished stream is a no-op in practice, but the original PR #783 contract was explicit: "no events on this line after cancellation." Existing `test_cancellation_midStream_doesNotEmitToolCallAfterStop` only exercises the *between-line* break, so this regression won't show up in CI.

Fix is a one-liner — either propagate a cancel signal back via the `Bool` return convention you already use in `processContent`/`processDoneFlush`:
```swift
func processToolCalls(_ parsed: ParsedLine) throws -> Bool {  // false ⇒ stop
    …
    if Task.isCancelled { return false }
    …
    return true
}
…
guard try processToolCalls(parsed) else { return }
```
or just hoist the cancellation check up to `handleLine` between steps. Either way please add a regression test (cancel after a content event, then a single line with `tool_calls[]` + non-empty `message.content` + `thinking`, assert no `.token` / `.thinkingToken` after cancel).

### Non-blocking

**2. Cohort tests are thinner than the brief asks for.** `SSEStreamParserCohortTests` covers the `ResponsesEventKind` lookup table (5 tests) and `ChatCompletionsStreamState` defaults (1 test). It does *not* feed canned SSE/NDJSON fixtures and assert emitted `[GenerationEvent]` vectors for thinking-only / tool-call / mixed thinking+content / terminal `done` / error / empty cases. That's defensible for a pure code-motion refactor — the `OpenAIBackendTests` / `OpenAIResponsesBackendTests` / `OllamaBackendTests` / `CloudThinkingTokenTests` / tool-calling suites already drive end-to-end fixtures through the new helpers — but it does mean a regression in the per-step state transitions surfaces as an integration failure rather than a tight diagnostic. Fine to defer; flagging that the bar from #964 isn't quite met.

**3. Option B (deferring `NDJSONStreamParser`) is the right call.** Cohort CC dropped to 7.67 (target ≤12) without it. The NDJSON byte loop in `OllamaBackend.parseResponseStream` is still local to one backend; symmetry with `SSEPayloadHandler` is net-positive but not blocking. Recommend an issue tracking it for whenever a second NDJSON backend lands — don't open it speculatively now (per CLAUDE.md issue hygiene).

**4. `ChatCompletionsStreamState` design is correct.** `StreamingArgumentAccumulator` is a `final class` (alias `StreamingToolCallAccumulator`), so `let toolAccumulator` in the struct still mutates the same instance through the `inout` chain. `thinking` and `finalisedToolCalls` are `var`. All previously-closure-captured locals are accounted for; no shared-state aliasing or lost-update risk in the OpenAI Chat Completions path.

**5. `ResponsesEventKind` switch preserves semantics.** Old code's `if name == "..."` chain matched on first hit; the new exhaustive switch matches on exact name. Each event name maps to one case → one handler. Reasoning suffix variants (`reasoning_summary_text` vs `reasoning_summary`) are coalesced correctly. Function-call lifecycle ordering (`output_item.added` → `function_call_arguments.delta` → `function_call_arguments.done` → `response.completed`) is preserved by the SSE wire order, not by the dispatcher. Note the old comment that listed `response.output_item.added` as "unknown silent" was stale — the new code (correctly) gives it its own case.

**6. CC claims verified independently with `lizard`.**
- `OpenAIBackend.parseResponseStream`: 24 → 8
- `OpenAIResponsesBackend.handleEvent`: 19 → 10
- `OllamaBackend.handleLine`: 25 → 5
- Cohort average: 22.7 → 7.67 ✓

### Looks good

- Pure code-motion in the OpenAI Chat Completions path: `reasoning_content` vs `reasoning` fallback (in `parseReasoningDelta`), `StreamingToolCallAccumulator` upsert/sticky-id semantics, `finish_reason: tool_calls` finalisation gate, `usage` emission timing, `ThinkingBlockManager` open/flush handoff — all preserved byte-for-byte across the helper boundaries.
- `processWholeToolCalls` still skips when `state.finalisedToolCalls` is set, preventing double emission when streaming + whole-message shapes coexist.
- Outer-catch and stream-end-fallback semantics for both OpenAI files preserved (flush thinking before rethrow; finalise tool calls only when `!Task.isCancelled`).
- Helper naming (`process*` for steps that may emit, `emit*` for the per-item yielders, `handle*` in Responses for named-event delegates) is consistent within each file.
- New cohort test file is correctly trait-gated under `#if CloudSaaS` and uses `@testable import` for both `BaseChatBackends` and `BaseChatInference`.

### Recommendation

Hold for the cancellation fix in `processToolCalls` + a regression test. Everything else can land as-is or as follow-ups.

Issue #597, opened 2026-04-20
MLX and Llama backends don't honour maxThinkingTokens=0


roryford:
## Context

PR #596 (`feat!: honor maxThinkingTokens=0 to disable thinking on Ollama …`) wired the `maxThinkingTokens=0` → *disable thinking entirely* semantic on `OllamaBackend` only. `GenerationConfig.maxThinkingTokens`'s public doc-comment promises this works "on supporting backends (Ollama with thinking-capable models, MLX/Llama with reasoning GGUFs)," but MLX and Llama currently **do not** honour the `0` case — they ignore it and behave as if `nil` were passed.

## Current per-backend behaviour

| Backend | `nil` | `0` | `N` |
|---|---|---|---|
| Ollama | reserves 2048 on thinking models only | `think:false` + no thinking budget | cap at N |
| Llama | `ThinkingParser` active per template | **treated as nil** | cap at N via `LlamaGenerationDriver` |
| MLX | `ThinkingParser` active per template | **treated as nil** | cap at N (post-#591) |
| Cloud (Anthropic) | thinking block sent if `N` set | **treated as nil** | budget_tokens=N |

## Fix

- **LlamaGenerationDriver**: when `maxThinkingTokens == 0`, short-circuit the `ThinkingParser` to discard or never-emit `<think>` content, and re-route any post-`</think>` content straight to `.token` events.
- **MLXBackend**: same treatment — parser instance should refuse to enter thinking state when `0`.
- **AnthropicBackend**: omit the `thinking` request block entirely when `maxThinkingTokens == 0` (Anthropic has no "0" equivalent — you either enable thinking or you don't).

## Tests

- Regression: `maxThinkingTokens=0` + thinking model → zero `.thinkingToken` events, zero `.thinkingComplete`, visible response non-empty. Add parallel tests to MLX + Llama thinking test suites.

## Related
Context: project_thinking_model_plan_2026_04.md (April 2026 super-session).
PR #596 shipped the Ollama half.

PR #971, opened 2026-05-02
refactor(backends): extract MLXGenerationDriver, drop generate CC from 62 to 18 (#964)

roryford:
Closes a slice of #964 (cyclomatic-complexity reduction).

## Summary

`MLXBackend.generate(prompt:systemPrompt:config:)` was 363 LOC at CC 62. It now collapses to a ~80-line orchestrator at **CC 18** (target ≤20) by lifting four pure helpers and moving the streaming loop into a new `MLXGenerationDriver` struct that mirrors `LlamaGenerationDriver`.

| Function | NLOC before | CC before | NLOC after | CC after |
|---|---|---|---|---|
| `MLXBackend.generate` | 279 | **62** | 131 | **18** |
| `MLXGenerationDriver.run` | — | — | 90 | **29** (target ≤30) |
| `MLXBackend.buildQwenToolBlock` | — | — | 27 | 4 |
| `MLXBackend.buildChatMessages` | — | — | 43 | 12 |
| `MLXBackend.resolveThinkingMarkers` | — | — | 7 | 2 |
| `MLXBackend.prepareInputAndCache` | — | — | 45 | 7 |

## Approach (one commit per extraction)

1. `buildQwenToolBlock(config:dialect:) -> String?` — Qwen 2.5 `<tools>…</tools>` block builder.
2. `buildChatMessages(...)` — assembles both shapes the MLX container's `prepare(...)` overloads need (`[Chat.Message]?` for vision, `[[String:String]]` for text).
3. `resolveThinkingMarkers(config:autoDetected:) -> ThinkingMarkers?` — config override → auto-detected, with the `maxThinkingTokens == 0` short-circuit (#597).
4. `prepareInputAndCache(...)` — prepare → makeCache → KV-reuse heuristic. **The reuse path is byte-identical** (`longestCommonPrefixLength` clamped to `promptTokenIds.count - 1`, gated by `isSupportedPromptCache` and `count > 1`) — the v0.5.3 KV-cache incident is the reason this matters.
5. `MLXGenerationDriver` — stateless `@MainActor` struct mirroring `LlamaGenerationDriver`'s shape. Owns the streaming loop, both parsers' two-stage routing, `maxOutputTokens`/`maxThinkingTokens` budgets, and the cooperative `microseconds(50)` yield. Lifecycle (snapshot scheduling, `setPhase(.done/.failed)`, `continuation.finish()`) stays in `MLXBackend.generate`.

The driver is `@MainActor` (unlike Llama's, which is not) because every MLX call shares the single-threaded GPU scheduler. The redundant `await MainActor.run { generationStream.setPhase(...) }` wrappers became direct calls — the Task already runs on `@MainActor`.

## Code motion only — zero behavior change

Every gate, parser routing path, budget enforcement, yield cadence, and snapshot-scheduling decision is preserved. The two nested `PromptCacheSnapshot`/`CachedLayerState` types were widened from `private` to default-internal so the lifted `prepareInputAndCache` static can take them in its signature.

## Tests

- 13 new XCTest unit tests in `MLXBackendHelpersTests.swift` cover the three pure helpers (`buildQwenToolBlock`, `resolveThinkingMarkers`, `buildChatMessages`). Hardware-free — they run in CI.
- The 4th helper (`prepareInputAndCache`) and `MLXGenerationDriver.run` stay covered by existing `MLXBackendGenerationTests` (mock container, no Metal) and `BaseChatMLXIntegrationTests` (Xcode-only, real models).
- **Sabotage check passed locally**: removing the `maxThinkingTokens == 0` short-circuit from `resolveThinkingMarkers` makes `test_resolveThinkingMarkers_returnsNilWhenMaxThinkingTokensIsZero` fail. Reverted before commit.
- Updated `Tests/BaseChatInferenceTests/silent_catch_allowlist.txt` so `SilentCatchAuditTest` finds the moved `try? await Task.sleep(for: .microseconds(50))` line at its new path under `MLX/MLXGenerationDriver.swift`.

## Test plan

- [x] `swift build --traits MLX --skip-update` green
- [x] `scripts/test.sh ... --disable-default-traits --skip-update` (XCTest invocation): 2488 passed / 0 failed
- [x] `scripts/test.sh --filter BaseChatInferenceSwiftTestingTests --disable-default-traits --skip-update`: 83 passed / 0 failed
- [x] `swift test --filter BaseChatBackendsTests --traits MLX --skip-update`: 214 passed (3 skipped) — including the 13 new helper tests and all existing MLX coverage
- [ ] CI green

### DIFF
--- a/Sources/BaseChatBackends/MLX/MLXGenerationDriver.swift
+++ b/Sources/BaseChatBackends/MLX/MLXGenerationDriver.swift
@@ -1,0 +1,158 @@
+#if MLX
+import Foundation
+import MLXLMCommon
+import os
+import BaseChatInference
+
+/// Owns the token-streaming loop for a single `MLXBackend.generate()` call.
+///
+/// `MLXGenerationDriver` is stateless — every dependency it needs is passed
+/// as an explicit parameter to `run()`. It mirrors `LlamaGenerationDriver`'s
+/// shape but runs `@MainActor` because every MLX call (`prepare`, `makeCache`,
+/// `generate`, KV-cache snapshot capture) shares the single-threaded GPU
+/// scheduler with the rest of the MLX runtime.
+@MainActor
+struct MLXGenerationDriver {
+
+    private static let logger = Logger(
+        subsystem: BaseChatConfiguration.shared.logSubsystem,
+        category: "inference"
+    )
+
+    /// Outcome of a `run(...)` call.
+    struct RunResult {
+        /// `true` when the loop completed without throwing and was not cancelled —
+        /// callers may snapshot the prompt cache for next-turn reuse.
+        let completedNormally: Bool
+    }
+
+    /// Drives the MLX stream:
+    ///   1. Calls `container.generate(...)` to materialise the underlying token stream.
+    ///   2. Routes each chunk through the optional tool-call parser, then the optional
+    ///      thinking parser.
+    ///   3. Enforces `config.maxOutputTokens` and `config.maxThinkingTokens`.
+    ///   4. Issues a cooperative `Task.sleep(50µs)` (or the test hook) every
+    ///      `config.yieldEveryNTokens` chunks to keep the WindowServer GPU queue moving.
+    ///   5. Flushes both parsers' tail buffers on exit.
+    ///
+    /// On any thrown error the caller wraps the call in a do/catch and is
+    /// responsible for setting the `generationStream`'s failure phase. This
+    /// helper only yields events into `continuation`; it does not call
+    /// `continuation.finish()` — the caller owns lifecycle.
+    func run(
+        container: any MLXModelContainerProtocol,
+        generationInput: MLXPreparedInput,
+        cache: MLXPromptCache,
+        generateConfig: GenerateParameters,
+        config: GenerationConfig,
+        dialect: MLXToolDialect,
+        markers: ThinkingMarkers?,
+        generationStream: GenerationStream,
+        continuation: AsyncThrowingStream<GenerationEvent, Error>.Continuation,
+        yieldHook: (@Sendable () async -> Void)?
+    ) async throws -> RunResult {
+        Self.logger.debug("MLXGenerationDriver run started")
+
+        let outputLimit = config.maxOutputTokens
+        var outputTokenCount = 0
+        var isFirstToken = true
+
+        let useThinkingParser = markers != nil
+        // ThinkingParser must always be initialized (its initializer can't take nil).
+        // When useThinkingParser is false the parser is never invoked, so the
+        // placeholder marker pair is never observed.
+        var thinkingParser = ThinkingParser(markers: markers ?? .qwen3)
+
+        // Tool-call parser activates only when tools are configured AND the model
+        // speaks a known dialect. It's a no-op pass-through otherwise.
+        let useToolParser = !config.tools.isEmpty && dialect != .unknown
+        var toolParser = MLXToolCallParser()
+
+        // A thinking model that runs away on a 16 GB Mac can OOM mid-generation;
+        // the budget gate breaks out of the stream once the limit is reached.
+        var thinkingTokenCount = 0
+        var thinkingLimitReached = false
+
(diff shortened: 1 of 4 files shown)


roryford (commented):
## Review

Code-motion refactor with the discipline I'd hope for. KV-cache reuse — the highest-risk piece per the v0.5.3 incident — is byte-identical, the lifecycle split is clean, and the `@MainActor` annotation on the driver is the truthful name for what was already happening.

### Looks good

**KV-cache reuse — verified byte-identical** (`prepareInputAndCache` vs original lines 870–912). All five gates (`kvCacheReuseEligible`, `let snapshot`, `let promptTokenIds`, `isSupportedPromptCache(cache.value)`, `promptTokenIds.count > 1`), the `min(commonPrefixLen, promptTokenIds.count - 1)` clamp, the `> 0` check before restore, and the `preparedInput.suffix(from:)` call are all preserved. The `kvCacheReuse(promptTokensReused:)` event moved from inside the if-block to the caller, gated by `prepared.reuseLen > 0` — emits on the same condition.

**`@MainActor` is correct, not a hazard.** The original `Task { @MainActor [weak self, generationStream] in ... }` (main, line 826) already pinned the entire streaming loop to MainActor — `await MainActor.run { ... }` wrappers were redundant. Annotating the driver `@MainActor` makes that isolation explicit and lets those wrappers become direct calls. The worker's stop-and-report flag is explainable: not a deviation from Llama for stylistic reasons, but a forced choice because MLX's GPU scheduler is single-threaded and the original loop was already MainActor-bound.

**Lifecycle split is clean.** The driver returns one bit (`completedNormally: !Task.isCancelled`) and mutates no backend state. Snapshot scheduling (`_promptCacheWriteToken`, `_pendingPromptCacheSnapshotTask`, `withStateLock`) stays in `MLXBackend.scheduleSnapshotCaptureLocked`, gated on `result.completedNormally && prepared.promptTokenIds`. No hidden coupling.

**Mirror fidelity vs Llama.** Same parameter discipline — every dependency passed in, no `MLXBackend` reference captured. Only structural delta is `@MainActor`, which is principled.

**CC claims verified with lizard:** `generate` 62 → 18, `run` = 29, `prepareInputAndCache` = 7, `buildChatMessages` = 12, `buildQwenToolBlock` = 4, `resolveThinkingMarkers` = 2. All within stated targets.

**Streaming-loop body byte-identical** against main lines 920–985 (stage-one/two routing, `outputTokenCount` gate, thinking-token budget short-circuit, `yieldEveryNTokens` cadence, both finalize() flushes).

**Allowlist update is a pure path move** — same `try? await Task.sleep(for: .microseconds(50))` line, relocated to the new file.

### Non-blocking nits

**Visibility widening looks unnecessary.** `PromptCacheSnapshot` and `CachedLayerState` were widened from `private` to default-internal. But `prepareInputAndCache` is a `static func` inside `MLXBackend` — a `private` (or `private static`) func in the same class can absolutely reference its sibling private nested types. `PreparedGenerationInputs` doesn't expose `PromptCacheSnapshot` in its fields. I grepped the module and these types have zero out-of-class references. Recommend reverting both to `private` to keep the lift tightly scoped.

**Driver `run` lacks direct unit coverage.** 13 helper tests are good. `prepareInputAndCache` and `MLXGenerationDriver.run` are covered transitively via the existing `MLXBackendGenerationTests` (which do exercise KV reuse with mock containers — `test_generate_reusesPromptCachePrefixOnMatchingTurn` etc.). Worth a follow-up to add a direct-injection test that constructs a driver and feeds it a mock `MLXModelContainerProtocol` so the orchestrator's plumbing isn't required to assert driver behavior. Not worth blocking this PR for.

### Verdict

LGTM to merge once CI is green. The lift preserves behavior on the path that has burned us before (KV reuse), the actor-isolation choice is principled, and the lifecycle split has no hidden coupling. Recommend a follow-up issue to revert `PromptCacheSnapshot`/`CachedLayerState` back to `private` and add a direct unit test on `MLXGenerationDriver.run` with an injected mock container.

PR #974, opened 2026-05-02
refactor(server): collapse BaseChatServerCore + BaseChatServerBackends into BaseChatServer

### LINKED ISSUE #949: refactor(server): collapse BaseChatServerCore + BaseChatServerBackends into BaseChatServer
The Server module is currently split three ways: `BaseChatServerCore` (library, 8 files), `BaseChatServerBackends` (library, 2 files), and `BaseChatServer` (executable, 1 file). All three are now trait-gated together behind the `Server` trait shipped in #946. Nothing else in the workspace links any one of them in isolation.

The 3-way split predates the `Server` trait. With trait-gating in place, the boundary inside the server module is pure overhead — extra module files to compile, extra cross-module linker work, extra `package`/`public` bookkeeping for symbols that only ever cross between server-internal modules.

## Current state (verified 2026-05-02 on `main`)

- `Sources/BaseChatServerCore/` — 8 files. Imports `BaseChatInference` + `Hummingbird`.
- `Sources/BaseChatServerBackends/` — 2 files. Imports `BaseChatServerCore` + `BaseChatInference` + `BaseChatBackends`.
- `Sources/BaseChatServer/main.swift` — 1 file (24 LOC, with the `#else` no-op stub from #946 keeping it linkable when the trait is off).
- `Tests/BaseChatServerTests/` — depends on `BaseChatServerCore` + `BaseChatServerBackends` + `BaseChatInference` + `BaseChatTestSupport`.
- All four targets share `swiftSettings: [.define(\"Server\", ...)]` and the same trait-condition.

## Proposal

Merge the 8 + 2 + 1 source files into a single `BaseChatServer` target. Keep the `BaseChatServer` library product (preserves the `import BaseChatServer` surface) plus the `BaseChatServer` executable product (re-pointed at the same target with a separate `main.swift` entry point, or use `@main` on a single source file).

If consumers genuinely need to depend on the server's domain-model layer without the executable wiring, that would justify a split — but no current consumer does. (Verified: `grep -rn \"import BaseChatServerCore\\|import BaseChatServerBackends\" Sources/ Tests/` returns only files inside the server-tests target itself.)

## Why finish this now

- The trait-gating PR established `Server` as the right unit of consumability. The internal split is now decoration.
- Smaller module count = fewer linker steps in the `server-tests` CI job (currently ~282s, mostly build).
- Removes `package`-visible types that only exist to cross the `Core`→`Backends` boundary.
- Aligns with the broader \"reduce module sprawl\" theme — same logic as #948 (close the `package` hatch by removing the boundary that needed it).

## PR-sized batches

This is one PR.

- [ ] Move `Sources/BaseChatServerCore/*.swift` and `Sources/BaseChatServerBackends/*.swift` into `Sources/BaseChatServer/`. Preserve `#if Server` guards.
- [ ] Update `Package.swift`: delete `BaseChatServerCore` + `BaseChatServerBackends` target declarations and the `BaseChatServerCore` library product. Update `BaseChatServer` executable target to depend directly on `BaseChatInference` + `BaseChatBackends` + `Hummingbird` (with the existing `condition: .when(traits: [\"Server\"])` on Hummingbird).
- [ ] Update `Tests/BaseChatServerTests/` deps in `Package.swift` to point at `BaseChatServer` instead of `BaseChatServerCore` + `BaseChatServerBackends`.
- [ ] Audit `package` visibility — any symbol that was `package` purely to cross the Core/Backends boundary can drop back to `internal`.
- [ ] Update CLAUDE.md target table.
- [ ] Drop docs references to `BaseChatServerCore` as a separately-consumable library (README + CHANGELOG breaking-change note).

## Out of scope

- Any reshape of the OpenAI request/response Codables or the Hummingbird wiring — purely a module-boundary fix.
- The IOS-AI-Server resume planning (separate downstream concern).

## Breaking-change posture

Removing the `BaseChatServerCore` library product breaks any external consumer that imports it directly. No in-tree consumer does. Minor version bump + Highlights note.

roryford:
Closes #949

## Summary

The original three-target server shape (`BaseChatServerCore` library + `BaseChatServerBackends` library + `BaseChatServer` executable) predates the `Server` trait and is now pure overhead — all three are trait-gated together and no in-tree consumer depends on `BaseChatServerCore` or `BaseChatServerBackends` separately.

This PR collapses everything into a single `BaseChatServer` executable target.

### Changes

- Moved 10 source files from `Sources/BaseChatServerCore/` and `Sources/BaseChatServerBackends/` into `Sources/BaseChatServer/`.
- Renamed `Sources/BaseChatServer/main.swift` → `BaseChatServerCommand.swift` so the `@main`-attributed `BaseChatServerCommand` works (single-file `main.swift` triggers top-level-code mode and is incompatible with `@main`).
- Dropped the `BaseChatServerCore` library product and the `BaseChatServerCore` + `BaseChatServerBackends` target declarations from `Package.swift`.
- Updated `BaseChatServerTests` to depend on `BaseChatServer` directly. Tests `@testable import BaseChatServer` (SwiftPM allows test targets to depend on executable targets, so no sibling `.library()` product was needed).
- Demoted all 45 `package` symbol sites to `internal`. They were only `package` to cross the original Core → Backends → executable boundary; with everything in one module, `internal` plus `@testable import` is sufficient. Confirmed no external (in-tree) consumer references any of them.
- Preserved every `#if Server` guard and the `Server` trait definition exactly as-is.
- Added a status note to `docs/plans/744-basechatserver.md` recording the collapse so the historical plan doc stays readable.

### Breaking change

Removing the `BaseChatServerCore` library product **breaks any external consumer that imports `BaseChatServerCore` directly**. No in-tree consumer does. The only public-facing surface remains the `BaseChatServer` executable, which is unchanged behaviorally.

This warrants a minor version bump.

### Behavior

Unchanged. Pure module-boundary refactor — no routing, codable, Hummingbird-wiring, or backend-selection logic was modified.

## Test plan

- [x] `swift build --build-tests --traits Server` — clean
- [x] `swift test --filter BaseChatServerTests --traits Server --skip-update` — 48 tests passing
- [x] `swift build --disable-default-traits` — clean (the `#else` no-op stub keeps `BaseChatServer` linkable when the trait is off)


roryford (commented):
Reviewed — found two stale-path references after the module collapse and fixed them in 98156ca.

## What was wrong

The `BaseChatInferenceTests.SilentCatchAuditTest` was failing in CI (only the `test` job — the targeted `server-tests` job ran the Server-traits subset and was green). Two issues, both consequences of the module rename rather than logic regressions:

1. **Stale allowlist fingerprint.** `Tests/BaseChatInferenceTests/silent_catch_allowlist.txt` had `BaseChatServerCore/ChatCompletionsAdapter.swift:if let string = try? decoder...` as the approved fingerprint. Post-collapse the audit walks `Sources/BaseChatServer/` and produces a `BaseChatServer/...` fingerprint, so it (a) flagged the existing intentional `try?` as unapproved and (b) flagged the old fingerprint as a stale allowlist entry. Updated the section header + fingerprint to the new path.
2. **Stale `main.swift` doc reference.** `Package.swift`'s `BaseChatServer` executableTarget doc comment ended with "see `main.swift`" — that file is now `BaseChatServerCommand.swift`. Comment-only fix.

## What I checked

- **`@main` semantics** — `BaseChatServerCommand.swift` carries `@main` on both branches (`BaseChatServerCommand` under `#if Server`, `BaseChatServerDisabled` stub under `#else`). No other `@main` and no top-level code in any of the 11 source files.
- **`#else` no-op stub** — `swift build --disable-default-traits` is clean; the executable links without dragging in Hummingbird.
- **Test target `@testable`** — all 9 server test files use `@testable import BaseChatServer`. The `internal` demotion is sound.
- **No leftover module references** — `grep -rn "BaseChatServerCore\|BaseChatServerBackends" Sources/ Tests/ Example/` is empty. The two remaining hits in `*.md` are correctly historical (CHANGELOG.md describes shipped releases; `docs/plans/744-basechatserver.md` already has an "Update (#949)" prologue calling out the collapse).
- **`package` → `internal` correctness** — spot-checked `ServerBackendProvider`, `ChatCompletionRole`, `ChatCompletionMessage`, `ServerCommandOptions`, `TraitAwareServerBackendProvider`. Each is a pure visibility narrowing — the only non-visibility deltas are the now-unnecessary `import BaseChatServerCore` lines being removed.
- **Package.swift correctness** — `BaseChatServerCore` library product gone; both library targets gone; executable deps are `BaseChatInference` + `BaseChatBackends` + ArgumentParser + `Hummingbird` (Hummingbird is the only one carrying `condition: .when(traits: ["Server"])`); test target depends on `BaseChatServer`; both targets carry `.define("Server", .when(traits: ["Server"]))`; the `Server` trait declaration is preserved.
- **Behavior preservation** — `git diff main..HEAD --find-renames=50 -- Sources/BaseChatServer/` shows pure file-move + visibility-narrowing + import-cleanup. Nothing reshapes the OpenAI Codables or Hummingbird wiring.

## Test gate

```
swift build --build-tests --traits Server                     # OK
swift test --filter BaseChatServerTests --traits Server       # 48/48 pass
swift test --filter SilentCatchAuditTest --disable-default-traits  # pass after fix
swift build --disable-default-traits                          # OK (#else stub links)
```

Not merging — leaving the merge to you.

Issue #977, opened 2026-05-02
feat(inference): catalog manifest layer over ModelStorageService


roryford:
## Context

`ModelStorageService.discoverModels()` scans the models directory on every call and reflects on each GGUF/MLX artifact to build `[ModelInfo]`. That works for a chat app with O(10) local models but degrades for any host that:

- Tracks more than disk presence (last-used timestamps, eviction priority, source provenance, integrity hashes).
- Wants to enforce a disk budget without re-scanning + re-summing on every check.
- Needs to record metadata that doesn't live in the artifact itself (HF repo ID, download timestamp, expected SHA from #367).

Today every host re-implements its own sidecar bookkeeping. A first-class manifest layer in BCK closes the gap once.

## Proposed shape

A JSON manifest at `<modelsDirectory>/.basechat-catalog.json` (dot-prefixed so it doesn't show up in user-visible directory listings), maintained by a new service that wraps `ModelStorageService`:

```swift
public final class ModelCatalog {
    public init(storage: ModelStorageService)

    /// Manifest-backed model list. Falls back to `discoverModels()` for entries
    /// not in the manifest (e.g. user-imported files dropped into the directory).
    public func catalog() async throws -> [CatalogEntry]

    /// Records a model after a successful download / import.
    public func record(_ entry: CatalogEntry) async throws

    /// Removes a model entry and (optionally) its on-disk artifact.
    public func evict(_ id: ModelInfo.ID, deleteArtifact: Bool) async throws

    /// LRU eviction down to a disk budget. Returns evicted entries.
    public func enforceDiskBudget(_ maxBytes: UInt64) async throws -> [CatalogEntry]

    /// Updates last-used timestamp; called on successful model load.
    public func touch(_ id: ModelInfo.ID) async throws
}

public struct CatalogEntry: Codable, Sendable, Hashable {
    public let id: UUID
    public let modelInfo: ModelInfo
    public let source: ModelSource         // .huggingFace(repo:file:) / .imported / .bundled
    public let downloadedAt: Date
    public var lastUsedAt: Date
    public let expectedSHA256: String?     // composes with #367 Phase 1
    public let sizeBytes: UInt64
    public var quantization: String?
}
```

`InferenceService` calls `catalog.touch(modelID)` on successful load; the catalog handles persistence atomically (write to temp + rename).

## Why composable with existing services, not a replacement

[...]

PR #981, opened 2026-05-02
feat(inference): ModelInfo factory for HuggingFace-sourced models

### LINKED ISSUE #975: feat(inference): ModelInfo factory for HuggingFace-sourced models
## Context

`ModelInfo` has two constructors today:

- `init?(ggufURL: URL)` — reflects on a local GGUF file
- `init?(mlxDirectory: URL, namespace: String?)` — reflects on a local MLX directory
- `init(...)` (full memberwise) — public but inconvenient

Neither covers the common path where a host app has HuggingFace metadata (`repo_id`, `file_name`, `size_bytes`) **before** the artifact has been downloaded. Today the workaround is to either:

1. Use the existing `DownloadableModel` for the pre-download phase and reconstruct a `ModelInfo` post-download via `init?(ggufURL:)`, or
2. Fabricate a placeholder `URL` and call the full memberwise init with hand-filled defaults.

Both paths force every host app implementing an HF browser/download flow to re-derive the same conversion. A first-class factory closes the gap.

## Open design questions

The shape isn't obvious because `ModelInfo.url` is non-optional and currently always a local file URL:

- **Option A** — keep `ModelInfo` as a post-download value type. Add a convenience that takes HF metadata **plus** a local URL once download finishes, and fills `name`/`fileName`/`fileSize`/`modelType` without re-reading the GGUF header. This is purely a build-from-metadata shortcut.
- **Option B** — relax `ModelInfo.url` to optional (or split a sibling type) so a pre-download `ModelInfo` is representable. This is more invasive and probably not warranted given `DownloadableModel` already covers that role.
- **Option C** — extend `DownloadableModel` instead and document it as the pre-download type. `ModelInfo` stays as the post-download type and gains only the metadata-completion convenience from Option A.

Option A + Option C together is probably the right answer: keep the type boundary clean, just remove the redundant GGUF header re-read on the post-download conversion path.

## Concrete proposal (Option A + C)

```swift
extension ModelInfo {
    /// Build a `ModelInfo` from a completed HuggingFace download.
    /// Avoids re-reading the GGUF header by trusting the manifest metadata
    /// the caller already had at download-decision time.
    public init?(
        huggingFaceRepoID: String,
        fileName: String,
        sizeBytes: UInt64,
        localURL: URL,
        modelType: ModelType = .llama
    )
}
```

The init still returns `ModelInfo?` (rather than throwing) to match the existing `ggufURL:` / `mlxDirectory:` factories.

## Why this is worth filing

A non-chat host wanting an HF browser/download flow today either rebuilds this conversion in-app or reaches around `ModelInfo` via `DownloadableModel` and `init?(ggufURL:)`. Both leak BCK internals into the host. Tracking it separately keeps it from getting lost when the bigger model-management decoupling (#950) lands.

## Out of scope

- Changing `ModelInfo.url` to optional (Option B).
- Any change to `DownloadableModel`'s shape.
- Pre-download `ModelInfo` instances (use `DownloadableModel` for that phase).

## Acceptance

- New factory compiles and returns a valid `ModelInfo`.
- Tests cover the happy path and at least one failure mode (e.g. zero-size file).
- DocC mentions the new init alongside the existing factories.
- No behavior change for existing call sites of `init?(ggufURL:)`.

roryford:
## Summary

Adds an additive `ModelInfo` factory for completed HuggingFace downloads:

```swift
extension ModelInfo {
    public init?(
        huggingFaceRepoID: String,
        fileName: String,
        sizeBytes: UInt64,
        localURL: URL,
        modelType: ModelType = .gguf
    )
}
```

The factory **trusts caller-supplied `sizeBytes`** (typically derived from the HuggingFace manifest) and skips the on-disk `attributesOfItem` lookup that `init?(ggufURL:)` performs. It still reads the GGUF header so the value-add stays — `detectedPromptTemplate`, `detectedContextLength`, `modelArchitecture`, `chatTemplateRaw`, and `estimatedKVBytesPerToken` are all populated. The mmproj companion-file scan is intentionally skipped (HF flows are single-file).

`fileName` is used verbatim from the caller, not derived from `localURL` — HF download flows pre-compute the canonical name and the on-disk path may be a cached blob.

## Non-breaking

This is **purely additive**:

- New optional stored property `huggingFaceRepoID: String?` on `ModelInfo` (defaults to `nil`).
- New factory init.
- Memberwise init gains a `huggingFaceRepoID: String? = nil` parameter at the end of the parameter list — existing call sites compile unchanged.
- No behavior change to `init?(ggufURL:)`, `init?(mlxDirectory:namespace:)`, or `ModelInfo.builtInFoundation`.

## New public surface

- `ModelInfo.init?(huggingFaceRepoID:fileName:sizeBytes:localURL:modelType:)`
- `ModelInfo.huggingFaceRepoID: String?`

## Tests

`Tests/BaseChatInferenceTests/ModelInfoHuggingFaceFactoryTests.swift` covers:

- Happy path: valid GGUF + HF metadata returns a fully-populated `ModelInfo`.
- Wrong extension (`.txt`) returns `nil`.
- Missing magic bytes (100-byte zero file with `.gguf` extension) returns `nil`.
- Missing file returns `nil`.
- `huggingFaceRepoID` round-trips onto the resulting model.
- `fileSize` matches caller-supplied `sizeBytes` (verifies no disk lookup).
- `fileName` is taken verbatim from the caller, not from `localURL`.
- Stable ID across repeated construction.

Sabotage check verified: removing the `isValidGGUF` guard makes the missing-magic-bytes test fail; restored before commit.

## Test plan

- [x] `swift build --build-tests --disable-default-traits`
- [x] `scripts/test.sh --filter BaseChatInferenceTests --disable-default-traits --skip-update` — 1233 passed, 0 failed, 5 skipped (XCTSkip)

Closes #975

### DIFF
--- a/Sources/BaseChatInference/Models/ModelInfo.swift
+++ b/Sources/BaseChatInference/Models/ModelInfo.swift
@@ -11,6 +11,12 @@ public enum ModelType: Hashable, Sendable {
 }
 
 /// Represents a model available on disk (either a GGUF file or an MLX model directory).
+///
+/// Construct via one of the factory initializers:
+/// - ``init(ggufURL:)`` for local `.gguf` files (reads size from disk).
+/// - ``init(mlxDirectory:namespace:)`` for MLX model directories.
+/// - ``init(huggingFaceRepoID:fileName:sizeBytes:localURL:modelType:)`` for completed
+///   HuggingFace downloads (trusts caller-supplied size, skips disk attribute lookup).
 public struct ModelInfo: Identifiable, Hashable, Sendable {
     public let id: UUID
     public let name: String
@@ -51,6 +57,15 @@ public struct ModelInfo: Identifiable, Hashable, Sendable {
 
     /// The most recent benchmark result for this model, if one has been run.
     public var benchmarkResult: ModelBenchmarkResult?
+
+    // MARK: - Provenance
+
+    /// The HuggingFace repository this model was downloaded from, when known
+    /// (e.g. `"bartowski/Llama-3.2-3B-Instruct-GGUF"`).
+    ///
+    /// Populated by ``init(huggingFaceRepoID:fileName:sizeBytes:localURL:modelType:)``.
+    /// `nil` for locally-discovered files and built-in models.
+    public var huggingFaceRepoID: String?
 
     /// Returns the benchmark-confirmed tier when one is available, otherwise falls back
     /// to the stored ``capabilityTier``, and finally to a static file-size estimate.
@@ -221,7 +236,8 @@ public struct ModelInfo: Identifiable, Hashable, Sendable {
         chatTemplateRaw: String? = nil,
         estimatedKVBytesPerToken: UInt64? = nil,
         capabilityTier: ModelCapabilityTier? = nil,
-        benchmarkResult: ModelBenchmarkResult? = nil
+        benchmarkResult: ModelBenchmarkResult? = nil,
+        huggingFaceRepoID: String? = nil
     ) {
         self.id = id
         self.name = name
@@ -237,6 +253,7 @@ public struct ModelInfo: Identifiable, Hashable, Sendable {
         self.estimatedKVBytesPerToken = estimatedKVBytesPerToken
         self.capabilityTier = capabilityTier
         self.benchmarkResult = benchmarkResult
+        self.huggingFaceRepoID = huggingFaceRepoID
     }
 
     // MARK: - Private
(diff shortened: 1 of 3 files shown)


roryford (commented):
LGTM — small additive PR, scope-tight. Review-only (cannot self-approve).

**Checked:**

- **API additive only.** New `huggingFaceRepoID: String? = nil` parameter on the memberwise init is at the end of the parameter list with a default — existing callsites unchanged. New stored property is `var` (correct for the #977 catalog write path). `Codable`/`Hashable`/`Sendable`/`Identifiable` conformances hold automatically since `String?` is auto-synthesizable. Existing `init?(ggufURL:)` and `init?(mlxDirectory:namespace:)` rely on Swift's implicit `nil` default for `var x: String?` in structs — verified by `swift build`.
- **Factory failure modes.** Both extension and magic-byte guards run before any `self.X = Y` assignment. `attributesOfItem` is correctly absent from the new factory (still present only in `init?(ggufURL:)`).
- **Tests are meaningful.** The "trust caller-supplied sizeBytes" test supplies 4 GB while the on-disk file is 64 bytes — sabotage-resistant, would fail if the factory secretly fell through to `attributesOfItem`. The "useCallerFileName_notDerivedFromURL" test similarly exercises the divergence between `localURL` basename and `fileName`. Fixture writer uses correct GGUF magic bytes `[0x47, 0x47, 0x55, 0x46]`.
- **silent_catch_allowlist.** New entry at line 43 mirrors the existing pattern at line 42 for `init?(ggufURL:)` — same `try? GGUFMetadataReader.readMetadata` shape. `SilentCatchAuditTest` passes.
- **DocC.** Type-level comment now lists all three factories as a brief 4-line bullet block — matches the existing terseness, no multi-paragraph bloat.
- **CI green.** `test` (4m49s), `security-audits` (5m10s), `fuzz`, `lint`, `macros-tests`, `server-tests` all pass on the latest commit.

**Local test gate (in worktree):** `swift build --build-tests --disable-default-traits` succeeds; `BaseChatInferenceTests` 1233 passed / 5 skipped / 0 failures; `SilentCatchAuditTest` passes.

Issue #1113, opened 2026-05-08
refactor: decompose 7 god-object files (>1000 LOC)


roryford:
## Problem

Seven files in Sources/ exceed 1,000 LOC. Each one's method clusters span 3+ unrelated concerns — a structural audit flagged them as god-objects. For LLM agents, every session that touches one of these files reloads the entire blob into context; for humans, the cohesion problem makes change risk hard to bound.

`InferenceService` already shipped the same treatment in PR #306 (decomp into `ModelLifecycleCoordinator` + `GenerationCoordinator` facade, since renamed `GenerationQueue` in #973). This is the next wave.

### [ ] 1. `BackgroundDownloadManager.swift` (1,345 LOC) — `ManifoldHuggingFace`
Three unrelated concerns:
- **Download lifecycle** — `startDownload`, snapshot/single-file flows, `URLSession` delegate, task tracking
- **Package validation** — `validateDownloadedFile`, `validateDiffusionPackage`, `validateDiffusionComponent`, `diffusionRole`, `writePackageManifest`
- **Temp/cache hygiene + pending-download metadata** — `cleanupStaleTempFiles`, `deleteOrphanedResumeDataFiles`, `persistResumeData`, `loadPendingMetadata`, `writePendingMetadata`, `migrateFromUserDefaults`

Suggested split: extract `DiffusionPackageValidator` + `DownloadHygieneJanitor` + `PendingDownloadStore` (value type). Core stays as the orchestrator.

### [ ] 2. `MCPOAuth.swift` (1,336 LOC) — `ManifoldMCP`
Concerns:
- **Flow orchestration** — `authorize`, `performAuthorizationCodeFlow`, `callbackScheme`, `parseAuthorizationCode`
- **Token store + refresh** — `activeTokens`, `singleFlightRefresh`, `exchangeAuthorizationCode`, `exchangeRefreshToken`, `tokenExchange`
- **Server-metadata discovery** — `discoverAuthorizationMetadata`, `verifyIssuer`, `resolveClientIdentifier`
- **Security primitives** — `enforceHTTPS`, `requireSuccess`, `validateBearerTransmission`, `randomBase64URL`, `constantTimeEqual`, `bearerRedacted`

Suggested split: `MCPOAuthClient` (facade) + `OAuthTokenStore` + `AuthorizationServerDiscovery` + free-function security helpers in their own file.

### [ ] 3. `ConversationRuntime.swift` (1,323 LOC) — `ManifoldRuntime`
[...]

PR #1161, opened 2026-05-09
refactor: decompose ConversationRuntime

roryford:
Part of #1113.

Summary:
- Extracts ConversationTurnExecutor and ConversationPersistencePort.
- Moves legacy input shims, stream handle, and in-flight stream registry into focused files.
- Keeps ConversationRuntime as the public facade with unchanged turn API/event behavior.

LOC:
- ConversationRuntime.swift: 1,477 -> 295.

Validation:
- scripts/test.sh --filter ManifoldRuntimeTests --disable-default-traits --skip-update
- Code review pass: no concurrency, event-ordering, cancellation, or API regressions found.


roryford (commented):
Reviewed the decomposition. Behavior-preserving extraction; structurally sound.

**What I checked**
- Public surface: `SendInput`/`RegenerateInput`/`EditInput`/`BranchInput`/`ConversationStreamHandle`/`ConversationRuntime` (+`processTurn`, `cancel`, deprecated `send`/`regenerate`/`edit`/`branch`) all preserved bit-for-bit; only relocated across files.
- `ConversationRuntime` is `final class: Sendable` (not `@MainActor`-pinned, as documented), so the `Task.detached { [self] in ... }` strong captures in `ConversationTurnExecutor` are correct (matches the "no [weak self] in detached tasks that must complete" rule and avoids the `Task.detached + @MainActor` Swift 6 footgun).
- `ConversationPersistencePort` boundary keeps SwiftData out — uses `MessageStore`/`SessionStore` protocols and plain-Swift `ChatMessageRecord`/`ChatSessionRecord`. Internal struct, not a leaked public protocol.
- `InFlightStreamRegistry` actor is identical to main; cancel/register/unregister semantics unchanged. The pre-register cancel race guard (`isCancelled` check after register) is preserved in `ConversationTurnExecutor.runGenerationTurn`.
- Stream lifecycle: cancel propagation through `markCancelled` → `inferenceService.cancelAsync(token)` works the same way; terminal `unregister` happens before `streamFinished` emit, matching the documented late-cancel-is-noop contract.
- No `try?`, no `DispatchSemaphore.wait()` in any deinit, no new `assertionFailure` traps. Legacy input types correctly marked `@available(*, deprecated, renamed: "TurnInput", ...)`.
- Branch flow's title-resolution semantics: nil-sessionStore path now goes through `ConversationPersistencePort.sessionTitle` which returns the `fallback` ("New Chat") — same observable result as main's inline branch.

**Minor observations (not blockers)**
- `ConversationPersistencePort` exposes 7 distinct `@MainActor`-isolated methods that wrap one-line forwards. Could be simplified post-merge if a follow-up extracts a smaller surface, but this is fine as-is for readability.
- `ConversationTurnExecutor.runGenerationTurn` is still ~350 LOC — reasonable for a turn loop, but a future split (drain loop vs finalisation) is conceivable.

**Verdict**: looks safe to merge once `test`/`server-tests`/`macros-tests` go green. Pure structural refactor with preserved semantics.

Issue #487, opened 2026-04-18
fix(ollama): preserve reasoning model thinking content (currently dropped)


roryford:
## Problem

`OllamaBackend.extractToken(from:)` (`Sources/BaseChatBackends/OllamaBackend.swift:236`) only reads `message.content` from each NDJSON line. Ollama's API returns reasoning content in a separate `"thinking"` field for models that support it (qwen3.5:4b, deepseek-r1, etc.). That field is never read, so the entire reasoning trace is silently discarded — the user waits 30–100 seconds and sees an empty assistant message.

## Repro

```
swift run fuzz-chat --model qwen3.5:4b --iterations 3 --seed 42
```

Output (3-for-3 findings via the new `empty-output-after-work` detector):

```
↑ flaky [empty-output-after-work/silent-empty] hash=fa567e41a546 :: totalMs=99877 raw.empty thinkingRaw.empty error=nil
↑ flaky [empty-output-after-work/silent-empty] hash=34a25ed76c97 :: totalMs=41369 raw.empty thinkingRaw.empty error=nil
↑ flaky [empty-output-after-work/silent-empty] hash=b0edcfe4d3c6 :: totalMs=53920 raw.empty thinkingRaw.empty error=nil
```

Direct curl confirms the data shape:
```
$ curl -s -X POST http://localhost:11434/api/generate -d '{"model":"qwen3.5:4b","prompt":"hello","stream":false,"options":{"num_predict":50}}'
{"response":"","thinking":"Thinking Process:\n\n1. **Analyze the Request:**\n...","done":true,"done_reason":"length",...}
```

## Fix

`extractToken` (and the streaming variant in `parseResponseStream`) need to also read `thinking` (top-level on `/api/generate`) and `message.thinking` (on `/api/chat`), and emit them as `GenerationEvent.thinkingToken(_)` followed by `GenerationEvent.thinkingComplete` when the field transitions back to empty or when `done == true` with a non-empty thinking accumulator.

Also worth confirming what the chat-endpoint payload looks like — `parseResponseStream` parses `message.content`, but the docs suggest `message.thinking` is the corresponding key on `/api/chat`.

## Why this slipped past tests

`OllamaE2ETests.test_realInference_generatesNonEmptyResponse` asserts the response is non-empty — but it uses `llama3.1:8b` (not a reasoning model), so it never exercised the thinking path. The reasoning-model failure mode is asymmetric: response is empty *only* when the model spent its budget reasoning.

## Discovered by

[...]

PR #1163, opened 2026-05-09
refactor: decompose OllamaBackend

roryford:
Part of #1113.

Summary:
- Extracts OllamaMessageEncoder, OllamaModelProbe, OllamaPayloadHandler, and OllamaStreamProcessor.
- Preserves Ollama wire format, manifest probing, SSE/NDJSON streaming, and tool-call cancellation behavior.
- Updates Ollama and cross-backend replay tests to target extracted helpers.

LOC:
- OllamaBackend.swift: 1,304 -> 468.

Validation:
- scripts/test.sh --filter ManifoldBackendsTests --traits Ollama --disable-default-traits --skip-update
- swift build --build-tests --traits Ollama,CloudSaaS --disable-default-traits
- Code review pass: no streaming, cancellation, trait-gating, or API regressions found.


roryford (commented):
## Review summary

**Verdict:** approve once CI is green. One blocking CI failure fixed in 9feb37f; two non-blocking observations below.

### CI fix shipped (9feb37f)
`TrafficBoundaryAuditTest` rule 1 was failing because the two extracted files (`OllamaModelProbe.swift`, `OllamaStreamProcessor.swift`) hold network I/O — `URLSession.AsyncBytes` and a `URLRequest` — but only the original `OllamaBackend.swift` was on `networkIOAllowlist`. Since both files inherit OllamaBackend's network-boundary status by extraction (no new outbound surface — just a file split), I added them to the allowlist with a justification comment pointing back to #1163. Allowlist count is 24/26 after the change. Pushed as a separate commit so the security/audit fix is reviewable independently.

### Decomp review

1. **Public surface unchanged on `OllamaBackend`** — diffed `public` declarations against `origin/main`, identical.
2. **SSE/NDJSON byte-stream semantics preserved in `OllamaStreamProcessor.parse`** — line buffer + `\n` framing, `done` flush, `<think>` fallback engagement, marker-detection precedence (override > probed > qwen3 default), `eval_count`-vs-line-counter cap precedence, end-of-stream `thinkingComplete` safety net all match the original byte loop. The #487 `thinking`-drop bug is not regressed — `processThinking` still routes `message.thinking` and top-level `thinking` before `processContent`, and the auto-detected markers from `OllamaModelProbe` thread through via the new `autoDetectedThinkingMarkers` field.
3. **Tool calling (#640 / v0.11.2 contract)** — `OllamaPayloadHandler` and `OllamaMessageEncoder.decodeToolCall` preserve both wire shapes (nested `function` envelope + flat `{id, name, arguments}`); `applyToolChoice` handles all four `ToolChoice` cases identically. `processToolCalls` still emits the start + arguments-delta + toolCall triple per #783 and respects `Task.isCancelled` between events.
4. **Concurrency** — `OllamaStreamProcessor` is a struct mutated only inside `parse`; helpers are local funcs over the same captured state. No new actor crossings; `OllamaBackend` keeps `@unchecked Sendable` with `withStateLock` guarding `_manifest` / `_autoDetectedThinkingMarkers` / `toolAwareHistory`.
5. **No `try?` in error paths** — `OllamaModelProbe` and `OllamaPayloadHandler` use explicit `do/catch` with `Log.network.info` / `Log.inference.warning`; only one `try?` reference in the whole decomp, and it's inside a doc comment. SilentCatchAuditTest is happy.
6. **Tests** — `OllamaPayloadHandlerTests` + the relevant `OllamaBackendTests` cases were updated to call the new top-level types (`OllamaPayloadHandler()`, `OllamaPayloadParser.parseLine`, `OllamaPayloadParser.extractToken/extractThinking`). Coverage of the per-line classification path is preserved.

### Non-blocking observations

- **New `await Task.yield()` + `try await Task.sleep(for: .milliseconds(1))` in `processToolCalls`** between `toolCallStart` and the rest of the triple. Not present on `main`. No test forces this and it adds 1ms+ to every tool call. If it's needed for downstream subscribers to observe the `.toolCallStart` before `.toolCall` arrives, fine — but worth a comment explaining intent or removing if it slipped in. Memory note `feedback_task_yield_fragility.md` is exactly this kind of timing-dependent code; flagging not blocking.
- **`OllamaBackend.parseResponseStream` now always allocates an `OllamaStreamProcessor`** before delegating; the cost is a single struct init per stream, negligible. Fine.

Once CI goes green on the security-audit fix, this is good to merge.

Issue #587, opened 2026-04-20
audit: ContextWindowManager reservation ignores maxThinkingTokens


roryford:
## Summary

`ContextWindowManager.trimMessages` uses a caller-supplied `responseBuffer` to reserve context for the model's response. No caller folds `GenerationConfig.maxThinkingTokens` into that reservation, so reasoning models with long chain-of-thought can consume KV slots past the trim budget.

Full audit memo lives in the DocC block at the top of `Sources/BaseChatInference/Services/ContextWindowManager.swift` (PR #TBD).

## Gaps to close

- [ ] `BaseChatInference.GenerationCoordinator.exactPreflightAndTrim` reserves `maxOutput` alone — fold in `config.maxThinkingTokens` (substitute a backend default when `nil`).
- [ ] `BaseChatUI.GenerationCoordinator` passes hardcoded `responseBuffer: 512` — derive from `maxOutputTokens + (maxThinkingTokens ?? 0)` instead.
- [ ] `BackendCapabilities` has no `defaultMaxThinkingTokens` — publish one so the coordinator can substitute when the host passes `nil`.
- [ ] `OllamaBackend.buildRequest` should send `think: false` and drop the thinking component from `num_predict` when `maxThinkingTokens == 0` (P4).

## Not a ship-blocker

Current code over-reserves via the hardcoded `512`, so the observable symptom today is "reasoning models lose a few turns of history sooner than they need to" rather than a crash.

## Related

- P4 (Ollama `maxThinkingTokens` public-semantics change) should close items 1 and 4.

PR #1165, opened 2026-05-09
refactor: decompose GenerationQueue

roryford:
Part of #1113.

Summary:
- Extracts GenerationHistoryInstaller, GenerationPreflightTrimmer, and GenerationToolDispatchLoop.
- Keeps GenerationQueue focused on generate/enqueue, queue drain, thermal pause, cancellation, discard, and stop coordination.
- Preserves structured-history, preflight/trim, and tool-loop behavior.

LOC:
- GenerationQueue.swift: 1,150 -> 756.

Validation:
- scripts/test.sh --filter ManifoldInferenceTests --disable-default-traits --skip-update
- scripts/test.sh --filter ManifoldInferenceSwiftTestingTests --disable-default-traits --skip-update
- git diff --check
- Code review pass: no queue/drain/cancel/tool-loop/preflight regressions found.

### DIFF
--- a/Sources/ManifoldInference/Services/GenerationHistoryInstaller.swift
+++ b/Sources/ManifoldInference/Services/GenerationHistoryInstaller.swift
@@ -1,0 +1,41 @@
+import Foundation
+
+/// Installs structured conversation history on backends that opt in, and
+/// provides the legacy flattened history shape for text-only consumers.
+enum GenerationHistoryInstaller {
+    /// Flattens ``StructuredMessage`` to the legacy `(role, content)` shape
+    /// for backends and helpers that operate on plain strings.
+    ///
+    /// Thinking parts are dropped because they would either bloat the prompt
+    /// with provider-internal reasoning or fail validation on the
+    /// non-Anthropic providers that don't accept replayed thinking.
+    /// ``StructuredHistoryReceiver`` adopters read the unflattened form
+    /// directly to preserve thinking signatures.
+    static func flatten(_ messages: [StructuredMessage]) -> [(role: String, content: String)] {
+        messages.map { (role: $0.role, content: $0.textContent) }
+    }
+
+    static func containsImages(_ messages: [StructuredMessage]) -> Bool {
+        messages.contains { message in
+            message.parts.contains { part in
+                if case .image = part { return true }
+                return false
+            }
+        }
+    }
+
+    /// Hands history to whichever receiver protocol the backend conforms to.
+    ///
+    /// A backend may conform to both — ``StructuredHistoryReceiver`` is set
+    /// first so a backend that needs structured access (Anthropic) gets the
+    /// authoritative shape, and the flattened ``ConversationHistoryReceiver``
+    /// fallback is set afterwards for backends that only inspect strings.
+    static func installHistory(on backend: InferenceBackend, structuredMessages: [StructuredMessage]) {
+        if let structuredReceiver = backend as? StructuredHistoryReceiver {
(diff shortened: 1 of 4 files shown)


roryford (commented):
Reviewed the decomposition end-to-end. Verdict: clean extraction, behavior-preserving, no fixes pushed.

**Checks performed:**

- **Public surface unchanged.** `git diff` on `GenerationQueue.swift` shows zero `public`/`open` additions or removals — all changes are internal call-site updates and method extraction. `GenerationQueue` remains `internal final class @MainActor @Observable` with the same init surface and methods.
- **Token discipline preserved.** `GenerationRequestToken` plumbing (`request.token`, `nextGenerationToken`, `continuations[token]`) is untouched. `LoadRequestToken` lives in `ModelLifecycleCoordinator` and is unrelated to this PR. The new `yieldEvent` closure correctly bounces through `continuations[request.token]?.yield(event)` per turn.
- **Preflight math byte-identical.** `GenerationPreflightTrimmer.exactPreflightAndTrim` reproduces the same `visibleReserve = config.maxOutputTokens ?? 2048`, `thinkingReserve = config.maxThinkingTokens ?? 0`, `promptTokens + maxOutput <= contextSize` predicate, the 20-attempt cap, the system-only and last-user-turn guards. Issue #587's `?? 0` thinking-reserve semantics are preserved verbatim.
- **Tool dispatch loop semantics unchanged.** `maxToolIterations` cap (with `max(1, …)`) still gates the while loop and emits `.toolLoopLimitReached`. Termination conditions match: `dispatchedInThisTurn.isEmpty` → return; `result.errorKind == .cancelled` → return; byte budget exceeded → synthesize `.permanent` and return. Repeat-call signature short-circuit is identical (same warning text, same "tool already called this turn" content). Lifecycle event order (`.toolCall` → `.toolDispatchStarted` → dispatch → `.toolResult` → `.toolDispatchCompleted`) is preserved exactly.
- **History install correct.** `GenerationHistoryInstaller.installHistory` calls both `StructuredHistoryReceiver` (first, authoritative) and `ConversationHistoryReceiver` (fallback) — same order as the original. `flatten` still maps every message regardless of role (system/assistant/user/tool all pass through). No role-based silent drop.
- **Concurrency clean.** `GenerationToolDispatchLoop` is `@MainActor`, called from `@MainActor GenerationQueue.runToolDispatchLoop`. No `Task.detached`, no Task spawning at all in the extracted helpers. The `[weak self]` captures in the four injected closures are appropriate for the queue's ownership of the loop value, and `generateWithConfig` correctly throws `inferenceFailure` if `self` deallocates mid-run rather than silently dropping work.
- **No `try?`** in the new files (SilentCatchAuditTest safe).
- **Log messages preserved.** Even the `"GenerationQueue:"` prefix is kept on the warning emitted by the extracted loop, so log-scraping consumers won't see a string-prefix change.

**Test gap assessment.**

Acceptable. The existing suites — `GenerationQueueToolLoopTests`, `GenerationQueueStructuredHistoryTests`, `GenerationQueueApprovalTests`, `ToolCancellationContractTests`, `KVReuseToolLoopTests` — exercise these code paths through `GenerationQueue`'s public-internal seams (enqueue/cancel/stream events). A pure refactor that preserves behavior at those seams is correctly validated by re-running existing coverage (CI green will confirm). Adding helper-level unit tests now would mostly duplicate the integration coverage. Worth opening a follow-up issue if/when these helpers grow new behavior beyond extracted logic.

**CI:** lint, fuzz, security-audits, server-tests are green. test/macros-tests/cold-start jobs still pending — recommend a final approval pass once those report.

PR #1250, opened 2026-05-15
fix: add curated model checksum audit test and unverified UI indicator

roryford:
## Summary

- Adds `CuratedModelChecksumAuditTest` to `ManifoldInferenceTests` to enforce SHA-256 integrity hygiene on curated GGUF downloads (closes SEC-14)
- Confirms the existing `DownloadableModelRow` "Unverified source" label is wired correctly through `CuratedModel.expectedSHA256 → DownloadableModel.expectedSHA256`

## Advisory vs. hard failure

The audit test has two tiers:

**Advisory (soft):** `test_allCuratedGGUFModelsHaveExpectedSHA256` prints a `WARNING:` line listing every GGUF entry whose `expectedSHA256` is `nil`. It does **not** call `XCTFail` — the list today is empty because `CuratedModel.all` is empty by default (apps populate it at startup). The comment in the test explains how to promote this to a hard failure once all LFS pointer hashes are populated.

**Hard gate (immediate XCTFail):** `test_populatedChecksums_areValid64CharHex` rejects any non-nil checksum that is not exactly 64 lowercase hex characters. This catches common mistakes: empty strings, MD5-length hashes, and checksums with non-hex characters. Two sabotage tests (`test_populatedChecksums_rejectsEmptyString`, `test_populatedChecksums_rejectsTooShortHex`) confirm the gate fires on known-bad values.

**Scope gate:** `test_nonGGUFModels_areNotAuditedForChecksums` verifies MLX entries are excluded so the GGUF-only audit doesn't produce false positives for MLX snapshot repos (which use a different integrity mechanism).

## How to harden further (after populating LFS hashes)

When all curated GGUF entries have `expectedSHA256` set, change the advisory test body from `print("WARNING: ...")` to `XCTFail(...)` and delete the comment. No other changes needed — the hard gate already enforces format once values are present.

## UI indicator

`DownloadableModelRow` already surfaces an "Unverified source" label when `model.expectedSHA256 == nil` (added in a prior pass — see `Sources/ManifoldUIModelManagement/Views/Models/DownloadableModelRow.swift:95`). This PR verifies the data path rather than adding new UI.

## Test plan

- [x] `scripts/test.sh --filter ManifoldInferenceTests --disable-default-traits --skip-update` — 1497 tests, 0 failures; advisory `WARNING` printed for the `example/unverified-GGUF` fixture
- [x] `scripts/test.sh --filter ManifoldInferenceSwiftTestingTests --disable-default-traits --skip-update` — passed
- [x] `swift build --build-tests --disable-default-traits` — Build complete

### DIFF
--- a/Tests/ManifoldInferenceTests/CuratedModelChecksumAuditTest.swift
+++ b/Tests/ManifoldInferenceTests/CuratedModelChecksumAuditTest.swift
@@ -1,0 +1,190 @@
+import XCTest
+@testable import ManifoldInference
+
+/// Audits the curated model list for SHA-256 integrity metadata.
+///
+/// Advisory tier (warning only):
+///   GGUF entries with `expectedSHA256 == nil` are enumerated and printed.
+///   This is intentionally a soft failure until all LFS pointer hashes
+///   are populated — see SEC-14.
+///
+/// Hard gate (XCTFail):
+///   Any populated `expectedSHA256` must be a 64-character lowercase hex
+///   string. An empty string or wrong-length value is silently useless and
+///   would pass validation while providing false confidence.
+final class CuratedModelChecksumAuditTest: XCTestCase {
+
+    // MARK: - Fixtures
+
+    private var previousAll: [CuratedModel] = []
+
+    override func setUp() {
+        super.setUp()
+        previousAll = CuratedModel.all
+        // Populate a representative set that covers all the paths the audit
+        // checks: a verified entry, an unverified entry, and (in the hard-gate
+        // test) a deliberately malformed entry.
+        CuratedModel.all = [
+            CuratedModel(
+                id: "audit-verified",
+                displayName: "Verified Model",
+                fileName: "verified.gguf",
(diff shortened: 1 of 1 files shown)


roryford:
Reviewed SEC-14 checksum audit test (48347e1).

All 5 checks pass:
- `CuratedModel.all` is the correct symbol (OSAllocatedUnfairLock-backed static var, publicly settable)
- `DownloadableModel.expectedSHA256` and `modelType` both exist with the right types
- Advisory test emits the expected `WARNING:` print in test output
- Sabotage tests are **not vacuously passing** — `setUp()` injects a synthetic fixture so the tests never depend on production data having non-nil checksums; the two hard-gate sabotage tests (`rejectsEmptyString`, `rejectsTooShortHex`) construct their own synthetic bad entries
- File is registered in `ManifoldInferenceTests` and all 5 tests ran and passed locally (1543 total, 0 failures)

No changes needed. LGTM.

PR #1453, opened 2026-05-24
fix(runtime): make session restore reliable after configure (#1447)

### LINKED ISSUE #1447: fix(ui): make SwiftUI session restore reliable after SessionManagerViewModel.configure
## Summary

`SessionManagerViewModel.configure(...)` does not make persisted sessions reliably available early enough for a fresh SwiftUI host to restore the prior conversation on relaunch. Across the 2026-05-24 `02-swiftui-chat` DX walkthrough pass, both working runs had to invent their own startup wait/restore heuristics, and the third run's recovered log never observed restored sessions at all.

## Reproduction signal

Autonomous SwiftUI walkthrough pass:
- `scripts/dx-walkthrough/runs/2026-05-24_v0.33.0_commence-2026-05-24/02-swiftui-chat/SUMMARY.md`

Per-run signal:

### run-1 (Foundation Models) — working, but only with restore heuristic

> Immediately after `configure(bootstrap:)`, `sessionVM.sessions` was still empty on relaunch even though the persistent store already contained a chat. My startup path created/selected an empty session before the persisted one arrived, so the app appeared to "lose" the previous conversation unless I added my own wait/restore heuristic.

Runner workaround:
- short polling wait for sessions to materialize
- then switch through candidates until `chatViewModel.messages` repopulated

### run-2 (local GGUF) — working, but only with restore heuristic

> In practice I had to add a small wait/retry before trusting `sessionManager.sessions`, and then explicitly scan sessions for one with existing messages to make relaunch land on a persisted conversation instead of a fresh empty chat.

### run-3 (Ollama) — partial, recovered log shows restore failure

Recovered `session.log`:

```text
[Run3SwiftUIChat] session-count=0
[Run3SwiftUIChat] active-session=EE8C5507-73A9-486C-9B36-1D1D46993B43
[Run3SwiftUIChat] restored-message-count=0
...
=== Relaunch persistence check ===
[Run3SwiftUIChat] session-count=0
[Run3SwiftUIChat] active-session=C99D1C88-BC11-4F1F-B104-6E365660D0A8
[Run3SwiftUIChat] restored-message-count=0
```

## Why this matters

The public docs imply that once the bootstrap and session manager are configured, a multi-session SwiftUI host can rely on persisted sessions being ready for selection on relaunch. In practice, the adopter has to discover one of these undocumented strategies:

- poll `sessionManager.sessions` until something appears
- explicitly prefer a non-empty session over the initial empty one
- re-select sessions until `chatViewModel.messages` catches up

That is too much bespoke startup logic for the documented happy path.

## Scope / hypotheses

This could be one of several shapes:

1. `SessionManagerViewModel.configure(...)` returns before its initial session load has materially populated `sessions`
2. the chat/session view models are not synchronized tightly enough at startup, so selecting a session too early still shows an empty transcript
3. docs over-promise "turnkey restore" for a flow that actually requires an explicit readiness or callback boundary

## Possible fixes

Any of these would help:

1. **Make initial load readiness explicit**
   - e.g. `await sessionManager.configure(...)` only returns after the first session list load settles, or expose a documented readiness property / async hook
2. **Document the startup contract clearly**
   - if the current behavior is intentional, say so in the SwiftUI docs and show the required restore pattern
3. **Strengthen default restore behavior**
   - for the common multi-session host path, prefer the most recent non-empty session once the first load lands

## Severity

**Blocker / major DX issue** for the SwiftUI persistence acceptance criterion. Two runs only succeeded after custom heuristics; the third failed the relaunch check entirely.

## Related

- issue #1430 is adjacent but different: that tracks the documented first-token / empty-state front door for SwiftUI, especially model loading and empty-state affordances
- this issue is specifically about **restoring persisted sessions/conversations on relaunch** once the app is already wired and generating

roryford:
Closes #1447

## Root cause

`SessionManagerViewModel.configure(bootstrap:)` calls `configure(persistence:autoLoad:true,...)`, which schedules `Task { [weak self] in await self?.loadSessions() }` as a fire-and-forget background task and returns immediately. Three independent DX agents hit the same symptom: `sessionVM.sessions` was still empty after `configure` returned, because the initial SwiftData fetch hadn't run yet. They were forced to add polling loops to recover — a clear signal the API was misleading.

## Fix

Added a new `async` method `configureAndLoad(bootstrap:)` on `SessionManagerViewModel` that configures with `autoLoad: false` and then explicitly `await`s `loadSessions()` before returning. Callers that need `sessions` populated synchronously (scene restore, launch-time session selection) use this new method. Callers that rely on `SessionListView`'s `.task {}` modifier can continue using the synchronous `configure(bootstrap:)` overload.

The synchronous `configure(bootstrap:)` doc comment now prominently warns about the fire-and-forget behaviour and points to the new method.

## Tests added (`RuntimeConfigurationTests`)

- `test_configureAndLoad_sessionsPopulatedBeforeReturn` — pre-populates two sessions, calls `configureAndLoad`, asserts `sessions.count == 2` immediately with no polling (regression for #1447)
- `test_configure_autoLoadEventuallyPopulateSessions` — documents and locks the synchronous overload's fire-and-forget contract: `sessions` is empty immediately, populated after `autoLoadTask` drains

🤖 Generated with Claude Code

### DIFF
--- a/Sources/ManifoldUI/ViewModels/RuntimeConfiguration.swift
+++ b/Sources/ManifoldUI/ViewModels/RuntimeConfiguration.swift
@@ -51,12 +51,53 @@ extension SessionManagerViewModel {
     /// a ``ChatRuntimeBootstrap``. Schedules the initial session load so the UI
     /// matches pre-Phase-1.0 behavior. Direct callers of
     /// ``configure(persistence:autoLoad:diagnostics:)`` must opt in explicitly.
+    ///
+    /// > Important: This overload is **fire-and-forget** — it schedules the
+    /// > initial session fetch as a background `Task` and returns before the
+    /// > fetch completes. ``sessions`` may still be empty immediately after
+    /// > this call returns. Use ``configureAndLoad(bootstrap:)`` when you need
+    /// > ``sessions`` populated before proceeding (e.g. session-restore on
+    /// > launch, scene restoration, or any context where you cannot rely on
+    /// > `SessionListView`'s `.task {}` modifier to trigger the load).
     public func configure(bootstrap: any ChatRuntimeBootstrap) {
         configure(
             persistence: bootstrap.persistenceStores,
             autoLoad: true,
             diagnostics: bootstrap.diagnosticsService
         )
+    }
+
+    /// Bootstrap path that **awaits** the initial session fetch before
+    /// returning.
+    ///
+    /// Identical to ``configure(bootstrap:)`` except that it `await`s
+    /// ``loadSessions()`` so ``sessions`` is populated by the time this method
+    /// returns. Use this from `App.body` launch sequences and scene-restoration
+    /// handlers that need to select or restore a session immediately — for
+    /// example:
+    ///
+    /// ```swift
+    /// .task {
+    ///     await sessionManager.configureAndLoad(bootstrap: runtime)
+    ///     // sessions is now populated; safe to restore active session
+    ///     if let id = restoredSessionID {
+    ///         sessionManager.activeSession = sessionManager.sessions
+    ///             .first { $0.id == id }
+    ///     }
+    /// }
+    /// ```
+    ///
+    /// Callers that use `SessionListView` and don't need to inspect
+    /// ``sessions`` immediately after configure can continue to use the
+    /// synchronous ``configure(bootstrap:)`` overload — `SessionListView`'s
+    /// `.task {}` modifier will trigger the load at the right time.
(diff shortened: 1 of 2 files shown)


roryford:
Closing — #1447 was already resolved in e68d0572 (feat(ui): expose SessionManagerViewModel from quickStart() and await initial session load) which landed in main before this PR was reviewed. The core fix (configureAndLoad async overload) ships in that commit.

PR #1454, opened 2026-05-24
fix(mcp): STDIO opt-in, auth required by default, sanitizer hardening (#1413)

### LINKED ISSUE #1413: security(mcp): pre-1.0 hardening — disable STDIO default, enforce OAuth 2.1, sanitize tool metadata
## Why

The MCP ecosystem disclosed a wave of critical vulnerabilities in April–May 2026. Shipping `ManifoldMCP` as a 1.0 surface without tightening defaults exposes consumers and their users.

### Evidence

- **STDIO transport RCE** — 200K+ servers affected, 4 exploit families. [OX Security disclosure](https://www.ox.security/blog/mcp-supply-chain-advisory-rce-vulnerabilities-across-the-ai-ecosystem/), independently confirmed by Microsoft.
- **CVE-2026-35394** — mobile MCP RCE (CVSS critical). [SentinelOne](https://www.sentinelone.com/vulnerability-database/cve-2026-35394/).
- **nginx-ui CVE-2026-33032** — CVSS 9.8 auth bypass in popular MCP server.
- **OAuth 2.1 capability gap** — 41% of public MCP servers ship zero auth despite the [spec requiring it](https://modelcontextprotocol.io/specification/draft/basic/authorization/).
- **Prompt injection via tool metadata: 85% success rate** ([ArXiv 2601.17548](https://arxiv.org/abs/2601.17548)). No general mitigation exists yet.
- **30+ CVEs across MCP ecosystem** in recent months; critical in `mcp-server-git`, `aws-mcp`, others.

Independent verdict on the hardening effort: **3–4 weeks of infrastructure work, not a 1-week feature.**

## What

Current state — already better than most: OAuth 2.0 + PKCE, refresh tokens in stricter Keychain class, SSRF hardening, redirect caps, 8 KB content sanitizer. Gaps to close before 1.0:

- [ ] **STDIO transport disabled by default.** Require explicit opt-in flag with a documented threat model. Default to HTTP+SSE with OAuth.
- [ ] **OAuth 2.1 required on every server descriptor.** Reject server registration without auth metadata; permit unauthenticated only via explicit `.unsafe` flag (loud naming).
- [ ] **Tool metadata sanitizer.** Treat `description`, `inputSchema.description`, parameter names as untrusted. Cap lengths, strip ANSI / control chars, log if a tool description contains imperative-mood instructions ("ignore previous", "system:", etc.).
- [ ] **Process isolation guidance.** Document — and where possible enforce — that the MCP client process is not the same process as any MCP server execution.
- [ ] **`MCPContentSanitizer` audit.** Fix the grapheme-count fallback gap flagged in the v0.33 audit.
- [ ] Update `SECURITY.md` with explicit MCP threat model and the four control points above.

## Acceptance

- [ ] Default-trait build of an MCP client cannot connect to a STDIO server without an explicit unsafe flag
- [ ] Unauthenticated server descriptor fails registration with a clear error
- [ ] Tool-metadata fuzz test exercising injection payloads passes
- [ ] SECURITY.md updated and linked from README MCP section

roryford:
Closes #1413 (partial — STDIO gate, auth requirement, sanitizer hardening, SECURITY.md)

## Changes

- **`MCPServerDescriptor.allowsSTDIOTransport`** (default `false`) — connecting to a `stdio` endpoint without setting this flag throws a clear error citing `SECURITY.md`. STDIO spawns a subprocess with the app's full process privileges, bypasses TLS and the SSRF guard, and has no revocation path — it must be an explicit opt-in.

- **`MCPServerDescriptor.isUnauthenticatedUnsafe`** (default `false`) — connecting to a server with `authorization: .none` without setting this flag throws a descriptive error. Unauthenticated servers have no cryptographic identity; tool call arguments travel in the clear.

- **`MCPContentSanitizer.logInjectionIndicators`** — new internal function called from `MCPToolSource` for every tool name and description. Logs a `Log.inference.warning` when known prompt-injection phrases (`"ignore previous"`, `"system:"`, `"override"`, `"disregard"`, `"STOP"`) are detected. Logs only — does not strip. Stripping silently would hide the attack from operators.

- **`MCPToolSource` grapheme fallback fix** — the `String(bytes:encoding:)` fallback after UTF-8 byte-cap truncation previously fell back to the uncapped raw string on decode failure. Fixed to fall back to `String.prefix(N)` by character (grapheme cluster) instead, closing the length-cap inconsistency.

- **`SECURITY.md`** — new `## MCP Threat Model` section covering STDIO opt-in rationale, auth requirement, metadata sanitisation behaviour, and process isolation guidance for host apps.

- **Tests** — three new tests in `MCPHardeningTests`: `test_stdioTransport_rejectedByDefault`, `test_unauthenticatedDescriptor_rejectedByDefault`, `test_sanitizer_logsInjectionIndicator`. Updated two existing tests (`test_connectRejectsSSRFBlockedTransportEndpoint`, `test_connectStdioUsesPlatformGating`) that call `MCPClient.connect()` with default `.none` auth to pass `isUnauthenticatedUnsafe: true` so they reach the assertion they were designed to test.

## Acceptance criteria

- [x] Default-trait build cannot connect to STDIO server without `allowsSTDIOTransport = true`
- [x] Unauthenticated descriptor fails registration without `isUnauthenticatedUnsafe = true`
- [x] Tool metadata with injection phrases logs a warning (log-only, not blocked)
- [x] SECURITY.md updated with MCP Threat Model section
- [x] 201 MCP tests passing, 0 failures

🤖 Generated with Claude Code

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -231,6 +231,76 @@ For deployments that need stricter at-rest sealing, set
 `ManifoldConfiguration.shared.fileProtectionClass = .complete` and ship background
 work that is robust to a locked device.
 
+## MCP Threat Model
+
+ManifoldKit's MCP client (`MCPClient`) connects to external tool servers. Each server is
+a trust boundary: a compromised or malicious server can attempt prompt injection, data
+exfiltration, and SSRF. This section documents the mitigations and the residual risks
+that host apps must address.
+
+### STDIO opt-in requirement
+
+STDIO transport spawns a local subprocess with the app's process privileges. Unlike
+HTTP, there is no TLS, no SSRF guard, and no revocation path — the transport is as
+trusted as the executable on disk. `MCPServerDescriptor.allowsSTDIOTransport` defaults
+to `false`; connecting to a `stdio` endpoint without setting it throws:
+
+```
+MCPError.transportFailure("STDIO transport requires explicit opt-in via
+MCPServerDescriptor.allowsSTDIOTransport. See SECURITY.md for the threat model.")
+```
+
+Before setting `allowsSTDIOTransport = true`, verify:
+
+- The executable is code-signed with a known requirement (set `codesignRequirement`
+  on `MCPStdioCommand` to enforce this at runtime on macOS).
+- The executable path cannot be replaced by a less-privileged user or an adversarial
+  package install script.
+- The working directory and environment passed to the subprocess do not contain
+  secret credentials.
+
+### Auth requirement
+
+`MCPServerDescriptor.isUnauthenticatedUnsafe` defaults to `false`. Connecting to a
+server with `authorization: .none` without setting this flag throws:
+
+```
+MCPError.transportFailure("MCP server has no auth configuration. Set
+isUnauthenticatedUnsafe: true to permit unauthenticated connections.")
+```
+
+Unauthenticated servers have no cryptographic identity. Tool call arguments and return
+values are sent in the clear. This is acceptable for loopback-only servers (e.g., a
+locally-launched STDIO process that is also the only user of the port), but is a
+significant risk for any network-reachable endpoint.
+
+### Metadata sanitization
+
+`MCPContentSanitizer` wraps all tool output in an untrusted-content envelope and strips
+ANSI/DEC terminal escape sequences, control characters, and envelope-escape injection
+attempts before the content reaches the model's context window.
+
+`MCPToolSource` caps tool names and descriptions by UTF-8 byte count. When content
+contains known prompt-injection indicator phrases (`"ignore previous"`, `"system:"`,
+`"override"`, `"disregard"`, `"STOP"`), a warning is written to `Log.inference` instead
+of silently dropping the content — stripping silently would hide the attack from
+operators. Host apps should forward `os_log` output from the `com.manifoldkit.inference`
+subsystem to their observability pipeline.
+
+### Process isolation guidance
+
+ManifoldKit does not sandbox MCP server processes. Process isolation is the host app's
+responsibility:
+
+- On macOS, launch STDIO server processes inside an `NSXPCConnection` with a restricted
+  sandbox profile or use App Sandbox entitlements on the server binary.
+- On iOS/iPadOS, STDIO is unavailable. Use `streamableHTTP` against a loopback server
+  launched via a macOS companion app or an on-device HTTP server library.
+- Restrict the server process's file-system access to the minimum needed. Do not pass
+  the app's home directory as the working directory.
+- Do not forward `HOME`, `PATH`, or other ambient credentials from the parent
+  environment unless the server explicitly requires them.
+
 ## Pending Mitigations
 
 The following are known gaps with tracking issues. Each is listed in
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -282,10 +282,18 @@ attempts before the content reaches the model's context window.
 
 `MCPToolSource` caps tool names and descriptions by UTF-8 byte count. When content
 contains known prompt-injection indicator phrases (`"ignore previous"`, `"system:"`,
-`"override"`, `"disregard"`, `"STOP"`), a warning is written to `Log.inference` instead
+`"override"`, `"disregard"`, `"[STOP]"`), a warning is written to `Log.inference` instead
 of silently dropping the content — stripping silently would hide the attack from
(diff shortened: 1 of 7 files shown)


roryford:
Review pass: one fix applied, otherwise the PR is solid.

**Finding fixed: injection indicator test was a no-op**

`test_sanitizer_logsInjectionIndicator` only verified "no crash" — the sabotage check in the comment explicitly acknowledged that commenting out the detection loop would leave the test green. This is a real gap: the test didn't demonstrate that the detection code path actually ran.

**Fix:** Made `logInjectionIndicators` return `@discardableResult Bool` (whether any indicator was detected). Production call sites in `MCPToolSource` discard it unchanged. The test now uses `XCTAssertTrue(detected, ...)` for each injection phrase and `XCTAssertFalse(clean, ...)` for the benign phrase. Commenting out the indicator-matching loop now makes every assertion fail. Amend-committed and force-pushed.

**Everything else looks good:**
- STDIO check fires before transport setup (line 76 in `MCPClient.connect`, before `makeTransport` at line 102) ✓
- Auth check reads `descriptor.authorization` (the stored field declaring what auth type is required) not the runtime `authorization` parameter — correct, fires at connect time before session starts ✓
- Both checks yield `.failed` to the state stream and throw `MCPError.transportFailure` with descriptive messages ✓
- Injection logging uses `Log.inference.warning` with correct privacy levels (`.public` for field, toolName, indicator — appropriate since these are server-controlled strings that operators need to see) ✓
- Grapheme-count fallback fix is correct: `?? String(rawName.prefix(config.maxMCPToolNameBytes))` uses grapheme-cluster prefix instead of the original uncapped `rawName` ✓
- No `try?` introduced in Sources/ManifoldMCP/ (the one in `AuthorizationServerDiscovery.swift` is pre-existing) ✓
- No `assertionFailure` for recoverable errors ✓
- `allowsSTDIOTransport` and `isUnauthenticatedUnsafe` are `public var` not `public let` — fine, Codable round-trips preserve the values ✓
- Existing tests updated correctly: `test_connectRejectsSSRFBlockedTransportEndpoint` and `test_connectStdioUsesPlatformGating` both set `isUnauthenticatedUnsafe: true` so they reach their intended failure modes rather than hitting the new early-exit guard ✓
- No `MockURLProtocol.reset()` added (tearDown call is pre-existing) ✓
- Build clean, 201 MCP tests pass including the three new ones ✓

roryford:
Second review pass: 4 issues found and fixed in commit 49bc68f5.

**Security gap: parameter descriptions not scanned for injection indicators**

`MCPContentSanitizer.logInjectionIndicators` was called on tool name and top-level description, but not on per-parameter `description` fields inside the JSON Schema. Parameter descriptions appear verbatim in the model's context window — an adversarial server that hides injection content in parameter metadata rather than the top-level description bypassed the detection entirely.

Fixed by adding `MCPContentSanitizer.logInjectionIndicatorsInSchema` (recursive walk over all `description` string values in any `JSONSchemaValue` tree) and wiring it into `MCPToolSource.parseToolsListResponse` after `inputSchema` is parsed.

**False-positive problem: "STOP" matched bare "stop" after lowercasing**

The indicator list had `"STOP"` and the comparison applied `.lowercased()` to both sides, meaning any tool description containing the word "stop" — including `"Stop the running process"` or `"non-stop streaming"` — would log an injection warning. This would flood operator logs with noise, making the signal useless.

Changed to `"[STOP]"` which targets the bracketed-token form used in adversarial payloads (e.g. `[STOP][IGNORE ABOVE]`) rather than the common English verb. SECURITY.md updated with a note about the false-positive tradeoff.

**Missing positive-case tests**

- No test confirmed `isUnauthenticatedUnsafe: true` actually passes the auth guard (the rejection test only verified the guard fires; nothing verified the opt-in path continues). Added `test_unauthenticatedDescriptor_withOptIn_passesAuthGuard`: sets `isUnauthenticatedUnsafe: true`, expects an `ssrfBlocked` error (not `transportFailure`), proving the auth guard was bypassed successfully.
- Added `test_sanitizer_logsInjectionIndicatorInParameterDescription` to exercise the new schema scan directly.
- Added a false-positive regression assertion: bare "stop the process" must not trigger the indicator.

**Test suite confirmed**: 203 tests, 0 failures (`swift test --traits MCP --filter ManifoldMCPTests`).

PR #1455, opened 2026-05-24
feat(observability): per-token latency + cost meter on SSECloudBackend (#1414)

### LINKED ISSUE #1414: feat(observability): per-token latency + cost meter on SSECloudBackend
## Why

Cloud-backed chat SDKs that ship without cost/latency instrumentation get blamed for spend surprises. Theory Delta has tracked this as a recurring 2026 production-priority pattern: **cost spirals, runaway iterations, and observability gaps now dominate enterprise constraints** for agent infrastructure — it's an architectural issue, not a model-capability issue.

Concrete cost cliff to expose: **GPT-5.5 charges 2× input pricing above 272K input tokens for the full session** (released 2026-04-23). Consumers cannot react to that without telemetry.

ManifoldKit's reliability primitives are already strong (retries, cert pinning, latest-wins handoff, memory admission). The missing primitive is per-call accounting.

## What

Lightweight, opt-in (default on for `CloudSaaS` trait) instrumentation surface on `SSECloudBackend`:

- [ ] Emit a structured `InferenceMetric` per request with: provider, model, prompt tokens (cached/uncached), completion tokens, TTFT, mean inter-token latency, wall-clock duration, error class if any.
- [ ] Pluggable sink — default to in-memory ring buffer (last N requests), with a protocol consumers can implement to forward to OSLog, MetricKit, or a custom backend.
- [ ] Cost estimator with a small static table per provider/model; mark estimates `approximate` and tie to a documented "as of YYYY-MM-DD" snapshot so consumers know it drifts.
- [ ] Surface a `CurrentSessionMetrics` view-model accessor for the UI to render a "this conversation has cost ~$X" badge.
- [ ] Mock backend emits synthetic metrics so consumer tests don't need a real provider.

## Non-goals

- A general analytics framework. Sinks are pluggable; we ship one default.
- Server-side billing reconciliation. Estimator is a UX safety net, not an invoice.

## Acceptance

- [ ] Unit test: metric emitted for every successful and failed cloud call
- [ ] Cost estimator returns within 10% of actual on a known-prompt fixture for Claude + OpenAI
- [ ] Memory footprint of default ring buffer documented and bounded
- [ ] Example app shows a live "tokens / latency / cost" overlay

## Source

Theory Delta memory `project_agent_loop_failure_modes_production_priority` (2026-05); block `frontier-models-2026.md` / memory `technical_gpt55_release_and_token_cost_cliff` (2026-04-23).

roryford:
Closes #1414

## Changes
- `InferenceMetric` value type + `InferenceMetricSink` protocol in `ManifoldCloudCore`
- `InMemoryMetricSink` actor (ring buffer, default capacity 100, shared singleton)
- `InferenceCostEstimator` with 2026-05-24 rate table for Claude + OpenAI models
- `SSECloudBackend` emits one metric per call via a thin observing wrapper stream; tracks TTFT and mean inter-token latency from the consumer side; defaults to `InMemoryMetricSink.shared`, set to `nil` to disable
- `MockInferenceBackend` emits synthetic zero-latency metrics via `mockMetricSink` property
- `ManifoldTestSupport` gains `ManifoldCloudCore` dependency (no cycle: CloudCore→Inference, TestSupport→CloudCore→Inference)
- 18 tests across `InferenceCostEstimatorTests`, `InMemoryMetricSinkTests`, `SSECloudBackendMetricTests` (trait-gated `Ollama || CloudSaaS`)

## Design notes
- Token timing is measured on the output side of the stream (consumer-visible latency) via a wrapper `AsyncThrowingStream`. This avoids modifying every `parseResponseStream` override in the subclass tree.
- Metric emission runs in a fire-and-forget `Task` after the stream completes, preserving the existing `generate()` return type.
- `GenerationMetricTracker` is `@unchecked Sendable` with `NSLock` — same pattern as `SendableCounter` in the same file.

### DIFF
--- a/Sources/ManifoldCloudCore/InferenceCostEstimator.swift
+++ b/Sources/ManifoldCloudCore/InferenceCostEstimator.swift
@@ -1,0 +1,60 @@
+import Foundation
+
+/// Static cost table and estimator for common cloud inference models.
+///
+/// Rates are per million tokens (input / output) and were captured on
+/// `costTableDate`. When a model is not in the table the estimator returns
+/// a zero cost marked as approximate so callers can distinguish "free"
+/// from "unknown".
+public enum InferenceCostEstimator {
+
+    /// ISO date string of the rate table bundled in this build.
+    public static let costTableDate = "2026-05-24"
+
+    // Per-million token prices: (inputUSD, outputUSD)
+    // Sourced from public pricing pages as of costTableDate.
+    private static let ratesPerMillion: [String: (input: Double, output: Double)] = [
+        "claude-opus-4-7":      (15.00, 75.00),
+        "claude-sonnet-4-6":    (3.00,  15.00),
+        "claude-haiku-4-5":     (0.80,  4.00),
+        "gpt-4o":               (2.50,  10.00),
+        "gpt-4o-mini":          (0.15,  0.60),
+    ]
+
+    /// Estimates the cost of a single inference call.
+    ///
+    /// - Parameters:
+    ///   - provider: Backend name (currently unused for routing; rates are
+    ///     keyed by model identifier alone).
+    ///   - model: Model identifier string. Compared case-insensitively against
+    ///     the rate table; leading/trailing whitespace is stripped.
+    ///   - promptTokens: Number of prompt tokens billed by the provider.
+    ///   - completionTokens: Number of completion tokens billed by the provider.
+    /// - Returns: A tuple with the estimated cost in USD and a flag indicating
+    ///   whether the estimate is approximate (i.e. the model was not in the
+    ///   rate table and the cost is zero rather than accurate).
+    public static func estimatedCost(
+        provider: String,
+        model: String,
+        promptTokens: Int,
+        completionTokens: Int
+    ) -> (usd: Double, isApproximate: Bool) {
+        let key = model.trimmingCharacters(in: .whitespaces).lowercased()
+        // Exact match first; fall back to prefix match for model variants
+        // (e.g. "claude-sonnet-4-6-20261201" → "claude-sonnet-4-6").
+        let rates: (input: Double, output: Double)?
+        if let exact = ratesPerMillion[key] {
+            rates = exact
+        } else {
+            rates = ratesPerMillion.first { key.hasPrefix($0.key) }?.value
+        }
+
+        guard let r = rates else {
+            return (usd: 0, isApproximate: true)
+        }
+
+        let inputCost  = Double(promptTokens)    / 1_000_000 * r.input
+        let outputCost = Double(completionTokens) / 1_000_000 * r.output
+        return (usd: inputCost + outputCost, isApproximate: false)
+    }
+}
--- a/Sources/ManifoldCloudCore/InferenceCostEstimator.swift
+++ b/Sources/ManifoldCloudCore/InferenceCostEstimator.swift
@@ -40,13 +40,18 @@ public enum InferenceCostEstimator {
         completionTokens: Int
     ) -> (usd: Double, isApproximate: Bool) {
         let key = model.trimmingCharacters(in: .whitespaces).lowercased()
-        // Exact match first; fall back to prefix match for model variants
+        // Exact match first; fall back to longest-prefix match for model variants
         // (e.g. "claude-sonnet-4-6-20261201" → "claude-sonnet-4-6").
+        // Longest-prefix wins to avoid "gpt-4o" matching "gpt-4o-mini-20261201"
+        // when both keys are present in the table.
         let rates: (input: Double, output: Double)?
         if let exact = ratesPerMillion[key] {
             rates = exact
         } else {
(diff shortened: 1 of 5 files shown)


roryford:
Review pass: found and fixed one dependency rule violation; all other checklist items passed.

**Fixed: ManifoldTestSupport → ManifoldCloudCore dependency (critical)**

The PR added `import ManifoldCloudCore` to `MockInferenceBackend.swift` and wired `ManifoldCloudCore` into `ManifoldTestSupport`'s Package.swift dependencies. This violates CLAUDE.md: `ManifoldTestSupport` is the shared mock layer imported by every test target including pure-inference tests, and adding `ManifoldCloudCore` would transitively drag cloud networking (SSE, TLS pinning, DNS rebind guard) into all of them.

Fix applied: removed `import ManifoldCloudCore` from `MockInferenceBackend.swift`, removed the `mockMetricSink` property and the two synthetic metric emission blocks, and removed the `ManifoldCloudCore` entry from `ManifoldTestSupport`'s Package.swift target. Tests testing the metric pipeline (the three `SSECloudBackendMetricTests` suites in `InferenceMetricTests.swift`) already work correctly against the real `SSECloudBackend` + `InMemoryMetricSink` and don't need a mock shortcut.

**Passed items:**
- `GenerationMetricTracker`: all mutable state (wallStart, firstTokenInstant, lastTokenInstant, interTokenGapsNs) accessed exclusively under NSLock in all three methods — thread-safe
- TTFT: correctly gated on `case .token` in the wrapper stream relay task
- Mean inter-token latency: correctly accumulated as per-gap intervals (not wall-clock/count) then averaged
- `InMemoryMetricSink.record()`: non-async actor method satisfies the `async` protocol requirement — Swift allows this; callers implicitly await the actor hop
- Cost estimator prefix matching: `claude-sonnet-4-6-20261201` → `claude-sonnet-4-6` tested and passing
- No `try?` in new Sources/ files
- All `InferenceMetric` properties are `public`
- SSE backend integration tests correctly guarded with `#if Ollama || CloudSaaS`
- Ring buffer FIFO eviction test verifies oldest entry is evicted (model == "second" is first after overflow)
- Both Claude and OpenAI models have cost estimator tests

**Build results:** `--disable-default-traits` clean, `--traits CloudSaaS` clean, 19/19 metric tests pass.

roryford:
Second review pass — 4 issues found and fixed in this pass, 1 documented-by-design, build green.

## Bugs fixed

**1. Prefix match non-determinism in `InferenceCostEstimator` (correctness bug)**

`gpt-4o-mini-20261201` has two matching prefixes in the rate table: `"gpt-4o"` ($2.50/M) and `"gpt-4o-mini"` ($0.15/M). The previous `.first { key.hasPrefix($0.key) }` relies on `Dictionary` iteration order, which is non-deterministic across runs. It could silently charge the wrong rate.

Fix: changed to pick the *longest* matching prefix key. Added regression test `gpt4oMiniVariantUsesLongestPrefix` that verifies `gpt-4o-mini-20261201` resolves to the mini rate.

**2. `InferenceMetric` missing `timestamp: Date`**

Without a timestamp, consumers storing metrics in a time-series sink (InfluxDB, Datadog) must snapshot the clock themselves. Added `timestamp: Date` with a default of `Date()` (non-breaking). `GenerationMetricTracker.start()` now captures `Date()` alongside `ContinuousClock` and threads it through `buildMetric` → `InferenceMetric.timestamp`.

**3. Vacuous catch in `emitsMetricOnFailure` test**

The error path silently swallowed the exception with a bare `// Expected` comment. If the 401 handler changed to not throw, the test would still pass (zero assertions before `#require`). Fixed to capture the error and `#expect(caughtError != nil)` before proceeding to check `errorClass`.

**4. `GenerationMetricTracker` nil-self zero-token case undocumented**

When `self` is deallocated mid-stream, `lastUsage` is unavailable and the metric is emitted with zero token counts (cost will be $0 but timing and `errorClass` are intact). Added an inline comment explaining the intentional degradation.

## Documented-by-design (no change)

**TTFT excludes `.thinkingToken` events** — intentional. TTFT measures output latency as perceived by the end user. `.thinkingToken` events are internal reasoning that doesn't reach the UI. Added doc comment to `InferenceMetric.timeToFirstToken` to make this explicit so future contributors don't "fix" it.

## Not changed

- `cachedPromptTokens` is hardcoded to 0 at the `SSECloudBackend` call site — `lastUsage` doesn't carry a cached-token field yet. That's a missing upstream feature in the usage-extraction protocol, not a bug in this PR.
- `classifyError` `default` branch covering future `CloudBackendError` cases — acceptable; the doc comment was overstating ("automatically descriptive"), so I didn't change the code but the comment now says "cloudError" is the fallback for unrecognised cases.

## Build / test status
```
swift build --build-tests --disable-default-traits  ✅ Build complete
swift test --filter "InferenceCostEstimator|InMemoryMetricSink"  ✅ 16/16 passed
```

PR #1525, opened 2026-05-30
feat(runtime): add PreTurnCompressionPolicy for pre-send history compression

### LINKED ISSUE #1519: feat(runtime): pre-turn compression stage for next-turn preparation
## Problem

`CompressionPolicy` currently runs after a successful generation turn, after post-generation hooks, and replaces persisted history before the runtime emits `historyCompressed`.

That is the right default for generic chat, but it does not match Fireside's story invariant. Fireside evaluates and applies compression before appending the next user story node. The memory brief must become part of next-turn preparation so the just-submitted user action remains recent and outside the compressed segment.

A direct mapping to current `CompressionPolicy` would invert that ordering:

```text
Fireside today:
pre-turn compression -> append user node -> append assistant node -> generate -> extraction/observer tail

ConversationRuntime today:
append user message -> generate -> hooks -> post-turn compression
```

For Fireside, this ordering affects active-path semantics, memory-node placement, graph reconciliation, and undo behavior.

## Proposed direction

Add an optional pre-turn compression stage to `ConversationRuntime`, distinct from the existing post-turn `CompressionPolicy`. Possible shapes:

```swift
public protocol PreTurnCompressionPolicy: Sendable {
    func shouldCompressBeforeTurn(request: PreTurnCompressionRequest) async throws -> Bool
    func compressBeforeTurn(
        request: PreTurnCompressionRequest,
        generate: @Sendable ([ChatMessageRecord]) async throws -> String
    ) async throws -> PreTurnCompressionResult
}
```

or extend `CompressionPolicy` with an explicit timing mode:

```swift
public enum CompressionTiming: Sendable {
    case postSuccessfulTurn
    case beforeUserAppend
}
```

The stage should run before the new user record is inserted, and should be able to emit/persist replacement records atomically like the current post-turn path.

## Acceptance criteria

- Host can opt into compression before user-message append without disabling current post-turn compression for existing apps.
- Runtime preserves atomic replacement semantics and emits an event carrying inserted records.
- Compression failures preserve existing history and abort or continue according to a documented policy.
- Generation hooks and history/context providers see a well-defined ordering relative to pre-turn compression.
- Tests cover ordering: old history compressed, new user message remains outside the compressed segment.

## Fireside impact

This is the primary blocker for adopting `ConversationRuntime` for Fireside story turns beyond isolated inference/history spikes. It would let Fireside preserve its memory brief ordering while reusing ManifoldKit's runtime infrastructure.

Related: #1245, #1239, #1236.

roryford:
## Summary

- Adds `PreTurnCompressionPolicy` — a new optional `ConversationRuntime` port that compresses history **before** the user message is appended, so the just-submitted action always falls outside the compressed segment
- Adds `ConversationError.preTurnCompressionFailed(any Error)` — pre-turn failures throw to the caller (unlike post-turn compression which logs and continues), since the host's ordering invariant depends on compression completing first
- Extracts the compression `generate` closure into a shared `makeCompressionGenerateClosure()` helper used by both pre-turn and post-turn paths
- 10 tests covering: policy triggered/skipped, Fireside ordering invariant (user outside compressed segment), event ordering (`historyCompressed` before `messageInserted(user)`), failure throws and preserves history, empty-return guard, `postCompressBeforeTurn` callback, no-op on regenerate, and correct arg forwarding

Closes #1519.

## Design rationale

This is a separate protocol from `CompressionPolicy` rather than a timing mode on the existing one because:
- The decision inputs differ: `shouldCompressBeforeTurn(messageCount:lastPromptTokens:)` vs `shouldCompress(promptTokens:contextSize:contextUtilization:)`
- The error policy differs: pre-turn failures abort the turn (the host's invariant requires it); post-turn failures silently continue
- Both can coexist on the same runtime instance — existing post-turn apps are unaffected

The `userMessage` record is created *after* the compression block so its `Date()` timestamp naturally follows the summary records in store sort order.

## Test plan

- [ ] `swift test --disable-default-traits --filter PreTurnCompressionPolicyTests` — all 10 pass
- [ ] `swift test --disable-default-traits --filter ManifoldRuntimeTests` — no regressions


roryford:
Superseded by #1524 (merged as fc6bb59b), which shipped PreTurnCompressionPolicy, ConversationError.preTurnCompressionFailed, and makeCompressionGenerateClosure() along with HistoryShaper and HostTurnContextProvider. The ManifoldBootstrap wiring from this branch (uncommitted locally) should come as a fresh PR against main.

Issue #1601, opened 2026-06-03
Make backend conformance meta-contract tests --parallel-safe (self-contained per method)


roryford:
## Problem

The per-backend conformance suites (`Tests/ManifoldBackendsTests/Conformance/*`) verify capabilities with a **cross-method accumulation** pattern: each `test_contract_<flag>_claim` method records a claim into a process-static registry (`BackendContractChecks` claims set), and a final `test_z_contract_metaContract` method reads the registry to assert every declared-`true` capability was claimed.

This pattern is **fundamentally incompatible with `swift test --parallel`**, which distributes individual test *methods* across separate worker processes. The claim-recording methods and the `z_meta` reader land in different processes with separate static registries, so the meta-check sees an empty/partial registry and fails.

**Proven:** `ClaudeBackendConformanceTests` passes serially (9 tests, exit 0) but fails under `--parallel` (the meta method reports all flags "unproven"). `MockInferenceBackendConformanceTests` passes either way because it already records-and-asserts within a single method.

This is why CLAUDE.md notes the local profile "deliberately does NOT pass `--parallel`", and why these suites currently run serially (build-modes.yml) / are compiled out of the parallel `test` job (`--traits MCP`).

## Why it matters

It blocks collapsing CI's trait-segregated test invocations into one parallel pass. A stopgap (track via #1597-area work) skips `test_z_contract_metaContract` in the parallel step and runs it in a small serial step; this issue is the durable fix.

## Proposed fix

Redesign the meta-contract to be **self-contained per method** (the `MockInferenceBackend` pattern): each conformance class runs its capability assertions and the meta-check within a single test method (one process), removing the dependency on cross-method/cross-process static state. Affects ~7 conformance classes + the `BackendContractChecks` harness. These are correctness gates — the redesign must preserve exactly what they verify (every declared-`true` tracked flag has a proving assertion; every declared-`false` fails closed).

## Acceptance

- All backend conformance suites pass under `swift test --parallel` with `--traits MCP,CloudSaaS`.
- No reliance on `BackendContractChecks` process-static claim accumulation across methods.
- The serial-metaContract stopgap step can be removed from CI.

roryford:
## Scope is broader than just the meta-contract

[...]

PR #1602, opened 2026-06-03
perf(ci): order test-step traits to do one rebuild in the headroom step (~60s/run)

roryford:
## What

The `test` job runs three `swift test` invocations. `swift test` rebuilds the **whole test tree** whenever the trait set changes (`--filter` only gates execution). The three steps had three different trait sets (`MCP` / none / `CloudSaaS`) → **two rebuilds** per run.

Constraints mean step 1 and step 3 can't share a trait set:
- **Step 1** (parallel XCTest) must stay `MCP`-only — adding CloudSaaS pulls the CloudSaaS backend suites into a `--parallel` pass and destabilises the conformance meta-contract + timing-sensitive tests (**#1601**).
- **Step 3** (PinnedSession) must be `CloudSaaS` and has a tight **120s** stall watchdog.

So exactly **one** rebuild is unavoidable. This PR puts it in **step 2** (Swift Testing), which is isolated and runs under the generous **240s** watchdog — by building `CloudSaaS` there, **step 3 reuses the `.build`** and stays test-execution-only under its 120s budget.

## Impact

- **One rebuild instead of two (~60s/run reclaimed).**
- Removes a latent flake: step 3 previously did its own `→CloudSaaS` rebuild under the tight 120s watchdog (this PR's first revision tripped exactly that — the compile ran past 120s and the watchdog killed it). Moving the rebuild to the 240s step fixes it.

Zero behavior change: `ManifoldInferenceSwiftTestingTests` needs neither trait; CloudSaaS is chosen only so step 3 can reuse the build (verified locally: 83 tests pass under `--traits CloudSaaS`).

## Not the full unification (#1587)

The full collapse onto one parallel pass is blocked on stabilising the CloudSaaS parallel-fragile suites — tracked in **#1601**. This PR is the safe slice.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -325,6 +325,23 @@ jobs:
       # `if: always()` so a failing XCTest step still surfaces Swift Testing
       # failures in the same run — otherwise breakage in both harnesses costs
       # two re-push cycles to triage. The job still fails because step 1 failed.
+      #
+      # Trait ordering across the three test steps minimises rebuilds.
+      # `swift test` rebuilds the whole test tree whenever the trait set changes
+      # (the filter only gates execution), and a rebuild differs per step:
+      #   step 1 (XCTest, parallel) — MUST be `MCP` only. Adding CloudSaaS pulls
+      #     the CloudSaaS backend suites into a parallel pass and destabilises the
+      #     conformance meta-contract + other timing-sensitive tests (see #1601).
+      #   step 3 (PinnedSession)    — MUST be `CloudSaaS`, and has a tight 120 s
+      #     stall watchdog that a full trait-change recompile blows through.
+      # So step 1 and step 3 unavoidably differ by one trait set → one rebuild.
+      # Put that rebuild HERE (step 2): this step is isolated, runs under the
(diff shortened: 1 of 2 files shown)


roryford:
Added a second commit: `fix(test)` bound-waits for the bridged-handle release in `ToolCancellationContractTests.test_bridgedCallbackExecutor_releasesHandle_onCancellation`. That test asserts the handle is released synchronously, but the deinit happens on a child-task unwinding timeline and races the assertion under `--parallel` — it flaked the XCTest step on the previous CI run here (and runs under `--traits MCP`, so it can flake on main too). Bound-wait deflakes it; a genuine leak still fails. 3/3 green locally.

roryford:
Closing in favor of #1603. The trait-reordering perf change here proved not worth the fragility: every variant detonated a different latent issue under CI's hard-won trait segregation — `step2→MCP` tripped the PinnedSession step's 120s compile watchdog, and `step2→CloudSaaS` caused `ManifoldInferenceSwiftTestingTests` to crash with SIGSEGV (7 suites, signal 11) in CI. The genuinely-valuable piece — the `ToolCancellationContractTests` flake fix — is split out into #1603. The trait-unification perf work stays blocked on the parallel-stabilization tracked in #1601.

Issue #1608, opened 2026-06-04
[arch-migration P1a] Extract ManifoldNetworking from the kernel


roryford:
**WWDC-independent — ready to start.** Target-architecture migration, Phase 1 (thin the kernel). Plan: `docs/plans/target-architecture-migration.md` §P1. Docs PR: #1604.

### Context
`ManifoldInference` is ~25.1k LOC but only ~3.4k (14%) is the actual Contract. P1 evicts the non-contract mass so the kernel becomes small and SemVer-stable. This issue extracts the networking infra.

### Scope
- New leaf module **`ManifoldNetworking`** holding `Sources/ManifoldInference/Networking/*` (`URLSessionFactory`, `CompositeURLSessionDelegate`, `RedirectGuardDelegate`, `NetworkPolicyURLProtocol`, `NetworkActivity*`, `PrivateIPClassifier`) **plus** `ManifoldCloudCore`'s SSE/TLS/DNS infra (`SSECloudBackend`, `PinnedSessionDelegate`, `DNSRebindingGuard`, `URLSessionProvider`, transports).
- `@_exported import ManifoldNetworking` shim from `ManifoldInference` so downstream imports don't break.
- Add dep edges from the adapters that need it.

### Test & CI impact (NOT a free move)
- The source-walking audit tests assert on `Sources/<module>/` paths — moving files relocates what they police. Update walked-path roots **in this PR** for `TrafficBoundaryAuditTest` and `DNSRebindingCoverageAuditTest`, plus their paired sabotage entries.
- New test target **`ManifoldNetworkingTests`**.
- The `foundation-only-build` ≤5 MB / no-MLX-symbol gate matters more (more leaf modules = more symbol-leak surface) — verify it stays green.

### Acceptance
- Networking source lives in `ManifoldNetworking`; `import ManifoldInference` still resolves the moved symbols (shim).
- Audit tests updated + green; full `scripts/test.sh --profile local` + trait-combo sweep green.

roryford:
Shipped in #1625

Issue #1590, opened 2026-06-03
perf(ci): Tier 2 — xcodebuild per-target test plans for compile-pruned selective testing


roryford:
## Context

The high-effort, high-reward tier of the CI-time work. **Do not start before Tier 1** (hub split) — without it, ~half of PRs bypass any selective scheme. Sibling issues: Tier 0 (execution-only selective) and Tier 1 (shrink `ManifoldInference`).

## Why this tier exists

Tier 0 is capped because **`swift test --filter` does not prune compile** (measured: clean `--filter ManifoldMCPTests` = 50s / 836 files / 59 modules — builds the whole bundle, filters execution only). Compile is the larger half of the floor (~127s build vs ~136s exec on CI).

`swift build --target X` **does** prune the graph (measured, clean):

| target | build |
|---|---|
| `ManifoldMCPTests` | 13s / 308 files |
| `ManifoldUITests` | 21s / 422 files |
| `ManifoldInferenceTests` | 18s / 417 files |
| full `--build-tests` | ~52s |

…but `swift build --target` builds without a clean run path. **xcodebuild + per-target schemes/test plans** prunes compile per scheme AND runs — the repo already ships `.xcscheme` files (e.g. `ManifoldPersistenceSwiftDataTests.xcscheme`).

## Goal

Move the per-PR test lane (or the narrow-PR path) to `xcodebuild test` driven by per-target schemes/test plans, so a narrow PR compiles and runs only the affected subgraph (~5–7m vs ~16m).

## Hard parts (scope these first)

- **Traits → xcconfig.** SwiftPM traits don't map to xcodebuild directly; the `--traits MCP,CloudSaaS` config must be expressed as build settings / `.define`s. This is the main risk.
- **Scheme/test-plan maintenance** for 24 test targets, kept in sync with the package graph.
- **Metal/simulator gating** already handled in scripts (`scripts/test-mlx-integration.sh`, `scripts/test-ios-simulator.sh`) — reuse, don't duplicate.
- Mixing XCTest + Swift Testing must stay process-isolated (#681).

## Expected payoff

Narrow PRs (~50%, post-Tier-1 more) drop to subgraph build+run. Combined with Tier 0: **~35–40% of total CI minutes**. Still not a uniform 3.5m — hub PRs (reduced by Tier 1) anchor the average.

## Safety net

Same as Tier 0: full suite on `main` push + nightly; force-full on `Package.*`/`.github/`/hub touch; conservative closure.

## Done when

[...]

Issue #1588, opened 2026-06-03
perf(ci): Tier 0 — execution-only selective testing via the changes job


roryford:
## Context

Per-PR CI median rose ~3.5m (early April) → ~16m (June) — a ~4.6× regression driven mostly by codebase growth (+30% Sources/Tests in three weeks). The `test` job builds **and** runs the entire test tree on every PR. This is the cheapest of three scoped tiers to claw time back. See the sibling issues for Tier 1 (hub split) and Tier 2 (xcodebuild test plans).

## Goal

Extend the existing `changes` job (`.github/workflows/ci.yml`, already runs `dorny/paths-filter`) to compute the **affected suite list** from the diff and pass it as the `--filter` set to the `test` job, instead of the hard-coded 11-suite list.

## Measured payoff (80 recent merged PRs)

- Median **2 modules** touched/PR; **51%** of PRs affect ≤3 of 11 suites.
- ~49% touch a hub (`ManifoldInference`→11/11, `Runtime`/`Persistence`→10/11) and must run full — **no selective scheme helps these** (that's Tier 1's job).
- 18% are docs-only (already path-skipped today).

## ⚠️ Known ceiling (measured)

`swift test --filter X` from clean = **50s / 836 files / 59 modules** — it builds the **whole bundle** and filters *execution only*. So Tier 0 saves the execution half (~136s), **not** compile (~127s). Realistic win: a narrow PR ~16m → ~12–13m; **~15–20% of total CI minutes**. Do not expect more from this tier alone — compile pruning is Tier 2.

## Design

1. In `changes`, map changed `Sources/<M>/` and `Tests/<M>Tests/` paths → affected suites using the SwiftPM dependency closure (`swift package describe --type json`; reverse-closure each test target's covered source targets).
2. Emit the affected-suite list as a job output; the `test` step builds its `--filter` args from it.
3. Keep all three test invocations on the unified `--traits MCP,CloudSaaS` set (see PR #1587).

## Safety net (required)

- **Full suite always on push to `main`** + nightly — selective applies to `pull_request` only.
- **Force-full** when the diff touches `Package.swift`, `Package.resolved`, `.github/`, or any hub module (`ManifoldInference`/`ManifoldRuntime`/`ManifoldPersistenceSwiftData`). ~11% of PRs hit this today.
- Conservative closure: when a path maps to no known module, run full.

## Validate-before-commit

[...]

PR #1626, opened 2026-06-05
perf(ci): Tier 0 selective-testing resolver (dry-run mode) (#1588)

roryford:
## What

Ships the **Tier 0 selective-testing resolver** from #1588 in **dry-run mode**: it computes which per-PR `test`-job suites a diff actually affects (from the SwiftPM dependency graph) and **logs** the decision, but the test steps still run the full hardcoded list. No behavior change to what executes — this PR cannot regress coverage.

Phase 2 (wiring the test steps to consume the resolver's outputs) lands in a follow-up **after ~1 week of dry-run data confirms the projected payoff**, per the issue's validate-before-commit requirement.

## Why dry-run first

`swift test --filter` prunes **execution only, not compile** (measured: clean filtered run still builds the whole bundle — see #1588/#1590). So the realistic Tier 0 win is the execution half: a narrow PR ~16m → ~12–13m, **~15–20% of total CI minutes**. We want to confirm that against live traffic before flipping the switch.

## Changes

- `scripts/ci-affected-suites.py` — the resolver:
  - Reads the **maximal** graph via `swift package --enable-all-traits describe` (load-bearing: plain `describe` drops trait-off test targets → under-include → false green).
  - Maps changed files to targets by longest-prefix on the real `.path` (`ManifoldBackends` lives in `Sources/ManifoldBackendsUmbrella/`).
  - Reverse-closure: a changed source target selects every CI suite whose dependency closure includes it; a changed CI test target selects itself.
  - **Fails open**: manifest / `.github/` / hub-module / unknown-path / describe-error all force a full run. Never an empty one.
- `.github/workflows/ci.yml`:
  - `test`-job checkout → `fetch-depth: 0` (needed for the `base...HEAD` merge-base diff).
  - New **"Resolve affected suites (dry-run)"** + **"Tier 0 dry-run summary"** steps (PR-only; `--force-full` on push/nightly).
  - Resolver script added to both `paths:` filters so editing it re-triggers CI.
- `scripts/ci-tier0-dryrun-report.sh` — aggregates the `TIER0_DRYRUN` markers across recent runs: % selective-vs-full and % of suite-executions that would be skipped (the go/no-go signal for Phase 2).

## How payoff is measured

After ~1 week of PRs, run:

```
scripts/ci-tier0-dryrun-report.sh --limit 120
```

Flip Phase 2 only if it shows the projected ~15–20% execution-time reduction.

## Test impact

CI-only (no Swift sources touched). The dry-run step + YAML are exercised by this PR's own CI run; resolver logic validated locally against the live package graph (leaf→1 suite, hub/manifest/unknown→force-full, umbrella path-mismatch handled).

Refs #1588. Tier 1 (#1608) and Tier 2 (#1590) are separate.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,7 @@ on:
       - "scripts/check-foundation-only-bundle.sh"
       - "scripts/test-ios-simulator.sh"
       - "scripts/lint-no-new-force-unwraps.sh"
+      - "scripts/ci-affected-suites.py"
   pull_request:
     branches: [main]
     paths:
@@ -31,6 +32,7 @@ on:
       - "scripts/check-foundation-only-bundle.sh"
       - "scripts/test-ios-simulator.sh"
       - "scripts/lint-no-new-force-unwraps.sh"
+      - "scripts/ci-affected-suites.py"
 
 concurrency:
   group: ci-${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}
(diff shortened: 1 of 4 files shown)


roryford:
Closing as a duplicate of #1619, which shipped the Tier-0 selective-testing dry-run resolver and merged to `main` as 3591821d. Both PRs independently implemented #1588 (this one in Python, #1619 in bash + jq); #1619 landed first and #1588 stays open for the switch-flip follow-up. Thanks!

Issue #1622, opened 2026-06-05
fix: resource-correctness bugs (crash / leak / OOM)


roryford:
## Summary

A batch of genuine, user-visible correctness bugs (crash / leak / OOM) found during a footgun audit. Each is a small, independent fix. Ship as one PR (or a short stack merged as a unit) per the "fewer, larger PRs" rule.

## Checklist

- [ ] **MLX GPU-memory leak — `MLXBackend` has no `deinit`** (`Sources/ManifoldMLX/MLXBackend.swift`, class :24, `unloadModel` :559). `LlamaBackend` releases C resources in `deinit` (`LlamaBackend.swift:170`); `MLXBackend` relies entirely on the caller. A dropped instance (VM teardown, error path, A/B swap) leaves a stale per-UUID claim in `MLXResourceArbiter` forever — and since the arbiter only fires `MLX.Memory.clearCache()` on the *last* release, that guarantee then never fires again and Metal buffers never return to the OS. Add a `deinit` mirroring Llama's retain/detach/release pattern.

- [ ] **RAG crash instead of fallback — `embed(...)[0]`** (`Sources/ManifoldRuntime/Services/RAGService.swift:143`). `EmbeddingBackend.embed` does not guarantee output count == input count; an empty/short return makes `[0]` a runtime *trap* that the surrounding `do/catch` (designed to fall back to keyword search) cannot catch. Use `.first else { throw }` so the failure routes into the existing catch.

- [ ] **Multi-term message search OOM** (`Sources/ManifoldPersistenceSwiftData/Adapters/SwiftDataPersistenceProvider.swift:239`). `fetchLimit` is set only for single-term queries. Multi-word search matches just the first term with **no limit**, loading every matching row into RAM before filtering in Swift. Common first word on a large history → OOM/stall. Set the limit on all branches (or route through one fetch-builder — see structural issue).

- [ ] **MCP request-timeout leaks the continuation + concurrency slot** (`Sources/ManifoldMCP/InternalMCPSession.swift:96,250`). `withTimeout` resumes the outer continuation but never cancels the operation task or removes `pendingRequests[id]`. A server that accepts a request but never answers that id leaks the slot permanently; after enough events every `sendRequest` throws "exceeded max concurrent" until reconnect. Also fails to send `notifications/cancelled` on the timeout path. Cancel the op task + evict the pending entry (+ notify) when the timeout wins.

[...]

PR #1629, opened 2026-06-05
fix: five correctness bugs from footgun audit (#1622)

roryford:
## Summary

Fixes five correctness bugs identified in the 2026-06-05 footgun audit:

- **MLXBackend deinit leak**: `MLXResourceArbiter` claim was never released when a backend instance was dropped without an explicit `unloadModel()` call. Fixed by adding `deinit { unloadModel() }`.
- **RAGService retrieve crash**: `backend.embed([query])[0]` would crash with a subscript-out-of-bounds if the backend returned an empty array (contract violation). Fixed with `guard let ... = .first else { throw RAGError.emptyEmbeddingResult }`, which falls back to keyword search via the existing catch block.
- **SwiftData searchMessages OOM**: `fetchLimit` was only applied to single-term queries; multi-term queries fetched unbounded rows into memory before in-memory filtering. Fixed by always setting `descriptor.fetchLimit = limit`.
- **MCPSession timeout slot leak**: A timed-out request left its entry in `pendingRequests`, permanently exhausting `maxConcurrentRequests` after enough timeouts. Fixed by calling `timeoutPendingRequest(id:)` (a new `async` bridge over the synchronous `failPendingRequest`) from the timeout callback. Also sends `notifications/cancelled` to the server.
- **RAGService ingest orphan**: If `documentStore.insertDocument` failed after `vectorStore.insert` succeeded, the vector chunks were orphaned permanently. Fixed with rollback: delete the vector entries and rethrow.

## Test plan

- [ ] `swift test --filter MCPSessionTests --traits MCP --disable-default-traits` — new `test_sendRequest_timeout_evictsPendingSlot` verifies slot is freed and `notifications/cancelled` is sent
- [ ] `swift test --filter RAGServiceTests --disable-default-traits` — `test_retrieve_embedReturnsEmpty_fallsBackToKeyword` and `test_ingest_documentStoreThrows_rollsBackVectors`
- [ ] `swift test --filter SwiftDataPersistenceProviderSearchTests --disable-default-traits` — `test_searchMessages_multiTermQuery_respectsLimit`
- [ ] `swift test --filter MLXBackendTests --traits MLX` — `test_deinit_afterMockInjection_callsUnloadModel`
- [ ] Trait-combo build: `swift build --build-tests --traits MLX,Llama,Ollama,CloudSaaS,MCP,MCPBuiltinCatalog,HuggingFace` ✅

### DIFF
--- a/Sources/ManifoldMCP/InternalMCPSession.swift
+++ b/Sources/ManifoldMCP/InternalMCPSession.swift
@@ -94,7 +94,18 @@ internal actor MCPSession {
         // synchronously and cannot be async, so we spawn a detached Task to deliver
         // notifications/cancelled before the server wastes work on an abandoned call.
         return try await withTaskCancellationHandler {
-            try await withTimeout(requestTimeout) { [self] in
+            try await withTimeout(requestTimeout, onTimeout: { [self] in
+                // Remove the pending slot and resume its continuation so the operation
+                // task completes and the concurrency slot is freed. Without this the
+                // slot leaks permanently — a server that accepts but never answers the
+                // request ID would exhaust maxConcurrentRequests after enough timeouts.
+                let cancelParams: JSONSchemaValue = .object([
+                    "requestId": .string(id.description),
+                    "reason": .string("Request timeout"),
+                ])
+                await sendNotificationIgnoringErrors(method: "notifications/cancelled", params: cancelParams)
+                await timeoutPendingRequest(id: id)
+            }) { [self] in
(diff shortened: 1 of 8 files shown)


roryford:
Superseded by #1627, which landed the same five fixes with a more complete implementation (metallib guard in MLX deinit, atomic claimResume in withTimeout, smarter fetchLimit multiplier for multi-term queries). Closing to avoid a no-op duplicate merge.

PR #1665, opened 2026-06-06
feat(inference): adaptive prefill memory headroom from measured per-model footprint

### LINKED ISSUE #1592: feat(inference): adaptive prefill memory headroom from measured per-model footprint
## Problem

`ModelLoadPlan` (`Sources/ManifoldInference/Services/ModelLoadPlan.swift`) decides memory fit and context clamps with a **static** `headroomFraction` (default 40%) plus a fixed iOS overhead reserve. The estimate is computed once, before load, from per-token KV-byte heuristics. It can't see what a given model actually costs during prefill on the current device under current memory pressure, so it is forced to be conservative (leaving capacity on the table) and can still be wrong when real conditions differ from the heuristic. The failure mode is a jetsam kill on iOS or a Metal allocation panic on macOS — neither of which the up-front static check anticipates or recovers from.

## What this brings MK

- **Safer large-context prefill on-device.** A footprint learned from the real model + device + current pressure catches the cases a fixed fraction misses.
- **Less wasted capacity.** When measured cost is below the heuristic, MK can allow a larger effective context instead of always over-reserving 40%.
- **A clean mid-prefill abort path.** Today a prefill that is about to exceed budget runs until the OS kills the process. A measured per-chunk estimate lets MK decline the *next* chunk and surface `InferenceError.memoryInsufficient(required:available:)` instead of dying.

## Proposed shape

- Add a per-model footprint estimator: an exponentially-weighted moving average of resident-byte delta per prefilled token, updated at chunk boundaries. MK already emits chunk progress via `GenerationEvent.prefillProgress(nPast:nTotal:tokensPerSecond:)`; sample actual resident bytes there (e.g. `task_vm_info.phys_footprint` via mach). **Skip negative deltas** (MLX cache-pool reclaim larger than the chunk's allocation) so they don't bias the estimate toward zero and cause an underestimate next time.
- Feed the running estimate back into `ModelLoadPlan.Outcome` so `effectiveContextSize` / verdict can be recomputed against a *measured* budget, falling back to the current static heuristic until the first sample exists.
- Add a pre-chunk guard: if predicted transient (`estimate × next-chunk-tokens × safety factor`) exceeds remaining headroom, abort with `.memoryInsufficient` rather than continuing into a panic.

Scope: on-device backends (MLX, Llama). Cloud / Foundation are out of scope.

## Acceptance

- Estimator unit-tested with synthetic delta sequences, including negative-delta rejection and first-sample fallback.
- Integration test: prefill on a small model produces a non-zero learned estimate and a measurably different effective context than the static path, asserted deterministically.
- Mid-prefill abort surfaces `.memoryInsufficient` (with a sabotage check).
- Default behavior unchanged until the documented sample threshold is reached.
- Tests and docs ship in the feature PR.

roryford:
Closes #1592.

## What existed vs. what this builds

Already shipped: static headroom budgeting in `ModelLoadPlan` (fixed 40% `headroomFraction`), the KV bytes/token heuristic (`GGUFKVCacheEstimator`), `GenerationEvent.prefillProgress`, resident-footprint sampling (`AppMemoryUsage.currentBytes()` → `task_vm_info.phys_footprint`), and `InferenceError.memoryInsufficient` (thrown only at load).

The gap was using a *measured* per-model footprint to adapt the budget at prefill time. This PR fills it:

### 1. `PrefillFootprintEstimator` (new, `Sources/ManifoldHardware/`)
Pure, `Sendable` value type. An exponentially-weighted moving average of resident-byte delta per prefilled token.

- **Negative-delta rejection.** A chunk can show a *negative* delta when an allocator/cache-pool reclaim frees more than the chunk allocated. Folding those in (even clamped to 0) biases the EWMA toward zero and makes the next prediction an underestimate — the wrong direction for a safety guard. Negative deltas are therefore **rejected**, not clamped, and counted in `rejectedSampleCount` for observability.
- **First-sample seed + dormant guard.** The estimate is `nil` until the first accepted sample; the abort guard (`wouldExceedHeadroom`) returns `false` while dormant, so pre-measurement behaviour is unchanged and a missing sample never triggers a spurious abort.

### 2. Chunk-boundary sampling + abort guard (Llama)
MLX prefill is opaque (it happens inside the mlx-swift-lm `generate()` call — no per-chunk callback to hook). **Llama** has a BCK-controlled chunked prompt-decode loop in `LlamaGenerationDriver`, which is the honest integration point:

- Resident footprint is sampled before/after each `llama_decode` and the delta fed into the EWMA.
- **Before** each chunk, if the learned cost predicts the chunk's transient (`estimate × tokens × safetyFactor`, default 1.5×) would overrun the free memory remaining *right now*, the driver aborts with `InferenceError.memoryInsufficient(required:available:)` instead of decoding into a jetsam/Metal kill. Samplers are injected (real Mach footprint + free-memory query by default) so the loop stays unit-testable.
- `LlamaBackend` retains the learned cost via a new `lastMeasuredBytesPerToken` accessor so a subsequent load can recompute its plan against a measured budget.

### 3. Measured-budget recompute (`ModelLoadPlan`)
`Inputs` gains an optional `measuredBytesPerToken`. When present and non-zero it overrides `kvBytesPerToken` for the memory-ceiling math, recomputing `effectiveContextSize`. The `compute(for: ModelInfo, …)` overloads thread it through. **When absent (the default), the outcome is byte-for-byte identical to the static path** — proven by a regression test asserting `Outcome` equality.

## GenerationEvent
**Unchanged.** The footprint sample is threaded internally through the driver, so no switched-enum change and no Cloud/Ollama exhaustiveness fallout. (An all-traits build sweep was run anyway since `ModelLoadPlan` is widely consumed.)

## Abort-guard policy
The guard is conservative: dormant until the first sample, inflates predictions by a 1.5× safety factor, samples free memory fresh each chunk (tracks live pressure), and aborts cleanly via a thrown error — never `fatalError`/`assertionFailure`. A pre-chunk abort marks the KV cache incoherent so the next turn clears it.

## Static-default regression guarantee
The adaptive path engages only once real samples exist. No measurement → identical plan, identical prefill behaviour. Locked by `test_absentMeasurement_outcomeIdenticalToStaticPath` and `test_measuredEqualsHeuristic_isNoOp`.

## Tests
- **Unit** (`PrefillFootprintEstimatorTests`, 11 tests, real & green): synthetic delta sequences — EWMA convergence, negative-delta rejection (+ all-negative, zero-token, nil-reading edge cases), first-sample fallback, safety-factor inflation, abort-guard fires/doesn't-fire. **Sabotage-verified**: dropping `residentBytesDelta >= 0` makes the rejection tests fail loudly; reverted.
- **Unit** (`ModelLoadPlanTests`, +3): measured cost recomputes effective context (4× cost → 50k → 15k), nil measurement is byte-for-byte identical, measured==heuristic is a no-op.
- **Integration** (`LlamaPrefillFootprintIntegrationTests`, skip-gated on Apple Silicon + GGUF): loads a real model, prefills, asserts a positive learned estimate that feeds a plan. **Ran green in 3.69s against `llama3.1-8b`** (real convergence, not faked). Skips cleanly with a clear message when no model is on disk.

The live mid-prefill OOM abort is intentionally **not** exercised as an integration test (forcing a true OOM in CI is unsafe); its decision logic is fully unit-tested via `wouldExceedHeadroom`.

## Gate
- `swift build --build-tests` (default traits): green.
- `swift build --traits MLX,Llama,Ollama,CloudSaaS,MCP,MCPBuiltinCatalog,HuggingFace`: green.
- `ManifoldHardwareTests` estimator + plan suites: 59 passed, 0 failed.
- Llama integration: 1 passed (with model) / skips cleanly without.

### DIFF
--- a/Sources/ManifoldHardware/ModelLoadPlan.swift
+++ b/Sources/ManifoldHardware/ModelLoadPlan.swift
@@ -192,9 +192,12 @@ public struct ModelLoadPlan: Sendable {
         // the static `kvBytesPerToken` otherwise. When no measurement exists this
         // resolves to `kvBytesPerToken`, keeping the static path byte-for-byte
         // identical to the pre-adaptive behaviour.
-        let effectiveBytesPerToken: UInt64 = (inputs.measuredBytesPerToken ?? 0) > 0
-            ? inputs.measuredBytesPerToken!
-            : inputs.kvBytesPerToken
+        let effectiveBytesPerToken: UInt64
+        if let measured = inputs.measuredBytesPerToken, measured > 0 {
+            effectiveBytesPerToken = measured
+        } else {
+            effectiveBytesPerToken = inputs.kvBytesPerToken
+        }
 
         // Memory-derived token ceiling. Guarding against div-by-zero when the caller
         // passed kvBytesPerToken == 0 (shouldn't happen in practice, but cheap to be safe).
--- a/Sources/ManifoldHardware/ModelLoadPlan.swift
+++ b/Sources/ManifoldHardware/ModelLoadPlan.swift
@@ -21,6 +21,12 @@ public struct ModelLoadPlan: Sendable {
         public let absoluteContextCeiling: Int
         public let headroomFraction: Double
         public let appOverheadBytes: UInt64
+        /// Measured per-token resident cost learned during a prior prefill
+        /// (see ``PrefillFootprintEstimator``). When present and non-zero it
+        /// overrides ``kvBytesPerToken`` for the memory-ceiling math, letting the
+        /// plan recompute `effectiveContextSize` against a *measured* budget. `nil`
+        /// (the default) leaves the static heuristic path byte-for-byte unchanged.
+        public let measuredBytesPerToken: UInt64?
 
         public init(
             modelFileSize: UInt64,
@@ -32,7 +38,8 @@ public struct ModelLoadPlan: Sendable {
             physicalMemoryBytes: UInt64,
             absoluteContextCeiling: Int,
             headroomFraction: Double,
-            appOverheadBytes: UInt64 = 0
+            appOverheadBytes: UInt64 = 0,
+            measuredBytesPerToken: UInt64? = nil
         ) {
             self.modelFileSize = modelFileSize
             self.memoryStrategy = memoryStrategy
@@ -44,6 +51,7 @@ public struct ModelLoadPlan: Sendable {
             self.absoluteContextCeiling = absoluteContextCeiling
             self.headroomFraction = headroomFraction
             self.appOverheadBytes = appOverheadBytes
+            self.measuredBytesPerToken = measuredBytesPerToken
         }
     }
 
@@ -179,11 +187,20 @@ public struct ModelLoadPlan: Sendable {
             ? reserveAfterHeadroom - residentBudget
             : 0
 
+        // Per-token cost for the memory math. Prefer a measured estimate (learned
+        // from a real prefill) when one diverges from the heuristic; fall back to
+        // the static `kvBytesPerToken` otherwise. When no measurement exists this
+        // resolves to `kvBytesPerToken`, keeping the static path byte-for-byte
+        // identical to the pre-adaptive behaviour.
+        let effectiveBytesPerToken: UInt64 = (inputs.measuredBytesPerToken ?? 0) > 0
+            ? inputs.measuredBytesPerToken!
+            : inputs.kvBytesPerToken
+
         // Memory-derived token ceiling. Guarding against div-by-zero when the caller
         // passed kvBytesPerToken == 0 (shouldn't happen in practice, but cheap to be safe).
-        let memoryCeiling: Int = inputs.kvBytesPerToken == 0
+        let memoryCeiling: Int = effectiveBytesPerToken == 0
             ? Int.max
-            : Int(kvBudget / inputs.kvBytesPerToken)
+            : Int(kvBudget / effectiveBytesPerToken)
 
         // Walk the ceilings in turn, recording a `Reason` each time a tighter bound wins.
         // `effective` tracks the winning value; we re-check against `requestedContextSize`
@@ -220,7 +237,7 @@ public struct ModelLoadPlan: Sendable {
         // Never pass 0 / negative to llama.cpp, even in pathological near-zero memory.
         let effectiveContextSize = max(1, effective)
 
-        let estimatedKV = UInt64(effectiveContextSize) * inputs.kvBytesPerToken
+        let estimatedKV = UInt64(effectiveContextSize) * effectiveBytesPerToken
         let total = residentBudget &+ estimatedKV  // residentBudget is already bounded
 
         // Impossible-fit guard: regardless of strategy, a model file that is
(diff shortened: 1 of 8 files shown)


roryford (commented):
**Summary**: Solid design overall — EWMA footprint tracking, mid-prefill abort on OOM prediction, and feedback into `ModelLoadPlan` are all the right moves. The dormant-guard policy (returns `false` until first sample) and the `measuredBytesPerToken: nil` regression test are done well. One issue needs fixing before merge.

---

**[MAJOR] `ModelLoadPlan.swift:45` — force-unwrap after `?? 0` guard**

```swift
// current
let effectiveBytesPerToken: UInt64 = (inputs.measuredBytesPerToken ?? 0) > 0
    ? inputs.measuredBytesPerToken!   // ← force-unwrap
    : inputs.kvBytesPerToken
```

Logically unreachable, but structurally unsafe — same failure class as `try?` in production paths. Use `if let` to make it compiler-proven:

```swift
let effectiveBytesPerToken: UInt64
if let measured = inputs.measuredBytesPerToken, measured > 0 {
    effectiveBytesPerToken = measured
} else {
    effectiveBytesPerToken = inputs.kvBytesPerToken
}
```

---

**[MINOR] `LlamaGenerationDriver.swift:366` — `MainActor.run` inconsistency**

The abort block uses `MainActor.run { setPhase(.failed(…)) }` but the nearby `promptDecodeFailed` path calls `setPhase` directly. If `setPhase` is `@MainActor`-isolated, both failure paths need the hop; if it isn't, the hop here is unnecessary. Make them consistent.

---

**[MINOR] `LlamaPrefillFootprintIntegrationTests.swift:499` — hard `XCTFail` on net-reclaim will flake on constrained CI**

The comment acknowledges "net reclaim across all chunks" is a valid real-world observation on memory-pressured hardware, but the test still fails loudly on it. Downgrade to `throw XCTSkip("Net reclaim observed — device under memory pressure; re-run on an unconstrained host")` so CI stays green on constrained runners.

---

**[NIT] `LlamaBackend.swift:292` — implicit `self` capture in `@Sendable` closure**

No actual retain cycle (the closure isn't stored), but adding `[self]` makes the intent explicit and keeps Swift 6 strict-concurrency lint quiet.

roryford:
Addressed all four review points (pushed 069dcbe):

1. **[MAJOR] ModelLoadPlan force-unwrap** — replaced `inputs.measuredBytesPerToken!` (after `?? 0` guard) with a compiler-proven `if let measured = ..., measured > 0`. Static path still resolves to `kvBytesPerToken` byte-for-byte.
2. **[MINOR] MainActor consistency** — verified `setPhase` lives on `@MainActor`-isolated `GenerationStream` and `run(...)` is non-isolated `async`, so the hop is required. Both the prefill-abort path and the `promptDecodeFailed` path already use `await MainActor.run { ... }` — consistent, no change needed.
3. **[MINOR] net-reclaim XCTFail** — downgraded to `throw XCTSkip(...)` so memory-pressured CI runners stay green (test method is `async throws`).
4. **[NIT] implicit self** — added explicit `[self]` capture to the `@Sendable onPrefillEstimate` closure in `LlamaBackend`.

Own correctness pass found nothing further: estimator rejects negative deltas (not clamped), seeds the EWMA on first sample, floors `safetyFactor` at 1.0 and saturates on overflow; the abort guard is dormant (returns false) until the first sample so single-chunk / pre-measurement behavior is unchanged. No other force-unwraps / `try?` / `fatalError` in the new production paths.

Gate: `swift build --build-tests --traits MLX,Llama,Ollama,CloudSaaS,MCP,MCPBuiltinCatalog,HuggingFace` green; estimator + plan suites 68/68 pass; Llama prefill integration suite compiles (skips — no GGUF on disk).

PR #1699, opened 2026-06-07
feat(fuzz): add --request-timeout to bound cloud fuzz request hangs

roryford:
## Problem

PR #1690 added `--backend openai` to `fuzz-chat` for fuzzing OpenAI-compatible cloud endpoints (OpenRouter). In live testing, slow/throttled free models hung up to **~300s per request** before timing out, because the cloud transport idle timeout is 300s (`URLSessionProvider.timeoutIntervalForRequest`, shared with production chat). The fuzz detectors already flag anything over 60s (race-stall first-token > 60_000ms; timeout total > 60_000ms), so waiting 300s adds no signal — it just destroys throughput.

## Fix

Add a `--request-timeout <seconds>` CLI flag (default **90**) that bounds the per-request HTTP idle timeout for the cloud (`--backend openai`) path **only**. A hung iteration now abandons at 90s instead of 300s while still capturing the slow-but-real cases the detectors flag at 60s.

Implementation:
- Threads the value through `OpenAIFuzzFactory` (new `requestTimeout: TimeInterval = 90` init param) onto the backend's existing per-request override `SSECloudBackend.requestIdleTimeout` (already `public`, already applied as `URLRequest.timeoutInterval` per generation — no access widening needed).
- Production access surface is untouched: the global 300s/600s session defaults in `URLSessionProvider` are **not** changed. Production chat (where long generations may legitimately need minutes) is unaffected.
- CLI validates the value is a positive number (rejects `<= 0`), documents it in `--help`, and prints a one-line note if it's passed with a non-openai backend (which has no HTTP idle window).
- `scripts/fuzz.sh` forwards the flag automatically (passthrough) and documents the rationale in both `--help` and the cloud preflight caveats.

## Invocation

```bash
swift run --traits Fuzz,CloudSaaS,MLX,Llama,Ollama fuzz-chat \
  --backend openai --base-url https://openrouter.ai/api \
  --model <slug> --request-timeout 90 ...
```

Detectors already flag anything over 60s, so the 90s bound loses no signal.

## Tests

`Tests/ManifoldFuzzTests/OpenAIFuzzFactoryTests.swift` (network-free): asserts `OpenAIFuzzFactory(requestTimeout: 42, ...)` propagates 42 to the backend's `requestIdleTimeout` override, and that the default is 90. Verified with a temporary sabotage (`requestTimeout + 1`) that both tests fail when propagation breaks; sabotage removed before commit.

## Gate

- `swift build --traits Fuzz,CloudSaaS,MLX,Llama,Ollama` — green
- `swift build --build-tests --traits MLX,Llama,Ollama,CloudSaaS,MCP,MCPBuiltinCatalog,HuggingFace,Fuzz` (trait-combo sweep) — green
- `scripts/test.sh --profile local` — **PASSED** (531 XCTest suites + 83 Swift Testing, 0 failures)


roryford:
Closing as a duplicate of #1700 (merged). Both PRs added the same `--request-timeout` flag — #1700 was branched off newer main and is the one that landed.

PR #1792, opened 2026-06-13
fix: branch flow uses transactional copy with rollback

roryford:
## Summary

Fixes verified correctness bugs in the branch flow of `ConversationTurnExecutor`. Findings confirmed by a 2/2 adversarial audit and re-verified against the live code in this branch.

### Finding 1+2 — branch orphans a session + non-transactional copy (high, correctness)
`Sources/ManifoldRuntime/Services/ConversationTurnExecutor.swift:338-352` (pre-fix): `runBranchFlow` inserted the new session, then copied messages with a per-message `insertMessage(...)` loop that rethrew on the first failure **without** rolling back the already-inserted session. A failed branch left a ghost/partially-copied session that surfaces as a phantom in the sidebar; the copy also bypassed the transactional batch API the edit flow already uses.

**Fix:** build `[MessageStoreMutation]` inserts for the slice and call `performMessageMutations(...)` once (atomic on transactional stores), wrapped in a `do/catch` that calls `deleteSession(newSessionID)` before rethrowing. `transaction(block:)` does not roll back on throw in this codebase, so the rollback is an explicit best-effort delete (the port's `deleteSession` already logs rather than throwing, preserving the original copy error as caller-visible).

### Finding 3 — handoff writes a stale full session record (medium, concurrency)
`Sources/ManifoldRuntime/Services/ConversationTurnExecutor.swift:828-833` (pre-fix): the mid-stream handoff path wrote the whole record via `updateSession(current)`, clobbering concurrent edits to other session columns.

**Fix:** switch to the narrow `setActiveAgent(sessionID:agentID:)` write, which the persistence port and `SessionStore` already expose for exactly this purpose. No new port surface added.

## Tests
Added two integration tests (in-memory SwiftData-shaped fakes) in `Tests/ManifoldRuntimeTests/ConversationRuntimeTests.swift`:
- `test_branch_midCopyInsertFails_rollsBackSessionLeavingNoOrphan` — extends the mock store with a fail-at-Nth-insert hook so the 2nd of three branch copies fails mid-batch; asserts the new session is rolled back (no orphan). Sabotage-verified: removing the `deleteSession` rollback fails this test (then removed).
- `test_branch_success_copiesEveryMessageInSlice` — asserts a successful branch copies the full slice.

Existing handoff scenario tests (`HandoffScenarioTests`) still pass — the default `setActiveAgent` extension routes through `updateSession`, so persisted state is unchanged.

## Gate (spike profile)
- `swift build --build-tests` — passed (deprecation warnings only)
- `swift test --filter ManifoldRuntime` — 466 tests, 0 failures, 2 skipped

No `Package.resolved` staged.

### DIFF
--- a/Sources/ManifoldRuntime/Services/ConversationTurnExecutor.swift
+++ b/Sources/ManifoldRuntime/Services/ConversationTurnExecutor.swift
@@ -336,19 +336,32 @@ package struct ConversationTurnExecutor: Sendable {
         }
 
         // Copy messages into the new session with fresh IDs and updated
-        // sessionID.
-        for original in slice {
-            let copy = ChatMessage(
+        // sessionID. Drive the copy through the transactional batch API (same
+        // as the edit flow) so a transactional store commits the whole slice
+        // atomically and never exposes a half-copied branch.
+        //
+        // The session insert above commits separately from the message batch
+        // (the adapter's transaction spans messages only), so if the copy
+        // throws we must manually unwind the orphaned session — otherwise a
+        // failed branch strands a ghost/empty session that surfaces as a
+        // phantom in the sidebar. `transaction(block:)` does NOT roll back on
+        // throw in this codebase, so the rollback is an explicit delete.
+        let copyMutations: [MessageStoreMutation] = slice.map { original in
+            .insert(ChatMessage(
                 role: original.role,
                 contentParts: original.contentParts,
                 timestamp: original.timestamp,
                 sessionID: newSessionID
-            )
-            do {
-                try await persistence.insertMessage(copy)
-            } catch {
-                throw ConversationError.persistence(error)
-            }
+            ))
+        }
+        do {
+            try await persistence.performMessageMutations(copyMutations)
+        } catch {
+            // Roll back the session we just inserted before surfacing the copy
+            // failure. `deleteSession` is best-effort and logs on its own
+            // failure so the original copy error stays the caller-visible one.
+            await persistence.deleteSession(newSessionID)
+            throw ConversationError.persistence(error)
         }
 
         emit(.sessionBranched(newSessionID: newSessionID, copiedCount: slice.count))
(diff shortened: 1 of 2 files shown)


roryford:
Independent review — verdict: **CLEAN**, no code changes needed. Gate (`swift build --build-tests` + `swift test --filter ManifoldRuntime`) is green: 466 tests, 0 failures, both new branch tests run and pass.

The most-scrutinized question — does the mid-stream handoff `setActiveAgent` truly write only the active-agent field, or does it route through `updateSession` and reintroduce the stale-snapshot clobber? — resolves cleanly:

- The **production SwiftData path** (`SwiftDataPersistenceProvider.setActiveAgent`, lines 155–160) is a genuine narrow in-place write: `fetch → session.activeAgentID = agentID → save`. It does **not** touch `updateSession`. The fix is real, not cosmetic, for the path that actually ships.
- The `SessionStore` **default extension** (lines 176–181) does route through `updateSession` — but it re-fetches the live session first and mutates only `activeAgentID`, so it is **not** the turn-start stale snapshot the old code passed. It's a read-modify-write that's documented as non-atomic and reserved for in-memory test doubles. No clobber reintroduced; worst case it loses a concurrent edit on a non-SwiftData store, which the doc comment already calls out.

So the lost-update bug the fix targets is genuinely closed on the SwiftData adapter.

Other findings:
- **Rollback correctness** is sound: the port's `deleteSession` is non-throwing (swallows + logs internally), so the original copy error is always the caller-visible one, and the delete runs before the rethrow. Order/atomicity correct.
- **Slice fidelity**: `sourceHistory[...branchIndex]` is byte-for-byte the same slice as the old loop; copies preserve order, role, contentParts, timestamp, fresh IDs, new sessionID. No off-by-one.

One minor coverage note (not blocking): `RuntimeMessageStore` is **not** a `TransactionalMessageStore`, so `test_branch_midCopyInsertFails` exercises the port's **non-transactional fallback loop** — a faithful real failure mode (the partial-write window that fallback stores genuinely have), and the session rollback is genuinely exercised there. It does **not** cover the SwiftData transactional path (atomic message rollback + separate session commit needing the explicit `deleteSession`). That path's session-rollback need is real but remains unit-untested here; a transactional-mock variant could close it. The existing test is valid and not a misrepresentation — it just doesn't cover both branches of the port `if`.

PR #1799, opened 2026-06-13
feat(inference): per-request Deep-backend routing seam (#799 P0)

roryford:
Phase 0 of Fireside #799 ("two engines, one story") — the generic infra for routing a single queued generation to a coexisting secondary backend without disturbing the primary.

## What does this change?

Adds a **host-owned secondary backend** (`deepBackend`) to `InferenceService` so that Deep-scoped turns can be dispatched to a stronger engine while the primary stays loaded. Introduces `GenerationRoute { .primary, .deep }` and threads it through the full queued path (`enqueue → drainQueue → runToolDispatchLoop → dispatchToBackend`), including correct `cancel`/`stopGeneration` targeting of the routed backend.

## Motivation

Fireside wants Deep-scoped turns to run on a stronger engine (first pairing: local Fast + cloud Deep). The single `InferenceService` unloads the primary on any endpoint switch, so naive "swap the primary" tears down the local model. Instead, this adds a host-owned secondary backend that a routed request dispatches to while the primary stays loaded — with a cloud Deep backend that coexists at zero local-memory cost.

This reuses the **ownership contract** of `fastBackend` (host-owned, never lifecycle-managed) but routes through the **queued** path (not the unqueued `runFastOrPrimary`), so Deep turns keep cancellation tokens, request-group scoping, and stream lifecycle.

#### Per-request Deep-backend routing seam

`InferenceService` now accepts a host-owned `deepBackend` that coexists with the primary without triggering a model unload. A new `GenerationRoute` parameter (`.primary` / `.deep`, defaulting to `.primary`) on every `enqueue` overload lets callers direct individual turns to the Deep backend while the local primary stays loaded.

```swift
inferenceService.deepBackend = myCloudBackend

// route this turn to the Deep backend; primary stays loaded
let (_, stream) = try inferenceService.enqueue(
    messages: history,
    route: .deep
)
```

Additive and default-preserving: omitting `route:` is byte-identical to existing behaviour.

## Testing

- `DeepBackendRoutingTests` (5 new tests): deep route serves & leaves primary loaded/not-unloaded; primary route ignores deep; nil-deep falls back to primary; stop targets the deep backend not the primary; introspection reflects state.
- Regression: 76 existing queue/enqueue/fast-backend tests green.

## Sabotage evidence (regression-test PRs only)

- [x] N/A (not a regression-test PR)
- [ ] Reverted the fix / broke the path under test
- [ ] Confirmed the new test went RED (log pasted or linked below)
- [ ] Re-applied the fix and confirmed GREEN

## Checklist

- [x] Tests added or updated for new behaviour
- [x] Public API changes have `///` doc comments
- [x] No hardcoded secrets, API keys, or personal data
- [ ] Breaking change? (if yes, describe migration path below)

> **API breakage:** every `enqueue` overload gains a trailing `route: GenerationRoute = .primary` parameter. The allowlist entries in `.github/api-breakage-allowlist.txt` cover the resulting selector renames; callers omitting the parameter compile unchanged.

## DX checklist

- [ ] Did this change a public API a user might write in their README? If yes, updated the README snippet.
- [ ] Did this add or change a trait, backend, or capability? If yes, updated `Sources/ManifoldKit/FeatureMatrix.swift` (when present).
- [ ] Is this a breaking change for an existing consumer? If yes, added a migration note to `CHANGELOG.md` or `docs/`.
- [ ] Did this change the `quickStart()` path or `MinimalExample`? If yes, the example still compiles and runs end-to-end.
- [ ] Did this introduce a new public error type or surface? If yes, it conforms to `LocalizedError` with a user-facing `errorDescription`.


roryford:
@copilot resolve the merge conflicts in this pull request

Issue #1832, opened 2026-06-14
feat(contract): surface an explicit terminal completion event on GenerationStream


roryford:
**Context:** UI/accessibility layers that drive announcements off the event stream want a single "response finished" signal in-band.

Today completion is observable only by the `AsyncThrowingStream` finishing or `phase == .done` — awkward to race against the event iterator. There is no `.completed`/`.done` event case.

**Ask:** add a terminal `.generationCompleted(reason:)` event (or a `@MainActor` completion closure on `GenerationStream`) so consumers can post a single "finished" announcement without racing iterator vs. phase observation.

roryford:
Shipped in #1838: `GenerationEvent` now carries a terminal `case generationCompleted(GenerationCompletion)` (`Sources/ManifoldContract/GenerationEvent.swift:246`) emitted exactly once orchestrator-side in `GenerationToolDispatchLoop` — consumers get a single in-band 'finished' signal without racing iterator vs. phase. Covered by `GenerationQueueToolLoopTests`.

Issue #1830, opened 2026-06-14
feat(voice): queueing and voice/rate options for SpeechSynthesizing readback


roryford:
**Context:** continuous "read a sequence of items aloud" experiences (a core accessibility affordance) need to enqueue successive strings and control the voice.

`SpeechSynthesizing.speak(_:)` takes a whole string and the Apple implementation cancels any in-flight utterance, so a second call cuts off the first — there is no queue. There is also no control over voice, rate, pitch, or language (rate is hardcoded to default).

**Ask:** add an enqueue mode (append rather than replace) and an options struct (voice id, rate, pitch, locale) to `SpeechSynthesizing`/`VoiceConversationController` to support continuous read-aloud.

roryford:
Shipped in #1838: `SpeechOptions` (voice/rate/pitch/locale) plus `speak(_:options:enqueue:)` on `SpeechSynthesizing` (`Sources/ManifoldVoice/VoiceTypes.swift:99,137`), with a real AVSpeech queue in `AppleSpeechSynthesizer` (`:20-54`). Old `speak(_:)` kept as a source-compatible default. Tests in `SpeechSynthesizingEnqueueTests`/`VoiceConversationControllerTests`.

Issue #1829, opened 2026-06-14
feat(ui): debounced accessibility-announcement utility for streaming UIs


roryford:
**Context:** streaming output needs VoiceOver/`AccessibilityNotification` announcements that fire on completed units, not on every token, or assistive tech is flooded.

MK ships no helper for this, so each host reinvents debounced "announce a completed sentence, coalesce + rate-limit + priority" logic. `ThinkingBlockView` even documents deliberately omitting `.accessibilityValue` to avoid re-announcing streaming reasoning — evidence the problem is already being worked around ad hoc.

**Ask:** a small, platform-agnostic utility in `ManifoldUI` (coalesce + rate-limit + priority) for posting accessibility announcements from streaming UIs, so hosts can announce sentence-level updates consistently.

roryford:
Shipped in #1838: `AccessibilityAnnouncer` (`Sources/ManifoldUI/Accessibility/AccessibilityAnnouncer.swift:61`) — coalesce + rate-limit + priority, injectable post seam, built on the new `SentenceCoalescer`. The `ThinkingBlockView` workaround called out here was migrated in the same PR. Tests in `AccessibilityAnnouncerTests`.

Issue #1828, opened 2026-06-14
feat(contract): emit sentence/segment-boundary events alongside token deltas in GenerationStream


roryford:
**Context:** host apps that read generated text aloud (TTS) or post paced VoiceOver/live-region announcements need to react at sentence granularity, not per token.

Today `GenerationEvent` exposes `.token(String)` as its finest granularity, so every host must re-implement sentence-boundary detection over the token stream (e.g. with `NaturalLanguage`). A repo-wide check shows no sentence segmentation exists in the generation path.

**Ask:** offer an opt-in transform that coalesces tokens into completed units and surfaces them as a `.segment(String, kind:)` event (or a composable `StreamTransform`) behind a config flag, so accessibility/readback layers get a stable "one complete unit is ready" signal without reinventing it.

roryford:
Shipped in #1838 via `SentenceCoalescer` (`Sources/ManifoldContract/SentenceCoalescer.swift`), an opt-in NaturalLanguage-backed transform exposing `coalescedSentences()`. Implemented as an out-of-band utility rather than a new `GenerationEvent` case deliberately — the event enum was frozen for 1.0 in #1826 and sentence segmentation is a derived consumer concern. Tests in `SentenceCoalescerTests`.

PR #1838, opened 2026-06-14
feat!: streaming readback accessibility — completion event, sentence coalescer, voice queue + announcer

roryford:
One coherent feature — **streaming text → spoken/announced output for VoiceOver & TTS accessibility** — implemented across four layers and shipped as a single PR (per repo "one feature = one PR" policy). Built by four parallel workers in isolated worktrees, integrated, and gated together.

## What's in it

| Issue | Layer | Change |
|-------|-------|--------|
| #1832 | `ManifoldContract` / `ManifoldInference` | Terminal `.generationCompleted(GenerationCompletion)` event with `.reason` (`.stop`/`.length`/`.toolIterationLimit`/`.cancelled`/`.error`), emitted **orchestrator-side** so consumers get one in-band "finished" signal without racing the iterator vs `phase == .done`. |
| #1828 | `ManifoldContract` | `SentenceCoalescer` + `AsyncSequence.coalescedSentences()` — opt-in, out-of-band coalescing of the token stream into completed sentences (uses `NaturalLanguage`). **No** new vocabulary case. |
| #1830 | `ManifoldVoice` | `SpeechOptions` (voice id / rate / pitch / locale) + `enqueue:` mode on `SpeechSynthesizing` for continuous read-aloud (real AVSpeech queue instead of cancel-and-replace). Additive + source-compatible. |
| #1829 | `ManifoldUI` | `AccessibilityAnnouncer` — coalesce + rate-limit + priority announcements for streaming UIs, driven by `ingest()` + `finish(reason:)`, with an injectable post seam. Consumes the two Contract pieces. |

## Design decision: the 1.0 vocabulary freeze

`#1826` froze the `GenerationEvent` vocabulary for 1.0. We cracked it **only** for the genuinely-primary completion signal (#1832) — "finished, and why" is foundational provider semantics (`finish_reason`), and pre-1.0 is the only window where adding it is a MINOR bump rather than a permanent major-only change. Sentence segmentation (#1828) is a derived, opt-in, consumer-side concern, so it stays **out** of the frozen enum as a utility. Hence the `feat!:` title (pre-1.0 → MINOR).

The new event is emitted from `GenerationToolDispatchLoop` (orchestrator), so the companion **manifold-mlx / manifold-llama** backend packages need **no changes** — they don't emit it.

## Blast radius / size note

41 files, ~+1230/−219. Above the usual ~40-file / ~800-line split threshold, but ~18 of those files are 1–2 line mechanical `switch` arms forced by adding a frozen-vocab case (the compiler enumerated every exhaustive switch lacking `@unknown default`). It's one feature with one breaking event; splitting would just fragment the switch sweep across PRs.

## Tests & gate

- New focused suites per layer (completion-event emission + ordering, coalescer fragment-split invariance, voice enqueue/options mapping, announcer coalesce/rate-limit/priority), each sabotage-verified then reverted.
- Full local gate (`scripts/test.sh --profile local`, both XCTest + Swift Testing invocations): **4816 tests, 0 failures, RESULT: PASSED.**

Implements #1828, #1829, #1830, #1832 (issues intentionally left open for tracking; not auto-closed).


roryford:
## Adversarial review pass (worker) — fixes applied

A reviewer worker audited the full diff for correctness/races/leaks/contract violations and fixed what it found. Pushed as `0188e40f` + `5368383d`.

**Fixed**
- 🐛 **RACE** — `VoiceConversationController` playback teardown: a replaced/cancelled predecessor task could decrement the *new* utterance's `activeUtterances` and clear `isSpeaking`/`playbackTask` while a fresh utterance was still speaking (latent today since the only public replace path guards with a stop, but a real bug at the `startPlayback` seam). Fixed with a generation token + regression test `test_stalePredecessorTaskDoesNotClearIsSpeakingForNewUtterance` (sabotage-verified). Also added `test_enqueueKeepsSpeakingUntilAllUtterancesFinish` and made the test mock faithfully model AVSpeech replace-cancels semantics.
- 🧹 Removed a dead `pendingHighPriority` field in `AccessibilityAnnouncer` (written, never read — misleading).

**Verified sound (no change needed)**
- `.generationCompleted` is emitted exactly once and last on every exit path (stop / tool-iteration-limit / cooperative cancel / thrown error — emitted before re-throw).
- The 1.0 vocab-freeze tripwire (`BackendSeamConsumer`) still meaningfully asserts the frozen set; it wasn't bumped-to-pass.
- `SentenceCoalescer` flushes its tail on both AsyncSequence termination paths; the per-utterance continuation map in `AppleSpeechSynthesizer` resolves exactly once with no newly-introduced leak.

**Deliberately left**
- External `cancel(_:)` delivers `CancellationError` rather than an in-band `.generationCompleted(.cancelled)` — forcing a yield after the continuation is finished would fight the AsyncStream contract. The cooperative-cancel path *does* deliver it.
- Pre-existing AVSpeech task-cancellation continuation leak (not introduced here; out of scope).
- `NLTokenizer` per-`push` re-tokenization is O(n²) on a bounded buffer — perf nit, not correctness.

Full local gate after fixes: **4818 tests, 0 failures, RESULT: PASSED** (both XCTest + Swift Testing invocations).

PR #1873, opened 2026-06-14
feat!: headless ModelSelection + synchronous #1312 dual-write collapse

roryford:
## What

PR 2 of the reusable model-selection plan (`docs/plans/reusable-model-selection-surfaces.md`). Builds on PR 0 (#1866) and PR 1 (#1864), which already merged.

### Headless `ModelSelection` (ManifoldInference, not ManifoldUI)
Lets consumers choose and load a model **without** a `ChatViewModel`. Composes:
- `ModelRegistry` — selection state + the synchronous `selectModel` entry.
- The `ModelInfo → ModelFitScore` bridge + `DeviceCapabilityService` — the device-fit recommendation, vended as data.
- The shared, service-vended `ModelLoadCoordinator` — one coordinator per `InferenceService`.

Exposes the **sorted / scored / grouped** (foundation vs downloaded) model list as data — that is the product. Auto-load is one synchronous `@MainActor` call. `ModelSelectionSortOrder` + `sortModels()` are hoisted out of `ModelSelectionTabView` onto this type; `ModelSelectionTabView` now delegates to it. A `protocol ModelSelecting` seam lets consumers (offgrid-ai) mock without re-wrapping.

### Synchronous #1312 collapse (Correction F)
`ModelRegistry.selectedModel`'s setter now fires a synchronous `onSelectionChanged` hook, wired by `ChatViewModel` to its endpoint-clear sync. This replaces the async `withObservationTracking` re-install hack that let `dispatchSelectedLoad` read a stale `selectedEndpoint` and dispatch the wrong `LoadIntent`. The public binding setter still syncs endpoints.

### Two-surface no-leak (Correction E)
"One coordinator per service" alone does not stop the single-owner chat callbacks from firing for a headless load. `ModelLoadCoordinator.dispatchLoad` gains a `drivesChatSeams` flag (default `true`; `false` for headless loads). Headless loads suppress the chat-only phase/error/prompt-template callbacks but still fan out progress via the shared `statusUpdates()` stream.

### Docs
New `docs/QUICKSTART-MODEL-SELECTION.md`, registered in `scripts/extract-snippets.sh` (`INPUTS` + `extract_one`) so it does not escape the snippet gate. Covers a worked `ModelLoadPlan.compute()` rejection, a per-family Foundation/local/cloud fork, and the `supportsReasoning` honest-false callout.

## Tests (authored, not just asserted)
- Two-surface topology: headless load does NOT leak into the chat phase (sabotage-verified).
[...]

PR #1874, opened 2026-06-14
feat(inference): auto-inject ToolSystemPromptBuilder for templates that don't render tools natively

### LINKED ISSUE #1856: feat(inference): auto-inject ToolSystemPromptBuilder for templates that don't render tools natively
**Context:** local non-Gemma instruct models receive no tool definitions unless the host hand-injects them.

`GenerationQueue` calls `PromptTemplate.format(…, tools:)` for every prompt-template backend, but all non-`.gemma4` templates (llama3, chatML, mistral, phi, gemma1-3) silently discard `tools` (`PromptTemplate.swift:98-119`). `ToolSystemPromptBuilder.preferTools(for:)` exists and has measured hit-rate gains, but has **zero pipeline call sites** — every host must remember to fold it into the system prompt by hand, and many won't, so tool calls never fire.

**Ask:** when `supportsToolCalling && !config.tools.isEmpty` and the selected template renders no native tool block, have the queue automatically fold `ToolSystemPromptBuilder.preferTools(for:)` into the system prompt.

roryford:
## What

Local non-Gemma instruct models received **no** tool definitions unless the host hand-injected them. `GenerationQueue` passed `config.tools` to `PromptTemplate.format(…:tools:)` for every prompt-template backend, but every non-`.gemma4` template silently discarded them — and `ToolSystemPromptBuilder.preferTools(for:)` (with measured 50% → 70–85% tool-recall gains) had **zero pipeline call sites**.

This wires auto-injection: when `supportsToolCalling && !config.tools.isEmpty` **and** the selected template does not render tools natively, the queue folds `ToolSystemPromptBuilder.preferTools(for:)` (`.standard`) into the system prompt before formatting. `.gemma4` (which renders a native `<|tool>` block) is a no-op for the system-prompt fold, so tools are never double-injected.

## Changes

- **`PromptTemplate.rendersToolsNatively`** — new predicate, `true` only for `.gemma4`. Used as the robust "renders tools natively" signal rather than hardcoding `== .gemma4`, kept in lock-step with the `tools`-consuming branch of `format(…)`.
- **`GenerationQueue.toolAugmentedSystemPrompt(...)`** — folds the preamble into the system prompt (`preamble + "\n\n" + appPrompt`, matching the documented manual recipe). Returns the system prompt unchanged when tools are empty, the backend can't call tools, or the template renders tools natively. Applied on **both** dispatch paths.
- **`GenerationPreflightTrimmer`** now forwards native tools to `format`. This fixes a latent gap: the local `TokenCountingBackend` path (llama.cpp) never passed `tools` at all, so even `.gemma4`'s native block was dropped on that path.
- **docs/LOCAL-TOOL-CALLING.md** — the manual Step-1 fold is now default-on auto-injection; hand-folding documented as the opt-in for `.strict`/`.minimal` styles.

## Tests

`GenerationQueueToolSystemPromptTests` (exercises the local `TokenCountingBackend` dispatch path, asserts on the assembled `lastPrompt`):
- non-Gemma template + tools → preamble + tool name folded into the prompt, host system prompt preserved
- `.gemma4` + tools → preamble **not** double-injected; native `<|tool>` block still present
- empty tools → no injection
- backend without tool-calling support → request rejected, nothing reaches the backend

Sabotage-verified (disabling the fold fails the injection test while gemma4/no-tools stay green), then reverted.

## Premise vs source

The issue cited `PromptTemplate.swift ~98-119` and the non-TokenCounting path. Accurate, but **incomplete**: the path local GGUF models actually take is `GenerationPreflightTrimmer` (TokenCounting), which never forwarded `tools` to `format` at all — so it dropped tools for *every* template including `.gemma4`. Both paths are fixed here.

Resolves #1856

### DIFF
--- a/Sources/ManifoldHardware/PromptTemplate.swift
+++ b/Sources/ManifoldHardware/PromptTemplate.swift
@@ -59,6 +59,24 @@ public enum PromptTemplate: String, CaseIterable, Sendable, Identifiable {
         return result
     }
 
+    /// Whether `format(messages:systemPrompt:tools:)` renders the `tools`
+    /// argument into the prompt itself (a native tool-declaration block).
+    ///
+    /// Only `.gemma4` does — it serialises each tool as a `<|tool>` block in the
+    /// system turn. Every other template silently discards `tools` and produces
+    /// the same output as `format(messages:systemPrompt:)`. Callers driving a
+    /// non-native template are responsible for folding tool descriptions into
+    /// `systemPrompt` (e.g. via `ToolSystemPromptBuilder.preferTools(for:)`);
+    /// `GenerationQueue` does this automatically (#1856). This predicate is the
+    /// signal the queue uses to decide whether that fold is needed — keep it in
+    /// lock-step with the `tools`-consuming branches of `format(…)`.
+    public var rendersToolsNatively: Bool {
+        switch self {
+        case .gemma4: return true
+        case .chatML, .llama3, .mistral, .alpaca, .gemma, .phi: return false
+        }
+    }
+
     /// Thinking markers for this template, or nil if the template does not emit reasoning blocks.
     public var thinkingMarkers: ThinkingMarkers? {
         switch self {
--- a/Sources/ManifoldInference/Services/GenerationPreflightTrimmer.swift
+++ b/Sources/ManifoldInference/Services/GenerationPreflightTrimmer.swift
@@ -49,10 +49,18 @@ struct GenerationPreflightTrimmer {
         var workingMessages = messages
         var attempt = 0
 
+        // Tools that the template renders natively (only `.gemma4` today) must be
+        // forwarded so the native tool block is emitted on this path too —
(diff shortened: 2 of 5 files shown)


roryford:
Orchestrator review (adversarial): **approve.** Traced both prompt paths — the shared `toolAugmentedSystemPrompt` is gated on `supportsToolCalling && !tools.isEmpty && !rendersToolsNatively`, and the trimmer forwards `config.tools` to `format` only for native templates (`[]` otherwise), so non-Gemma gets guidance via the system prompt and Gemma gets its native `<|tool>` block — no double-injection on either path. The `rendersToolsNatively` predicate is an exhaustive switch (forces a compile-time decision on future templates). Good catch on the real bug: the `GenerationPreflightTrimmer` (local TokenCounting path) never passed `tools` to `format` at all, so the fix is broader and more correct than the issue described. Merging on green; sequencing after #1873 (file-disjoint, so trivial).

PR #1875, opened 2026-06-14
feat(contract)!: tool-call parse-failure + truncation diagnostics

### LINKED ISSUE #1857: feat(contract): emit a diagnostic event when a delimited tool-call body fails to parse
**Context:** hosts can't tell "model emitted a broken tool call" from "model emitted no tool call".

`ToolCallTransform` drops a malformed/unknown tool call by returning `nil` from `parseBody`, emitting no event at all (`ToolCallTransform.swift:99-102`; documented "silently drops"). This forces hosts into brittle plain-text fallbacks and silent loss of intent.

**Ask:** add a non-fatal diagnostic — a `GenerationEvent` malformed-tool-call case (or a surfaced parse-failure signal) carrying the raw body — so hosts can recover or report instead of silently losing the call.

### LINKED ISSUE #1858: feat(contract): surface the buffered body of an unterminated tool-call block at finalize
**Context:** a stream cut mid-tool-call vanishes with no content and no signal.

`ToolCallTransform.finalize()` discards an open-but-unclosed tool block including all buffered body text (`:178-187`), so a truncated tool call disappears silently.

**Ask:** provide an opt-in to flush the partial body as `.token` (or a diagnostic) so a mid-tool-call stream truncation is observable rather than lost.

roryford:
## What

Adds two non-fatal `GenerationEvent` diagnostics so hosts can observe tool calls that previously vanished silently inside `ToolCallTransform` (ManifoldContract).

### #1857 — `.toolCallParseFailed(rawBody:)`
When a well-formed open/close marker pair surrounds a body the dialect parser rejects (`parseBody` returns `nil`), the transform now emits a non-fatal diagnostic carrying the raw body **instead of dropping the call with no event**. Hosts can finally distinguish "model emitted a broken tool call" from "model emitted no tool call" and recover/report.

### #1858 — `.toolCallTruncated(rawBody:)` (opt-in)
`ToolCallTransform.finalize()` (and the body-size cap) discarded an open-but-unclosed tool block including all buffered body text, so a truncated tool call disappeared silently. New **opt-in** seam:

```swift
ToolCallTransform(markers: markers, surfaceTruncatedToolBody: true)
```

Default is `false` — **default behavior is unchanged** (still drops silently). When enabled, the partial body is surfaced as a `.toolCallTruncated(rawBody:)` diagnostic so a mid-stream truncation is observable.

Both cases follow the existing `throttleDiagnostic(reason:)` precedent: advisory metadata, `Sendable`/`Equatable` `String` payloads, no chat-message state mutation.

## Freeze interaction

`GenerationEvent` carries a **"Vocabulary freeze (1.0)"** header. We are pre-1.0 (0.51), so completing the vocabulary is appropriate `feat!:` work. Done with full freeze hygiene:
- Freeze-doc header updated to list both new cases as part of the frozen vocabulary.
- Every exhaustive `switch` over `GenerationEvent` across `Sources/` and `Tests/` updated (**12 sites**): `GenerationStreamConsumer`, `EventRecorder`, `ScenarioRunner`, the `APIFreezeTests/BackendSeamConsumer` freeze fixture, and 8 backend/contract test switches (`ToolCallContractTests` ×2, `ParallelToolCallOrderingTests`, `Claude`/`Ollama`/`OpenAI`/`OpenAIResponses` stream-extractor `eventKey` switches, `CloudThinkingTokenTests`, `OpenAIResponsesBackendTests`, `OllamaToolCallLiveReplayTests`). Core-internal switches stay exhaustive (new arms added); none introduced a new `@unknown default` convention.
- `.github/api-breakage-allowlist.txt`: two `enumelement ... has been added as a new enum case` lines (matching the `generationCompleted` precedent) plus a line for the `ToolCallTransform.init(markers:)` signature change (added a defaulted parameter; existing `markers:`-only callers still compile). Digester run locally → **exit 0**.

## Tests

In `OutputParserSessionTests`:
- #1857: a malformed closed tool body now emits `.toolCallParseFailed` with the raw body (was: vanished); a parse failure does not poison a following valid call.
- #1858: opt-in OFF → finalize still drops with no diagnostic (default unchanged); opt-in ON → truncated block surfaces the partial body (incl. a held-back partial close suffix folded in); a fully-closed block emits no truncation.

Sabotage checks were used during development and removed before commit.

## Verification
- `swift build --build-tests` — clean
- `swift build --build-tests --traits Server,Macros` — clean (switched-enum trait sweep)
- `OutputParserSessionTests` slice: 18/18 pass
- API digester (`ManifoldContract`): exit 0 with updated allowlist

BREAKING CHANGE: `GenerationEvent` gains `.toolCallParseFailed(rawBody:)` and `.toolCallTruncated(rawBody:)`; exhaustive switches over `GenerationEvent` without a `default`/`@unknown default` arm must add handling for the new cases.

Resolves #1857
Resolves #1858

### DIFF
--- a/.github/api-breakage-allowlist.txt
+++ b/.github/api-breakage-allowlist.txt
@@ -97,3 +97,6 @@ API breakage: enumelement MessagePart.generatedVideo has been removed
 API breakage: var GBNFSchemaPreValidator.CVEAuditRecord.fixedAtBuild has been removed
 API breakage: var GBNFSchemaPreValidator.CVEAuditRecord.vendoredBuild has been removed
 API breakage: func ModelLoadCoordinator.dispatchLoad(_:) has been renamed to func dispatchLoad(_:drivesChatSeams:)
+API breakage: enumelement GenerationEvent.toolCallParseFailed has been added as a new enum case
+API breakage: enumelement GenerationEvent.toolCallTruncated has been added as a new enum case
+API breakage: constructor ToolCallTransform.init(markers:) has been removed
--- a/Sources/ManifoldContract/GenerationEvent.swift
+++ b/Sources/ManifoldContract/GenerationEvent.swift
@@ -14,6 +14,15 @@
 /// a major (`feat!:`) release. Public-facing / cross-module consumers should add
 /// an `@unknown default:` arm to stay resilient to a future major.
 ///
+/// The two non-fatal tool-call diagnostics —
+/// ``toolCallParseFailed(rawBody:)`` and ``toolCallTruncated(rawBody:)`` — are
+/// part of this frozen vocabulary. They were the last pre-1.0 additions
+/// (`feat!:`, #1857 / #1858): a delimited tool-call body that fails to parse,
+/// and an unterminated tool block surfaced at finalize, are now observable
+/// instead of being silently dropped. Both follow the
+/// ``throttleDiagnostic(reason:)`` precedent — advisory metadata with no
+/// chat-message state mutation.
+///
 /// Payloads that are expected to grow are modelled as **associated structs**
 /// rather than bare enum parameters so their fields can grow non-breakingly
 /// after the freeze:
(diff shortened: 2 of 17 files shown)


roryford:
Orchestrator review (adversarial): **approve.** Freeze handling is correct — both new cases (`toolCallParseFailed`/`toolCallTruncated`) follow the `throttleDiagnostic` precedent, the freeze-doc header is updated to list them, and the allowlist has bare `API breakage:` lines (no `#` comments) incl. the source-compatible `init(markers:)` defaulted-param signature change. The switch-site sweep is complete *by compiler-truth* — both `swift build --build-tests` and `--traits Server,Macros` compile, and Swift enforces exhaustiveness, so a missed site couldn't build. #1857 is always-on (per the issue); #1858 is opt-in via `surfaceTruncatedToolBody` (default false → silent-discard unchanged), gating both the body-cap drop and the `finalize()` flush. Good catch that the brief's `GenerationEventClosedAuditTest` premise was wrong (it's ConversationEvent/RunEvent disjointness) — the real freeze fixture `BackendSeamConsumer.swift` was updated instead. Merging on green.

PR #1877, opened 2026-06-14
feat: public ModelPicker sample over ModelSelection + attachImage/togglePin gap closers

### LINKED ISSUE #1298: [API] Extend ModelCapabilities with codeGeneration and multilingual axes
## Use case

Consumer apps that present a model browser with at-a-glance capability badges (\"Vision\", \"Code\", \"Reasoning\", \"Multilingual\") need a canonical capability set inferred from the model itself — not hardcoded per backend or guessed from the model's display name. Today `ModelCapabilities` (`Sources/ManifoldInference/Services/ModelCapabilityProbe.swift:10`) covers `supportsVision` and `supportsAudio`, and `BackendCapabilities.supportsThinking` covers reasoning, but two common axes are missing:

- **Code generation** — distinguishes general chat models from coder-tuned variants (Qwen-Coder, DeepSeek-Coder, etc.). Consumers want to badge these and route code-input prompts to them when present.
- **Multilingual** — distinguishes English-only checkpoints from broadly multilingual ones. Drives both UI badges and prompt-template defaults.

Inferring these per-consumer from filename heuristics is unreliable and drifts.

## Proposed API

```swift
public extension ModelCapabilities {
    /// `true` when the model card or config flags this as a code-specialised
    /// checkpoint (e.g. \"coder\" / \"code\" architecture tags, code-specific
    /// tokenizer vocabularies).
    var supportsCodeGeneration: Bool { get }

    /// `true` when the model card declares ≥ 2 supported natural languages or
    /// the tokenizer vocabulary indicates multilingual training.
    var supportsMultilingual: Bool { get }
}
```

Inferred by `ModelCapabilityProbe.probe(modelDirectory:)` from `config.json` / model card metadata (HF \"language\" tag, architecture name, README front-matter where available).

## Why this belongs in ManifoldKit

- The signal is in MK's existing probe pipeline (`config.json` + tokenizer config). Consumers cannot replicate it without re-reading the same files.
- Every consumer building a model browser will want the same axes; codifying them in MK prevents per-app filename-regex drift.

## Notes

- `BackendCapabilities` (`Sources/ManifoldInference/Protocols/BackendCapabilities.swift:41`) already has `supportsVision` and `supportsThinking`; the gap is in `ModelCapabilities` (`Sources/ManifoldInference/Services/ModelCapabilityProbe.swift:10`) and the projection of those onto a per-model badge set.
- For Apple Foundation, the capability surface evolves between OS releases; `FoundationBackend.capabilities` (`Sources/ManifoldFoundation/FoundationBackend.swift:112`) should report whichever axes the current OS supports rather than hardcoded.
- Backwards-compatible: new fields default to `false`; no existing callers break.

### LINKED ISSUE #1300: [API] Bulk session deletion on SessionManagerViewModel
## Use case

Consumer apps shipping a privacy or data-management surface (Settings → "Erase All Chats", account-reset flows, GDPR-style purge) need an atomic, batch-efficient way to delete every persisted session. Today the only public path is iterating `sessions` and calling `deleteSession(_:)` N times. That pattern is O(N) round-trips through `SessionListService`, emits N `sessionDeleted` events the SwiftUI list will animate one-by-one, and races mid-iteration mutations (a new session created from another scene-phase path can survive the loop).

## Proposed API

```swift
extension SessionManagerViewModel {
    /// Deletes every persisted session and its messages in a single atomic
    /// pass. Emits one terminal event (`.allSessionsDeleted` or a final
    /// `.sessionsLoaded(records: [], …)`) rather than N `.sessionDeleted`.
    public func deleteAllSessions() async throws
}
```

Implementation can lower to a single `SessionStore.deleteAll()` call inside `SessionListService` so it's one transaction at the persistence layer.

## Why this belongs in ManifoldKit

- The atomicity guarantee can only be made by the layer that owns the persistence transaction — consumers can't synthesise it from `deleteSession(_:)` calls without races.
- Every consumer building a "wipe data" affordance needs the same primitive; today each invents a partial workaround.
- The single-event shape lets SwiftUI lists collapse cleanly instead of animating N row removals.

## Notes

- `SessionManagerViewModel.deleteSession(_:)` lives at `Sources/ManifoldUI/ViewModels/SessionManagerViewModel.swift:213` and delegates to `SessionListService.deleteSession(_:)`.
- Naturally pairs with the existing `clearSearch()` reset semantics.

roryford:
PR 3 of the reusable model-selection plan (`docs/plans/reusable-model-selection-surfaces.md`). Builds on PR 0 (#1866), PR 1 (#1864), PR 2 (#1873). **Draft** — do not merge.

## 1. Public `ModelPicker` sample over `ModelSelection`
- Promotes the private `ModelSelectionTabView` → public `ModelPicker` in `ManifoldUIModelManagement`: a thin, styleable **sample** selector. The headless `ModelSelection` stays the product; `ModelPicker` is the default, not the only path.
- Adds a grouping seam (`grouped: Bool`) that renders the foundation-vs-downloaded sections from `ModelSelection.groupModels(_:by:)`. Defaults to `false` (the historical flat list).
- Hoists the grouping into a static `ModelSelection.groupModels(_:by:)` so the sample view and the headless instance method share one implementation — no `ModelLoadCoordinator` needed to render. `ModelPicker` drives selection through a `ModelRegistry` it already owns (no service widening; respects the service-sharing rule).

## 2. `ModelManagementSheet` composes `ModelPicker`
- Re-expressed to use `ModelPicker(grouped: false)` in the Select tab — **no behavior change**. Existing logic/snapshot/explicit-init tests updated to the new name.

## 3. Gap closers
- `ChatViewModel.attachImage(_:mimeType:)` (#1298) — convenience over `stageAttachment` that builds the `MessagePart.image` and routes through the same internal path the bundled composer uses (placeholder-hash parity verified).
- `SessionManagerViewModel.togglePin(_:)` (#1300) — pins/unpins based on the record's `isPinned`.

## 4. Docs snippet-gate fix (Correction G / Decision 6)
- Adds `ManifoldUIModelManagement` as a **third linked product** in `scripts/extract-snippets-test.sh` (it is not umbrella-re-exported) so `ModelPicker` doc snippets compile.
- New `ManifoldUIModelManagement.docc/ModelPickerSample.md` article with three **compile-checked** `ModelPicker` snippets (not `no-build`, not under a blocked heading). Links into the existing `docs/QUICKSTART-MODEL-SELECTION.md` (owned by PR 2) via prose rather than editing it.

## Validation (local)
- `swift build --build-tests` — clean.
- New slices green (`ModelPickerTests`, `attachImage`, `togglePin`); sabotage-checked then reverted.
- Snippet gate: `extract-snippets.sh` + `extract-snippets-test.sh` → **16 passed, 0 failed**, including all three ModelPicker snippets, under the new 3-product scaffold. `check-readme.sh` passes.
- API digester clean on `ManifoldInference` / `ManifoldUI` / `ManifoldUIModelManagement` — all additions are additive (no allowlist entry needed).

Resolves #1298
Resolves #1300


roryford:
Orchestrator review (adversarial): **approve.** `groupModels` hoist is a clean single-implementation refactor (instance `groupedModels` forwards to the new static), and the sample view correctly drives selection through its own `ModelRegistry` rather than the service coordinator — no service-sharing-rule violation. `attachImage` delegates to the same internal `stageDraftAttachment` path as ChatInputBar (not a parallel path); `togglePin` reads `isPinned` at call time over the existing idempotent pin/unpin. Snapshot test is a mechanical view-rename, not a masked regression. Digester clean (all additive). Disjoint from #1874/#1875 — merging on green.

PR #1885, opened 2026-06-15
feat(runtime): batteries-included compression policies (truncating / extractive / anchored)

roryford:
## Summary

MK already ships the `CompressionPolicy` / `PreTurnCompressionPolicy` *seams* (wired into `ConversationRuntime` via `TurnCompressionCoordinator`), but **no concrete strategy** — the protocol docs show a hand-wave `compress()` stub and every consumer had to write their own. This PR adds three batteries-included strategies + one policy wrapper.

Ported and rewritten from Fireside's `StoryCompression` subsystem, adapted from its internal `Message`/`Result` tuples to MK's `[ChatMessage]` seam (`generate` is a parameter, not stored → clean `Sendable` structs, no `@MainActor`/`@unchecked`).

## Strategies (a ladder)

- **`TruncatingCompressionStrategy`** — zero-inference sliding window (keep system + newest-by-budget tail, drop oldest). The canonical baseline. *(new — not in Fireside)*
- **`ExtractiveCompressionStrategy`** — zero-inference scored selection (recency / length / keyword-density), verbatim tail, greedy within budget. Adds an optional **`headBudgetFraction`** knob to pin establishing context (middle-out / lost-in-the-middle).
- **`AnchoredCompressionStrategy`** — summarize old messages via `generate`, prepend a **`.memory("summary")`** record + keep a verbatim recency tail. Keeps Fireside's `summarizerInputWindow` input-sizing decoupling, chunk-and-fold for over-window input, summary-floor logic, `CancellationError` early-return, and **fall-back-to-extractive** on failure/empty/missing-generate.

## Policy wrapper

`DefaultCompressionPolicy` is a `Sendable` struct conforming to **both** `CompressionPolicy` (post-turn) and `PreTurnCompressionPolicy` (pre-turn). The trigger asymmetry is handled internally — the post-turn path uses `contextUtilization`; the pre-turn path receives only `messageCount`/`lastPromptTokens`, so the policy stores `contextSize` as config to compute utilization.

```swift
runtime.compressionPolicy        = .extractive(threshold: 0.75)
runtime.preTurnCompressionPolicy = .anchored(threshold: 0.85, contextSize: 8192)
```

## Tests

`Tests/ManifoldRuntimeTests/DefaultCompressionPolicyTests.swift` — ported from Fireside's `StoryCompressionTests` + P0 extractive-edge-case / anchored-fallback / summarizer-starvation suites, rewritten against `[ChatMessage]`, plus new truncating-strategy, `headBudgetFraction`, and pre-turn-vs-post-turn trigger tests.

## Out of scope (deliberately)

- Recursive / hierarchical (summary-of-summaries) summarization
- Embedding / semantic-relevance selection (belongs with RAG)
- KV-cache eviction (token/attention layer, not message history)
- The graph reconciler `postCompress` hook (left as the default no-op for consumers to fill)

## Notes for review

- Verify the `isPinned` resolution against `ChatMessage` (Fireside's `Message` had `isPinned`; check whether MK exposes a pin/kind equivalent or whether v1 treats none-pinned as acceptable with a TODO).
- Draft pending CI + a review-and-fix pass.

### DIFF
--- a/Sources/ManifoldInference/Services/ContextWindowManager.swift
+++ b/Sources/ManifoldInference/Services/ContextWindowManager.swift
@@ -71,6 +71,60 @@ public enum ContextWindowManager {
         return HeuristicTokenizer().tokenCount(text)
     }
 
+    /// Fixed per-part token cost for a non-text modality whose true token
+    /// footprint isn't visible from the `[ChatMessage]` value alone.
+    ///
+    /// `.content` only exposes `.text` parts, so an image/audio part would
+    /// otherwise estimate as **zero** — a multimodal turn could then sail past
+    /// the compression trigger and overflow at generation. These are coarse
+    /// placeholders, not provider-accurate counts (a single image is ~85–1.5k
+    /// tokens depending on resolution / provider), deliberately sized to be
+    /// "clearly non-zero, never catastrophically wrong": they make the budget
+    /// math notice the part exists. A backend that knows its real tiling cost
+    /// should account for that at the wire layer, not here.
+    static let imagePartTokenEstimate = 768
+    static let audioPartTokenEstimate = 384
+    static let generatedMediaPartTokenEstimate = 256
+
+    /// Estimates the token cost of a single `MessagePart`.
+    ///
+    /// - `.text` / `.thinking`: counted via the tokenizer (heuristic when nil).
+    /// - `.image` / `.audio` / `.generatedMedia`: a documented fixed estimate
+    ///   (``imagePartTokenEstimate`` etc.) since the byte payload's true token
+    ///   footprint isn't derivable here.
+    /// - `.toolCall`: the tool name plus its serialized JSON arguments.
+    /// - `.toolResult`: the serialized result payload (content + dialog).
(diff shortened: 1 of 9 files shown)


roryford:
Review-pass note (from the adversarial review + fix pass, commit `477830ba`):

- **Deliberate divergence from the Fireside original:** `ExtractiveCompressionStrategy` drops Fireside's explicit `candidateBudget = budget - tailBudget` cap and fills candidates to the full history `budget`. This is intentional — strictly better budget utilization, still bounded, never overflows context. Not a port bug.
- **`isPinned`:** no per-message pin equivalent exists on the `[ChatMessage]` seam (pins live on `ChatSession.pinnedMessageIDs`, invisible here). The port preserves `.system` + `.memory` records verbatim as 'load-bearing' and leaves a TODO to honor a future pinned-IDs channel.
- **Known coverage gap (not a bug):** the anchored chunk-and-fold over-window path (`summarizerInputWindow` set + input exceeds it) has no direct test.
- Local gate: `scripts/test.sh --profile local` → 4932 tests, 0 failures, 15 skipped, clean first run.

Issue #1834, opened 2026-06-14
Pre-1.0 Contract API hardening — deferred items from adversarial review


roryford:
Tracking issue for Contract-API findings from an adversarial review (Jun 2026) that were **not** fixed in the first hardening pass. The three highest-severity, freeze-irreversible items (`JSONSchemaValue` integer fidelity, `ToolResult.ErrorKind` escape hatch, `BackendCapabilities` tolerant decode) are handled in the `feat!: harden Contract wire types before 1.0 freeze` PR. The items below survived red-team scrutiny but were judged lower-severity or design-trade decisions.

Each item notes whether it is **freeze-blocking** (must land before the 1.0 vocabulary freeze because it can't be fixed non-breakingly after) or **deferrable** (can ship post-1.0).

### Freeze-relevant (decide before 1.0)
- [ ] **`EmbeddingBackend` is thinly specified vs `InferenceBackend`** (#15). No `capabilities`, no thread-safety doc, `dimensions` undefined before `loadModel`, and `embed(_:)` has no stated guarantee that output vector count matches input count. `EmbeddingError.dimensionMismatch` partially mitigates. For a third-party-implementable protocol this contract is under-defined. *Freeze-blocking if we want to add `capabilities` to the protocol later.*
- [ ] **Document the throw-vs-silent-ignore rule in `GenerationConfig`** (#1). The rule is principled (capability-gated *guarantees* like `grammar` throw `unsupportedGrammar`; advisory hints like `seed`/`minP`/`jsonMode` degrade silently) but is only discoverable by reading individual field docs. State the rule once at the type level. *Deferrable (doc-only).*
- [ ] **Vendor-specific sampler knobs on the shared `GenerationConfig`** (#2). `llamaDRY`/`llamaXTC`/`llamaMirostatV2` (llama.cpp-only), `yieldEveryNTokens` (MLX-only), `streamPrefillProgress` (OpenAI-compat-only) are frozen into the "shared across all backends" type. Consider a `backendSpecific` side-channel before freeze. *Freeze-blocking — can't relocate frozen public fields after 1.0.*

### Deferrable (post-1.0 safe)
[...]

Issue #15, opened 2026-03-30
fix: stable reverse-scroll when prepending older messages


roryford:
## Summary

Fix the scroll-jump problem in the chat message list when loading older messages at the top of the conversation.

## Problem

SwiftUI's `List` and `ScrollView` don't handle prepending content gracefully — when older messages are inserted above the current viewport, the scroll position jumps because the content size changes without a corresponding offset adjustment. This is a well-known SwiftUI limitation that affects any chat UI with "load more" or infinite scroll.

## Expected behavior

When the user scrolls to the top and older messages load, the viewport should remain anchored to the same message the user was looking at. No visible jump.

## Prior art

- [FluidGroup/swiftui-messaging-ui](https://github.com/FluidGroup/swiftui-messaging-ui) (40 stars, Apache 2.0) — a focused library that solves exactly this problem with stable prepending in SwiftUI. Worth studying their approach.
- UIKit's `UICollectionView` handles this natively via `contentOffset` adjustment — SwiftUI has no equivalent.

## Possible approaches

1. **Anchor-based scroll restoration** — record the ID/offset of the first visible message before prepend, then scroll back to it after insertion
2. **Bottom-anchored layout** — render the list from the bottom up (newest first), so prepending is actually appending from the layout's perspective
3. **Adopt FluidGroup's technique** — they use a geometry-reader approach to measure and compensate for content size changes
4. **Wait for SwiftUI improvements** — Apple may address this in a future release, but no indication yet

## Impact

This is a quality-of-life issue for any app with long conversations. Not a blocker, but noticeable and a polish differentiator.

PR #1901, opened 2026-06-18
feat: harden EmbeddingBackend documented contract for 1.0 freeze

roryford:
Worker B on #1834 (pre-1.0 Contract API hardening), item #15: `EmbeddingBackend` was thinly specified vs `InferenceBackend`. This brings its documented contract up to the same level so it is safe to **freeze for 1.0**.

## What changed (doc-only, `Sources/ManifoldContract/EmbeddingBackend.swift`)

- **Top-level protocol doc + `## Thread Safety` section**, modeled on `InferenceBackend`. States that conformers are `Sendable`, methods may be called from **any** thread, and — unlike `InferenceBackend` — there is **no** service-level serialization (RAG indexes batches in parallel, and `loadModel`/`unloadModel` may arrive from a different actor concurrently with `embed`). Mutable-state conformers must self-synchronize; existing conformers (`NLEmbeddingBackend`, `OllamaEmbeddingBackend`) hold immutable/thread-safe state under `@unchecked Sendable`.
- **`dimensions` precondition** stated explicitly: undefined until `isModelLoaded == true`; value before load is backend-defined and carries no contract.
- **`embed(_:)` count/length guarantee** made explicit: on non-throwing return, `result.count == texts.count` in input order, each vector of length `dimensions` (callers may zip without a length check). A backend that cannot satisfy this does not return a short/ragged array — it throws `EmbeddingError.dimensionMismatch` (the noted partial mitigation).
- **`EmbeddingError` cases documented.**

No behavior change. Existing `EmbeddingBackendProtocolTests` (7 tests) already cover the postcondition, capabilities default, and error descriptions — no new test needed for a doc-only change.

## Freeze relevance / capabilities decision: **DONE (already on main)**

The freeze-critical question was whether to add a `capabilities` member now (un-addable non-breakingly to a third-party-implementable protocol after 1.0). **This was already closed on `origin/main`**: `EmbeddingCapabilities` (batch/input bounds + normalization) plus a protocol-extension `default` so existing conformers — including the companion `manifold-mlx` / `manifold-llama` embedders — don't break. So this PR needed **no** API break; it's the remaining doc-hardening that the capabilities work didn't cover.

## Gate

- `swift build` — Build complete (167s)
- `swift test --filter EmbeddingBackendProtocolTests` — 7 tests, 0 failures

Scoped check only; full CI is the orchestrator's job.

### DIFF
--- a/Sources/ManifoldContract/EmbeddingBackend.swift
+++ b/Sources/ManifoldContract/EmbeddingBackend.swift
@@ -32,15 +32,37 @@ public struct EmbeddingCapabilities: Sendable, Codable, Equatable, Hashable {
     public static let `default` = EmbeddingCapabilities()
 }
 
+/// Common interface for text-embedding backends.
+///
+/// Each backend wraps a different sentence-embedding engine (Apple
+/// `NaturalLanguage`, an on-device MLX/llama embedder, a remote Ollama
+/// embedder, …) and exposes the same async API. RAG retrieval delegates all
+/// embedding work here through the injected conformer; when none is supplied,
+/// `ManifoldBootstrap` resolves the bundled `NLEmbeddingBackend`.
+///
+/// ## Thread Safety
+///
+/// Conformers are `Sendable` and their methods may be called from **any**
+/// thread. There is no service-level serialization analogous to
+/// `InferenceBackend`'s generation queue: a host may invoke ``embed(_:)``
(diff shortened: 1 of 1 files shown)


roryford:
Folded into #1902 to avoid a separate single-file cold CI run (per repo CI-hygiene). The EmbeddingBackend doc commit was cherry-picked onto `feat/1834-contract-polish` — see #1902.

Issue #1937, opened 2026-06-20
test: first-party RAG eval harness (faithfulness / relevancy / hit-rate / MRR)


roryford:
**Cluster:** RAG quality (testing) · **Impact:** High · **Effort:** Med

A first-party RAG evaluation harness: faithfulness/grounding, answer/context relevancy, retrieval hit-rate/MRR, plus a fixture corpus + golden queries asserting recall@k and citation correctness across chunking/rerank configs. Optionally synthetic Q&A dataset generation for regression baselines.

**Why:** LlamaIndex's first-party evaluators are a major differentiator. ManifoldKit ships RAG but has no way to *measure* it, so the RAG-quality features above can't be tuned or defended.

---
_Filed from the June 2026 competitive-research sweep (4 parallel clusters: Swift-native ecosystem, local-UI/RAG/voice, agent/MCP/durable-execution, cross-platform runtimes). Verify the cited primitive in `Sources/` before implementing — workers confirmed the gaps but did not exhaustively trace existing depth._

---

## 🔬 Investigation & implementation plan (2026-06-20)

**VERDICT: Proceed — and this gates #1919/#1920/#1921.** Premise confirmed: no RAG eval harness exists (zero hits for recall@k/MRR/hit-rate/faithfulness across `Tests/`). This is the keystone of the cluster — the quality features above are untunable and indefensible without it.

### A. Ground truth
- **No eval harness.** Grep for `recall@`, `hitRate`, `MRR`, `faithfulness` across `Tests/` → nothing. RAG tests today assert *mechanics* (`RAGServiceTests`, `DocumentChunkerTests`, `RAGServiceRerankIntegrationTests`, `FlatFileVectorStoreTests`, `NLEmbeddingRAGIntegrationTests`) but never *retrieval quality*.
- **Reusable substrate already exists**: `RAGService.retrieve` returns scored `[VectorSearchHit]` + `[Citation]` (`Sources/ManifoldRuntime/Services/RAGService.swift` L156–247); `ManifoldBootstrap.makeInMemory` + `GlassBoxDemoRAG` (`Sources/ManifoldTestSupport/GlassBoxDemoRAG.swift` L54–59) already ingest a corpus into in-memory stores — the harness can build on this.
[...]

PR #1978, opened 2026-06-20
feat(rag): hybrid BM25 + dense retrieval with Reciprocal Rank Fusion (#1919)

### LINKED ISSUE #1919: feat: hybrid retrieval (BM25 + dense) with Reciprocal Rank Fusion
**Cluster:** RAG quality · **Impact:** High · **Effort:** Med

Add a sparse (BM25) retriever alongside the existing dense path and fuse with Reciprocal Rank Fusion.

**Why:** Open WebUI, AnythingLLM "Accuracy Optimized". Dense-only misses codes/rare terms/exact tokens; RRF is the scale-free standard. Current RAGService stops at dense + top-k + rerank.

---
_Filed from the June 2026 competitive-research sweep (4 parallel clusters: Swift-native ecosystem, local-UI/RAG/voice, agent/MCP/durable-execution, cross-platform runtimes). Verify the cited primitive in `Sources/` before implementing — workers confirmed the gaps but did not exhaustively trace existing depth._

---

## 🔬 Investigation & implementation plan (2026-06-20)

**VERDICT: Proceed (descoped + reordered).** The gap is real — keyword search is naive substring matching, never *fused* with dense — but ship behind eval harness #1937, which must land first to defend the RRF tuning.

### A. Ground truth
- Dense path is real cosine over a flat file: `FlatFileVectorStore.search` L66–92 (`Sources/ManifoldPersistenceSwiftData/Adapters/FlatFileVectorStore.swift`), L2-normalized dot product, O(n).
- The "sparse" path is **not BM25** — it's `keywordSearch` L94–114: `$0.text.lowercased().contains(lower)` with a hard-coded `score = 1.0`. No tokenization, no term weighting, no IDF.
- Dense and keyword are **mutually exclusive fallback**, never fused: `RAGService.retrieve` L184–202 uses keyword *only* when the embedding model is absent or the embed call throws. There is no RRF and no path where both run. The issue's "Current RAGService stops at dense + top-k + rerank" is accurate — there is no sparse-fusion stage.
- `VectorStore` protocol (`Sources/ManifoldRuntime/Protocols/VectorStore.swift`) has `search` + `keywordSearch` as separate methods returning `[VectorSearchHit]`; both already carry per-hit scores RRF can rank-fuse.

### B. Adversarial challenge
- **Premise correct**, but the framing "BM25 + dense" oversells the needed lift. A true BM25 index (term-frequency postings, IDF, length-normalization) over a flat-file store is real complexity for a corpus the docstring itself caps at ~50k chunks (`FlatFileVectorStore` L9). **Recommendation: descope BM25 → a real tokenized term-overlap scorer** (lowercased token set intersection + IDF-lite over the in-memory record set) and reserve full BM25 for the sqlite-vec/ANN milestone ("Phase 3" per CitationsView preview text). The scale-free win of #1919 is the *RRF fusion of two rankings*, not BM25 specifically.
- **Hidden complexity**: BM25 needs a persisted inverted index in SwiftData (postings + doc-frequency table) to avoid recomputing IDF on every query; that's a schema migration + its own correctness surface. The descoped in-memory scorer avoids the migration entirely since `FlatFileVectorStore` already loads all records into memory.
- **Worth it vs simpler path**: RRF over (cosine ranking, term-overlap ranking) is ~30 lines and is the documented standard. Full BM25 is a separate, larger issue — split it out only if eval (#1937) shows term-overlap is insufficient.

### C. Implementation
**Approach**: add a fusion stage in `RAGService.retrieve` that, when the embedding backend is ready, *also* runs the sparse ranker and fuses both rankings with RRF before the rerank stage.

Files/types:
- `Sources/ManifoldRuntime/Services/RAGService.swift` — new private `fuse(dense:sparse:limit:)` using RRF: `score(d) = Σ 1/(k + rank_i(d))`, `k = 60` (standard). Run sparse search unconditionally when dense is active (currently it's fallback-only). Widen both pools to `candidateLimit` first, fuse, then hand the fused top-`candidateLimit` to the existing reranker stage (L210–220) — RRF slots cleanly *before* rerank.
- `FlatFileVectorStore.keywordSearch` L94–114 — replace `.contains` with tokenized scoring: lowercase + split on non-alphanumerics into a term set, score each record by `Σ idf(term)` over matched terms where `idf(t)=log(N/df(t))`, df computed once per `ensureLoaded()`. Return real scores instead of `1.0`. This is also a standalone quality win for the no-embedding fallback.
- New `RetrievalStrategy` enum on `RAGService.init` (`.dense`, `.sparse`, `.hybrid`) defaulting to `.hybrid` when both are available — keeps the dense-only and keyword-only paths byte-compatible for adopters who want them.

API sketch:
```swift
public enum RetrievalStrategy: Sendable { case dense, sparse, hybrid }
// RAGService.init(..., retrievalStrategy: RetrievalStrategy = .hybrid)
// private func reciprocalRankFusion(_ rankings: [[VectorSearchHit]], k: Int = 60, limit: Int) -> [VectorSearchHit]
```
Fusion identity by `chunk.id`; carry forward the higher source score into the fused `VectorSearchHit.score` so citations stay meaningful.

**Test plan** (`Tests/ManifoldRuntimeTests/RAGServiceTests.swift` + `FlatFileVectorStoreTests.swift`, XCTest, in-memory stores, no `try?`):
- `test_hybrid_fusesRankings_rrfOrdersByReciprocalRank` — construct dense and sparse rankings that *disagree*, assert RRF order matches hand-computed `Σ1/(60+rank)`. Sabotage: force k=0 / single-source → order changes → confirms fusion is live.
- `test_keywordSearch_idfWeighted_raresTermsOutrankCommon` — corpus where a rare term and a common term both match; assert the rare-term doc ranks higher (impossible under the old `1.0` constant). Sabotage: revert to constant 1.0 → assertion fails.
- `test_strategy_dense_and_sparse_preserveLegacyBehaviour` — `.dense`/`.sparse` reproduce pre-fusion ordering byte-for-byte.
- Fusion must run *before* rerank: add a `ClosureReranker` (already in `RAGServiceRerankIntegrationTests`) asserting it receives the fused pool.

**Re-estimated effort**: Med (descoped). ~1 PR, ~250 LOC. Full BM25 + inverted index would be L (separate issue).

### D. Ordering / dependencies
- **#1937 (eval harness) should land first** — RRF k, strategy default, and term-overlap-vs-BM25 are all tuning decisions that need recall@k/MRR numbers to defend. Without eval this is unfalsifiable.
- Independent of #1920 (rerank) and #1921 (citation UI); RRF slots before the existing rerank stage with no interface change.

### Open questions
- Default `k=60` or expose it? (Recommend constant; expose only if eval demands.)
- Should `.hybrid` be the default for *existing* adopters, or opt-in to avoid changing their retrieval silently? (Lean: default `.hybrid` but gate behind a bootstrap flag for one release.)
- Do we want full BM25 at all pre-sqlite-vec, or is IDF-weighted term overlap the permanent answer below ANN scale?

roryford:
## Summary

Adds a sparse **BM25** retriever alongside the existing dense cosine path and fuses the two ranked lists with **Reciprocal Rank Fusion (RRF)** before the existing top-k + rerank stages. The entire hybrid path is gated behind a **new default-OFF flag** — when off, retrieval is byte-identical to today's dense-or-keyword-fallback pipeline.

Closes #1919.

## What changed

- **`BM25Scorer`** (`Sources/ManifoldRuntime/Services/BM25Scorer.swift`) — pure, value-type BM25 scorer: tokenize (lowercase + split on non-alphanumerics), term frequency, Robertson/Spärck-Jones IDF (`+1` smoothing so common terms never go negative), standard length-normalized saturation. `k1 = 1.2`, `b = 0.75`.
- **`ReciprocalRankFusion`** (`Sources/ManifoldRuntime/Services/ReciprocalRankFusion.swift`) — pure RRF stage: `score(d) = Σ 1/(k + rank_i(d))`, `k = 60`. Fuses by `chunk.id`; carries the higher source score into the fused hit so citations stay meaningful.
- **`VectorStore.bm25Search`** (`Sources/ManifoldRuntime/Protocols/VectorStore.swift`) — new protocol method with a **default impl delegating to `keywordSearch`**, so existing conformances keep compiling unchanged. `FlatFileVectorStore` overrides it with the real BM25 index.
- **`RetrievalStrategy { dense, sparse, hybrid }`** on `RAGService` (default `.dense`). `.hybrid` runs the dense and BM25 legs, RRF-fuses them, and hands the fused top-`candidateLimit` into the **existing rerank stage unchanged** (RRF slots cleanly before rerank). The dense leg keeps its keyword fallback, so a missing/throwing embedding model degrades hybrid → (BM25 ⊕ keyword) rather than failing the turn.
- **`RAGConfiguration.hybridRetrieval`** (default `false`) in `ManifoldBootstrap` maps to `.hybrid` / `.dense`.

## Flag

- **Name:** `RAGConfiguration.hybridRetrieval` — **default `false`** (OFF).
- Off → `RAGService(retrievalStrategy: .dense)` → behaviour byte-identical to today. Confirmed by the existing 31 `RAGServiceTests` passing unchanged plus a dedicated flag-off integration test.

## Tuning constants are PROVISIONAL — held for #1937

The RRF / BM25 constants (`k = 60`, `k1 = 1.2`, `b = 0.75`) are the published defaults but are **unvalidated for this corpus** without the RAG eval harness (**#1937**, not yet merged). Because this PR is default-OFF and non-regressing it is safe to land, but **please hold it for review pending #1937 — do not auto-merge.** Recall@k / MRR numbers from the harness should defend (or retune) the constants before hybrid becomes a recommended default.

## Tests

- **Pure unit** — `BM25ScorerTests` (8): tokenization, IDF rare>common, exact ranking on a known corpus, tf saturation, zero-match drop, df==N positivity. `ReciprocalRankFusionTests` (6): hand-computed `Σ 1/(k+rank)` fused order, both-lists doc outranks singletons, higher-score-carried, single-list pass-through, k-sensitivity sabotage.
- **Store unit** — `FlatFileVectorStoreTests` BM25 cases: rare-term ranks first with distinct real scores (not the legacy constant `1.0`).
- **Integration (in-memory, real `FlatFileVectorStore` + real `SwiftDataDocumentStore`, no mocked persistence)** — `HybridRetrievalIntegrationTests` (3): with the flag ON a rare exact token that dense ranks last is surfaced via BM25+RRF within top-2; with the flag OFF the same query/corpus misses it (contrast proof); flag-off dense ordering preserved.

All 29 new tests + 31 existing `RAGServiceTests` green locally.

## Note for reviewers

Making `FlatFileVectorStore.bm25Search` `async throws` (not sync `throws`) is load-bearing: with a protocol-extension default present, a sync actor witness was shadowed by the default — the `async` signature binds the concrete BM25 implementation.


roryford:
### Follow-up for #1937 (eval-harness tuning pass): BM25-score vs cosine-threshold unit mismatch

Flagging a latent unit mismatch to resolve **before hybrid retrieval is turned on by default**:

- Under the opt-in `.hybrid` / `.sparse` strategies, `hit.score` carries **raw, unbounded BM25** scores.
- `similarityThreshold` (added in #1947) is a **cosine floor in `[0, 1]`** and is applied against `hit.score` at `RAGService.swift:302`.

So a non-zero `similarityThreshold` combined with a hybrid/sparse strategy compares **mismatched units** — a cosine-style cutoff against unbounded BM25 magnitudes, which would filter unpredictably.

**Why it's dormant today (not a default-path bug):** the default `similarityThreshold` is `0` (no-op) and the default strategy is `.dense` (hybrid is opt-in and flagged provisional). The `.dense` path is unaffected — its scores are cosine, so the threshold semantics are correct there.

**Suggested fix during the #1937 pass:** either normalize fused/BM25 scores into a comparable range before threshold application, or make the threshold strategy-aware (cosine floor for `.dense`, a separate rank/score cutoff for BM25/RRF). The eval harness (#1937) is the right place to validate whatever normalization is chosen alongside the `k=60 / k1=1.2 / b=0.75` constants.

(Surfaced during a second review pass; the rest of the BM25/RRF math and the default `.dense` path were verified correct.)

Issue #1920, opened 2026-06-20
feat: first-party cloud reranker (Cohere/Jina) — on-device rerank stage already ships


roryford:
**Cluster:** RAG quality · **Impact:** High · **Effort:** Med

Retrieve ~20 candidates, rerank to top ~5 via an on-device cross-encoder (optional Cohere/Jina).

**Why:** AnythingLLM reranker, Msty Jina reranking. ~100–500ms for a large recall-precision gain. (Confirm current `RAGService` rerank depth first.)

---
_Filed from the June 2026 competitive-research sweep (4 parallel clusters: Swift-native ecosystem, local-UI/RAG/voice, agent/MCP/durable-execution, cross-platform runtimes). Verify the cited primitive in `Sources/` before implementing — workers confirmed the gaps but did not exhaustively trace existing depth._

---

## 🔬 Investigation & implementation plan (2026-06-20)

**VERDICT: Descope (~80% already shipped).** The reranker *stage* is fully built and wired — protocol, candidate widening, graceful fallback, bootstrap injection, integration tests. The only real gap is a **first-party on-device reranker implementation**; the orchestration this issue describes already exists.

### A. Ground truth — the stage EXISTS
- `Reranker` protocol + `PassthroughReranker` default: `Sources/ManifoldContract/Reranker.swift` (full docstring on placement, fallback contract, cross-encoder rationale).
- `RAGService` already runs the rerank stage: `Sources/ManifoldRuntime/Services/RAGService.swift` L28 (stored `reranker`), L38 `rerankCandidateMultiplier = 3`, L179–182 candidate widening when `reranker?.isReady`, L210–220 rerank call with **graceful degradation** (throwing reranker falls back to first-stage order, L214–217).
- Bootstrap injection: `Sources/ManifoldPersistenceSwiftData/ManifoldBootstrap.swift` L21–45 documents passing a `reranker:` and points adopters at `LlamaReranker` from the companion `manifold-llama` package.
- Integration tests: `Tests/ManifoldPersistenceSwiftDataTests/RAGServiceRerankIntegrationTests.swift` — `ClosureReranker` (L118) proves the stage reorders through real stores; `test_retrieve_noReranker_preservesCosineOrderThroughRealStores` (L87) proves the no-reranker path is unchanged.

So "retrieve ~20, rerank to top ~5, graceful fallback" is **done**. The 3× multiplier × default limit 5 = 15 candidates (close to the issue's "~20", trivially tunable).

[...]

PR #1988, opened 2026-06-21
feat: first-party cloud reranker (Cohere/Jina) over the shipped Reranker seam (#1920)

roryford:
## What shipped

A first-party cloud cross-encoder reranker, `CloudReranker` (in `ManifoldCloudSaaS`), conforming to the **already-shipped** `Reranker` protocol. The rerank stage itself (`Reranker` + `PassthroughReranker` default; `RAGService` candidate-widening + graceful-fallback call) was already wired — the only gap was a concrete cloud conformer, which this adds.

- POSTs `{model, query, documents}` to a hosted `/rerank` endpoint (Bearer auth).
- Maps each returned `relevance_score` back onto its source hit's `VectorSearchHit.score`, reorders by descending relevance, returns top-`limit`.
- `isReady == false` while the API key is empty (matches SaaS key handling) → keeps `RAGService` on its pre-rerank path.
- `do/catch` + `Log.network` throughout, never `try?`. Errors surface as `CloudBackendError` (network / serverError / parseError).
- **No change** to `RAGService` or the `Reranker` protocol — the seam already accepts any conformer.
- Extends the `ManifoldBootstrap` / `RAGConfiguration` docstring with the cloud option alongside the on-device `LlamaReranker`.

## One-class-vs-two decision

**One `CloudReranker` parameterized by endpoint + model, with `.cohere(...)` / `.jina(...)` presets** — the issue's preferred option. Cohere and Jina `/rerank` share the same request (`{model, query, documents}`) and response (`{results: [{index, relevance_score}]}`) shapes, so a single class with presets is cleaner than two near-identical types and mirrors the existing cloud-adapter convention.

## Tests

`Tests/ManifoldBackendsTests/CloudRerankerTests.swift` (XCTest, `MockURLProtocol` with UUID-based `.test` hosts per suite, no `try?`):
- `test_rerank_mapsApiScoresAndReorders` — API response that **inverts** cosine order; asserts output order + scores follow the API, not pass-through.
- `test_rerank_respectsLimit` — `limit` truncates after the API-driven reorder.
- `test_rerank_throwsOnHttpError` — 500 → throws.
- `test_isReady_falseWithoutKey` — empty key → not ready.

> Placed in `ManifoldBackendsTests` (the existing home for the cloud/Foundation surface) rather than a new `ManifoldCloudSaaSTests` target, which does not exist — avoids adding a CI test target.

## Gate

- `swift build --build-tests` clean (only pre-existing unrelated deprecation warnings).
- `CloudRerankerTests` + `ManifoldBackendsTests` run result reported in a follow-up comment.

### DIFF
--- a/Sources/ManifoldCloudSaaS/CloudReranker.swift
+++ b/Sources/ManifoldCloudSaaS/CloudReranker.swift
@@ -1,0 +1,200 @@
+import Foundation
+import ManifoldContract
+import ManifoldCloudCore
+
+/// A first-party cloud cross-encoder reranker conforming to ``Reranker``.
+///
+/// `CloudReranker` is the cloud counterpart to the on-device `LlamaReranker`
+/// (from `ManifoldLlama`): it POSTs the query and the widened candidate set to
+/// a hosted `/rerank` endpoint, maps the returned relevance scores back onto
+/// each ``VectorSearchHit/score``, and returns the top `limit` hits in
+/// descending relevance order. It plugs into the shipped ``Reranker`` seam
+/// without any change to `RAGService` — the retriever already widens its
+/// candidate pool and calls `rerank(query:candidates:limit:)` with graceful
+/// fallback when the call throws.
+///
+/// One class covers both supported providers because the Cohere and Jina
+/// `/rerank` request and response shapes are identical at the level we use:
+/// the request is `{model, query, documents}` and the response is
+/// `{results: [{index, relevance_score}]}`. Construct via the ``cohere`` /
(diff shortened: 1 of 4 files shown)


roryford:
## Gate results

- `swift build --build-tests` — clean (only pre-existing unrelated deprecation warnings).
- `CloudRerankerTests` — 4/4 passed.
- `ManifoldBackendsTests` (the existing cloud/Foundation surface, now including the 4 new tests) — 665 XCTest + 183 Swift Testing, 0 failures (4 pre-existing skips).

PR #2027, opened 2026-06-22
feat(tools): attribute transcripts by backend/model/quant and add conformance scorer

roryford:
## What

The `manifold-tools` JSONL transcript previously carried only a `scenario` field — not the model or backend. A multi-model run (`--model A,B,C`) appends every (scenario × model) run's events to one file, so per-model scoring meant parsing stdout. This blocks building a `(model × quant × backend)` tool-call conformance comparison across the Ollama / MLX / llama backends.

## Part A — per-record attribution

`TranscriptLogger.init` gains optional `backend`, `model`, and `quant` parameters. Every emitted record is stamped with these keys when set.

**Approach chosen: per-record, not a run-header row.** Because the CLI appends every model's events to a single file, a one-off header is ambiguous once runs interleave. Per-record attribution makes each line independently scorable. The change is fully additive — keys are only written when non-nil, so transcripts from callers that don't supply attribution keep their exact original shape and field names.

The CLI (`main.swift`) now constructs one logger per (backend, model) run, all appending to the same `--output` file, passing the real backend (`ollama`/`mock`) and model. A best-effort `quantLabel(from:)` derives a quant label from common GGUF/Ollama suffixes (`q4_K_M`, `q8_0`, `int4`, `fp16`, …); `nil` when nothing matches.

## Part B — ConformanceScorer

New `ConformanceScorer` (in the existing `ManifoldTools` target — no new product, no file moves) parses a transcript JSONL and emits normalized `Codable` `ResultRow`s, one per (backend, model, quant, scenario), with `assertionsPassed`, `assertionsFailed`, `errored`, `toolCallCount`, and a derived `Verdict` (`.pass` / `.partial` / `.fail` / `.errored`). Includes `encodeJSON` and a CSV helper. Malformed/torn lines are skipped, not fatal.

**Follow-up (step 2, not this PR):** promote `ConformanceScorer` to a shared `.library` product so the companion manifold-mlx / manifold-llama repos can reuse it.

Also added an optional `manifold-tools score <file.jsonl> [--csv]` subcommand that prints the matrix.

## Tests

Added to `Tests/ManifoldToolsTests/ScenarioRunnerTests.swift` (XCTest, matching the existing file):
- backend/model/quant stamped on every record, recoverable after round-trip
- quant omitted when nil; bare logger keeps the original record shape (backward compat)
- scorer over a fixture covering all-pass / partial / errored across two models, plus a torn final line
- fail-verdict when no assertions pass
- JSON round-trip of result rows

## Files changed
- `Sources/ManifoldTools/TranscriptLogger.swift`
- `Sources/ManifoldTools/ConformanceScorer.swift` (new)
- `Sources/manifold-tools/main.swift`
- `Tests/ManifoldToolsTests/ScenarioRunnerTests.swift`

Compiles (`swift build --product manifold-tools` + `swift build --build-tests`). CI is the authoritative gate.

### DIFF
--- a/Sources/ManifoldTools/ConformanceScorer.swift
+++ b/Sources/ManifoldTools/ConformanceScorer.swift
@@ -1,0 +1,218 @@
+import Foundation
+
+/// Parses a ``TranscriptLogger`` JSONL file and reduces it to normalized,
+/// scorable result rows — one per (backend, model, quant, scenario).
+///
+/// The transcript interleaves every (scenario × model) run the CLI executed, so
+/// scoring per-model previously meant parsing free-form stdout. With per-record
+/// attribution (`backend` / `model` / `quant` on each row), this scorer groups
+/// the events back into self-contained result rows that downstream tooling — a
+/// CI matrix, a `(model × quant × backend)` comparison across the Ollama / MLX /
+/// llama backends — can consume directly.
+///
+/// Designed to be promotable to a shared `.library` product later so the
+/// companion MLX/llama repos can reuse it; for now it lives inside
+/// `ManifoldTools` (no new product, no file moves — that promotion is follow-up).
+public enum ConformanceScorer {
+
+    /// Derived per-run verdict.
+    public enum Verdict: String, Codable, Sendable {
+        /// Every assertion passed, at least one assertion ran, and no error.
+        case pass
+        /// Some assertions passed and some failed (no error).
+        case partial
+        /// No assertion passed, or zero assertions ran (and no error).
+        case fail
+        /// The run recorded a harness/tool error.
+        case errored
+    }
+
+    /// One normalized result row per (backend, model, quant, scenario).
+    public struct ResultRow: Codable, Sendable, Equatable {
+        public let backend: String?
+        public let model: String?
+        public let quant: String?
+        public let scenario: String
+        public let assertionsPassed: Int
+        public let assertionsFailed: Int
+        public let errored: Bool
+        public let toolCallCount: Int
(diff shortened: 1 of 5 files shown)


roryford:
## Update: scorer aligned with the shared cross-backend metric

Follow-up commit `936a9946` reworks the scorer so it speaks the **same tool-selection metric the MLX/llama soak CLIs already report**, instead of a parallel verdict-only vocabulary — otherwise the Ollama results couldn't be keyed alongside MLX/llama in a (model × quant × backend) matrix (the whole point).

- `TranscriptLogger.prompt` now records the scenario's `requiredTools` (expected tool set) — additive, tolerated-absent for older transcripts.
- `ConformanceScorer` reuses **`ManifoldInference.ConfusionCounts` / `MacroAveragedMetrics`**: per-row tool-selection `tp/fp/fn` (called vs required; decoys → FP) with computed precision/recall/F1, plus `aggregate()` macro-average over tool-bearing rows. The assertion-derived `verdict` stays as the scenario-level rollup.
- `score` subcommand prints a `SUMMARY rows=… tool_bearing=… precision=… recall=… f1=…` line to **stderr** (stdout stays pure JSON/CSV).

Step-2 follow-up unchanged: promote the scorer to a shared `.library` product so manifold-mlx / manifold-llama import it directly (llama currently has no scoring at all).

Local: `swift test --filter ManifoldToolsTests` → **53 executed, 0 failures** (incl. a new tool-selection ConfusionCounts test). No new force-unwraps, no Package.resolved churn.

PR #2049, opened 2026-06-26
fix(tools): remove two conformance-harness false-negatives

roryford:
## What

Two scoring artifacts in the tool-calling conformance harness were **understating real scores** — neither is a model defect. Found while auditing the `soak-20260626-100115` results.

### Bug A — `oversize-tool-output` demanded a literal phrasing (affects every backend)
The final-answer assertion was `containsAll: ["exceeds maxBytes", "narrower"]`. A behaviourally-perfect answer (e.g. qwen3.5-9b: calls `read_file`, hits the policy, *"…output size (41,203 bytes) exceeds the maximum allowed limit of 32,768 bytes… read a smaller portion"*, no hallucination) **fails** because it paraphrases instead of echoing those exact substrings — **0 passes across all 5 soak models** despite F1=1.0 tool selection.

Fix: add a `containsAny` assertion kind and re-express the oversize final-answer check as a size-acknowledgement OR-set **and** a narrowing OR-set. The `toolResultErrorKind` honesty gate is unchanged, so a hallucinated answer still fails.

### Bug B — a built-in's required tool was hidden from companion scoring (affects llama.cpp / MLX)
The scorer's expected-set recovery — the fallback for transcripts that omit the `requiredTools` prompt field (companion soaks) — only extracts a tool named in the backtick or bare-dispatch form. Scenarios **06** and **07** worded their `read_file` requirement as free prose, so a *correct* `read_file` call was mis-scored as a false positive. Verified against a real llama-Mistral transcript: those cells read F1≈0.81 with `parallel-readme-comparison` wrongly at 0.0.

Fix: name the tool in backticks in both `toolInvoked` messages (Ollama, which writes authoritative `requiredTools`, was unaffected). A new `ScenarioRecoveryTests` case asserts **every** built-in `toolInvoked` assertion exposes its tool to recovery, so this can't reappear when a scenario is added or reworded.

## Not in scope (already fixed, verified during the audit)
- No-tool F1 drag — already handled (`toolSelection: nil`, excluded from the mean).
- llama.cpp TP attribution (F1=0.000) — already fixed by #2043; current scorer scores that transcript F1≈0.81.

## Tests
- `containsAny` pass / fail / missing-values
- oversize accepts the canonical correct paraphrase, still rejects a hallucinated answer
- built-in recovery coverage for all scenarios

## Backend checklist
- [x] Ollama (unaffected by B; A flips `oversize` to pass for competent models)
- [x] llama.cpp / MLX (B fixes understated `read_file` cells — confirmed on a retained transcript; re-measured by the next cross-backend soak)

### DIFF
--- a/Sources/ManifoldTools/ConformanceScorer.swift
+++ b/Sources/ManifoldTools/ConformanceScorer.swift
@@ -352,6 +352,17 @@ public enum ConformanceScorer {
     /// safe (it never manufactures a false positive *or* a false true-positive),
     /// whereas loose prose matching could. A genuinely wrong tool therefore still
     /// scores as FP.
+    ///
+    /// Because this matcher is deliberately strict, it only recovers a tool whose
+    /// dispatch-requirement assertion *names it in the recognized form* (backtick or
+    /// the bare `… to actually be dispatched` frame). A built-in scenario that words
+    /// its requirement as free prose hides the tool here, so a companion transcript
+    /// (which omits `requiredTools`) mis-scores the correct call as an FP — keep the
+    /// `toolInvoked` messages naming the tool in backticks (a `ScenarioRecoveryTests`
+    /// case guards this for all built-ins). The durable fix is to log the
+    /// `toolInvoked` assertion's structured tool `value` in the transcript and prefer
+    /// it over message parsing; it is deferred because it changes the transcript
+    /// schema and the companions would have to re-vendor + re-run to benefit.
     static func expectedToolsFromAssertion(_ message: String) -> [String] {
         guard message.hasPrefix("Scenario requires") else { return [] }
         var tools: [String] = []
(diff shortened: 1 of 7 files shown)


roryford:
Local `scripts/test.sh --profile local` gate: **5640 passed, 2 failed** — both failures are pre-existing **local-environment** issues unrelated to this diff (which is 7 files, all under `ManifoldTools`):

- `QuickStartTests.test_quickStart_seededEndpoint_canLoadViaSelectedEndpoint` — hits live Ollama at `localhost:11434` and got `404 Model not found` because this dev box's Ollama (running for soak work) doesn't have the seeded model pulled.
- `ModelManagementSheetLogicTests.test_discoverModelsOffMain_matchesSyncDiscovery` — sensitive to the real `~/Documents/Models/` contents (full of soak GGUFs); `clean-leaked-test-artifacts.sh` found 0 leaks.

Neither code path is reachable from this change; the `ManifoldTools` target is fully green (94 tests). Relying on clean-env CI as the authoritative gate for those two.

